PR #3, opened 2025-07-16
[Wave] fix scalar type annotation

willghatch:
The order was backwards (compared to the implementation and uses in tests), and the value can be an int as well as a float.


harsh-nod (approved):
thanks!

PR #5, opened 2025-07-16
[Wave] fix DCE

willghatch:
The recursion in DCE was checking only for global writes, not for other kinds of side-effects that should not be removed (such as WIP printf inside conditionals).


willghatch:
There are currently no unit tests for passes like dead code elimination, as far as I see.  I'm happy to add a test.

willghatch:
@harsh-nod I've added a test that fails without this patch.

harsh-nod:
This seems to be failing the pre-commit hook for imports. Not sure why the builders didn't catch that.

willghatch:
Except that was a newly added file, so that reasoning can't apply...

PR #7, opened 2025-07-16
[Wave] Build wave runtime in setup.py

harsh-nod:
This PR includes the build of wave runtime in the top-level setup.py. With the recent removal of torch and hip deps in wave runtime, it became easier to build wave runtime as part of iree-turbine.

Here we move the setup.py from iree/turbine/kernel/wave/runtime into the top-level setup.py and update the pyproject.toml build deps accordingly.


.github/workflows/build_packages.yml
```diff
@@ -22,6 +22,29 @@ jobs:
...
+            package: wave-lang
+            python-version: cp311-cp311
+          - runs-on: ubuntu-24.04
+            platform: linux-x86_64
+            package: wave-lang
+            python-version: cp312-cp312
+          - runs-on: ubuntu-24.04
+            platform: linux-x86_64
+            package: wave-lang
+            python-version: cp313-cp313
```
marbre, line 47: Quick comment: You can build for different Python versions in the `build_tools/build_linux_package.sh` script:
```
PYTHON_VERSIONS="${OVERRIDE_PYTHON_VERSIONS:-cp310-cp310 cp311-cp311 cp312-cp312 cp313-cp313}"
```
No need to span a matrix here.

build_tools/build_linux_package.sh
```diff
@@ -0,0 +1,181 @@
...
+      echo "ERROR: Could not find python: ${python_dir} (skipping)"
+      continue
+    fi
+    export PATH="${python_dir}/bin:${orig_path}"
+    echo ":::: Python version $(python --version)"
+
+    # TODO: Switch to wave on repo change
+    clean_wheels "iree_turbine" "${python_version}"
+    build_wave
+    run_audit_wheel "iree_turbine" "${python_version}"
```
marbre, line 122: Follow what the TODO says? ;)

marbre (changes requested):
Just a quick feedback. Beside the inline comments, `pyproject.toml` needs improvement, see https://packaging.python.org/en/latest/guides/writing-pyproject-toml/ and our other projects.

Furthermore, the PR title doesn't match the changes.

.github/workflows/build_packages.yml
harsh-nod: thanks, will remove from here.

harsh-nod:
> Just a quick feedback. Beside the inline comments, `pyproject.toml` needs improvement, see https://packaging.python.org/en/latest/guides/writing-pyproject-toml/ and our other projects.
> 
> Also, the `setup.py` technically can stay close to the source. There is no need to move it top level, especially if you're aiming to build multiple packages from the sources within the repo (?).
> 
> The PR title should also be updated, as a `requirements.txt` isn't touched.

What other changes to pyproject.toml would you suggest? Anything specific to the C++ CMake build or just general improvements?

I am hoping to have just one package built so would like to consolidate everything into a single package, so we can just do
pip install wave.

Thanks, I have updated the PR title.

harsh-nod:
@ScottTodd @marbre - is this good to land ?

.github/workflows/build_packages.yml
marbre: I was actually wrong and you want to spin the matrix here to build for different Python versions in parallel.

.github/workflows/build_packages.yml
```diff
@@ -39,6 +43,11 @@ jobs:
           ./build_tools/compute_local_version.py -rc --write-json
           ./build_tools/build_release.py --no-download
 
+      - name: Build C++ dependencies (Linux x86_64)
```
marbre, line 46: This misses
```
      env:
        OVERRIDE_PYTHON_VERSIONS: "${{ matrix.python-version }}"
```
see for example https://github.com/nod-ai/shark-ai/blob/main/.github/workflows/build_packages.yml.

That is also the reason why you want to spin the matrix. If you don't override here, there is no need for the matrix and that is what confused me during the first round of reviews.

marbre:
> @marbre - does this look good now?

Can you trigger a test run to see if it works as expected?

harsh-nod:
> > @marbre - does this look good now?
> 
> Can you trigger a test run to see if it works as expected?

Just did that and it passed! https://github.com/iree-org/wave/actions/runs/16372868034

harsh-nod:
> > > @marbre - does this look good now?
> > 
> > 
> > Can you trigger a test run to see if it works as expected?
> 
> Just did that and it passed! https://github.com/iree-org/wave/actions/runs/16372868034

@marbre - would you like me to try anything else?

PR #8, opened 2025-07-16
[Wave] Update documentation page

harsh-nod:
This PR updates the titles of the docs and the links to the documentation for wave. A new read the docs site has been created that can be used to host all wave related documentation.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,14 +1,14 @@
-# IREE Turbine Documentation
+# Wave Documentation
 
 This directory uses Sphinx to build documentation that is hosted at
-https://iree-turbine.readthedocs.io/.
+https://wave-lang.readthedocs.io/.
 
 ## Publishing on Read the Docs
 
 The project dashboard is here:
-https://app.readthedocs.org/projects/iree-turbine/.
+https://app.readthedocs.org/projects/wave-lang/.
 
-A webhook is configured in the iree-turbine GitHub repository to notify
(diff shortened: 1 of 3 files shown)


docs/README.md
```diff
@@ -26,7 +26,7 @@ source .venv/bin/activate
 python -m pip install -r requirements.txt
 python -m pip install -r ../pytorch-cpu-requirements.txt
```
xintin, line 27: nit: requirements and pytorch-cpu-requirements are at the same dir level. We do not need to go one level up.

docs/README.md
```diff
@@ -26,7 +26,7 @@ source .venv/bin/activate
...
 
-# Install iree-turbine itself.
+# Install wave itself.
 # Editable so you can make local changes and preview them easily.
 python -m pip install -e ..
```
xintin, line 31: same as earlier: parent dir vs current dir.

xintin (changes requested):
left a couple of nits

docs/README.md
harsh-nod: The requirements.txt for docs is at the docs level and pytorch-cpu-requirements is one level up. This is not the same requirements.txt at the top level.

docs/README.md
harsh-nod: This is for editing from the docs directory. That's why the parent directory. I will update the top level README after we land the setup.py changes.

PR #10, opened 2025-07-17
[TKW] In-thread transpose

Hardcode84:
Add an in-thread transpose optimization. in-thread transpose is a special case of minimize-global-loads optimization when the global read also doing the transpose.

Let's say minimize-global-loads will result 4 x vector<8xf16> writes to shared mem, In this case we can do 8 x vector<4xf16> loads from global and transpose the tile in registers.


harsh-nod (approved):
thanks! lgtm!

PR #11, opened 2025-07-17
[Wave] Enable async kernel execution with iree runtime and fully switch to Launchable

Hardcode84:
Drop our `invoke_vmfb` code entirely, use `turbine.runtime.Lauchable` and udate dispatch codegen to support async launch.

Before:
```
  func.func @isolated_benchmark(%arg0: tensor<8x?x128x6xf32>, %arg1: tensor<8x?x6xf32>, %arg2: tensor<?xi32>, %arg3: tensor<?x6x128xf16>, %arg4: index) -> tensor<?x6x128xf16> {
    %0 = flow.dispatch @phase_1::@phase_1[%arg4](%arg0, %arg1, %arg2, %arg3, %arg4) : (tensor<8x?x128x6xf32>{%arg4}, tensor<8x?x6xf32>{%arg4}, tensor<?xi32>{%arg4}, tensor<?x6x128xf16>{%arg4}, index) -> %arg3{%arg4}
    return %0 : tensor<?x6x128xf16>
  }
```
After:
```
  func.func @isolated_benchmark$async(%arg0: !hal.buffer_view, %arg1: !hal.buffer_view, %arg2: !hal.buffer_view, %arg3: !hal.buffer_view, %arg4: index, %arg5: !hal.fence, %arg6: !hal.fence) -> !hal.buffer_view {
    %0 = hal.tensor.import wait(%arg5) => %arg0 : !hal.buffer_view -> tensor<8x?x128x6xf32>{%arg4}
    %1 = hal.tensor.import wait(%arg5) => %arg1 : !hal.buffer_view -> tensor<8x?x6xf32>{%arg4}
    %2 = hal.tensor.import wait(%arg5) => %arg2 : !hal.buffer_view -> tensor<?xi32>{%arg4}
    %3 = hal.tensor.import wait(%arg5) => %arg3 : !hal.buffer_view -> tensor<?x6x128xf16>{%arg4}
    %4 = flow.dispatch @phase_1::@phase_1[%arg4](%0, %1, %2, %3, %arg4) : (tensor<8x?x128x6xf32>{%arg4}, tensor<8x?x6xf32>{%arg4}, tensor<?xi32>{%arg4}, tensor<?x6x128xf16>{%arg4}, index) -> %3{%arg4}
    %5 = hal.tensor.barrier join(%4 : tensor<?x6x128xf16>) => %arg6 : !hal.fence
    %6 = hal.tensor.export %5 : tensor<?x6x128xf16>{%arg4} -> !hal.buffer_view
    return %6 : !hal.buffer_view
  }
```

Also, add some python profiling code to `WaveKernel` launch func.

Launch overhead for the iree runtime is still bad (around x5 of `wave_runtime`) but ability to run kernels async is overall improvement.


wave_lang/kernel/wave/compile.py
```diff
@@ -94,24 +113,82 @@ def invoke(self, *args, **kwargs):
...
+                self.gpu_func,
+            )
+        else:
+            tensors = [t.data for t in chain(kernel_inputs, kernel_outputs)]
+            self.launchable(*tensors, *scalar_args, *dynamic_symbols)
+
+            if self.options.run_bench:
+                benchmark_flags = {}
+                benchmark_flags["batch_size"] = self.options.benchmark_batch_size
+
```
martin-luecke, line 133: If `self.options.run_bench` is True, we will not actually bench this kernel if `self.options.wave_runtime` is also true, right? 
While we do not have this functionality in wave runtime let's abort or print a warning if both options are true.

lit_tests/kernel/wave/codegen.py
```diff
@@ -2178,7 +2178,7 @@ def scalar_codegen_f32(
 
     # Final dispatch args dtype
     # CHECK: flow.dispatch @scalar_codegen_f32::@scalar_codegen_f32(
-    # CHECK-SAME: %arg0, %arg1, %arg2, %arg3)
+    # CHECK-SAME: %{{.*}}, %{{.*}}, %arg2, %arg3)
```
martin-luecke, line 2181: Why is this necessary for arg0 and arg1?

lit_tests/kernel/wave/codegen.py
```diff
@@ -2231,7 +2231,7 @@ def scalar_codegen_i32(
 
     # Final dispatch args dtype
     # CHECK: flow.dispatch @scalar_codegen_i32::@scalar_codegen_i32(
-    # CHECK-SAME: %arg0, %arg1, %arg2, %arg3)
+    # CHECK-SAME: %{{.*}}, %{{.*}}, %arg2, %arg3)
```
martin-luecke, line 2234: Same: Why is this necessary for arg0 and arg1?

wave_lang/kernel/wave/utils/run_utils.py
```diff
@@ -6,182 +6,35 @@
...
-import iree.runtime as rt
 import torch
 
 from wave_lang.kernel.lang import IndexSymbol
 
 from ..compile_options import WaveCompileOptions
-from ..profiling import benchmark_module
-
-# Cache for the system context and vm function.
-RUNTIME_CACHE: dict[str, tuple[rt.SystemContext, rt.VmFunction]] = {}
```
martin-luecke, line 16: Could you briefly explain the effect this and `invoke_vmfb` has on caching? Do we not cash at all when using wave runtime and only using iree runtime or not entirely anymore? Is this documented somewhere?

martin-luecke (approved):
LGTM after addressing a few questions

wave_lang/kernel/wave/compile.py
Hardcode84: yeah, I will add a check

wave_lang/kernel/wave/utils/run_utils.py
Hardcode84: `Launchable` is doing the same type of caching internally.

lit_tests/kernel/wave/codegen.py
martin-luecke: Does it then make sense to match these using SSA names as well? Or possibly checking that the DAG is correct, i.e. matching that the SSA value resulting from a specific op is used here?

PR #20, opened 2025-07-18
[Wave] Add schedule reordering support for BMK,NK->BMN

raikonenfnu:
Linear layers are typically of this format BMK,NK->BMN, previously we would have failed to determine batch dim because rhs do not have B, but with this patch, we implemented a batch filtering dimension by checking the vector_shapes from hw_constraint.

### DIFF
--- a/lit_tests/kernel/wave/scaled_gemm.py
+++ b/lit_tests/kernel/wave/scaled_gemm.py
@@ -358,3 +358,151 @@ def packed_mxfp4_test():
 
     # Epilogue MFMA
     # CHECK-COUNT-64: amdgpu.scaled_mfma
+
+
+@run_test
+def batched_prefetch_mxfp4_test():
+    mfma_variant = tkw.ScaledMMAType.F32_16x16x128_F8F6F4
+    # Input sizes
+    B = tkl.sym.B
+    M = tkl.sym.M
+    N = tkl.sym.N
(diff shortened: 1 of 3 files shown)


lit_tests/kernel/wave/scaled_gemm.py
```diff
@@ -358,3 +358,151 @@ def repeat(acc: tkl.Register[M, N, tkl.f32]) -> tkl.Register[M, N, tkl.f32]:
...
+    constraints += [
+        tkw.HardwareConstraint(
+            threads_per_wave=64, vector_shapes={B: 0}, mma_type=mfma_variant
+        )
+    ]
+
+    @tkw.wave(constraints)
+    def batched_gemm_mxfp4_prefetch(
+        a: tkl.Memory[B, M, K / 2, ADDRESS_SPACE, tkl.i8],
+        a_scale: tkl.Memory[B, M, K / 32, ADDRESS_SPACE, tkl.i8],
```
nithinsubbiah, line 396: curious why does the scale's dimension match the vector dim?

nithinsubbiah (approved):

lit_tests/kernel/wave/scaled_gemm.py
nithinsubbiah: `K/2` and `K/32` for `a` vector and its scale

PR #28, opened 2025-07-18
[Wave] Fix visualization test

harsh-nod:
With the recent removal of waves_per_block, the visualization test was failing since it requires the waves_per_block to be specified. This PR fixes that.


Hardcode84:
I'm late to the review, can we actually make it tested on the CI (by installing the graphviz there I suppose)?

PR #30, opened 2025-07-18
[Wave] More tests cleanup

Hardcode84:
* Remove remaining `--runperf`
* Run CI on draft PRs
* Create temp files in temp dir

### DIFF
--- a/.github/workflows/ci-tk.yaml
+++ b/.github/workflows/ci-tk.yaml
@@ -107,7 +107,7 @@ jobs:
 
 
   test:
-    name: "Unit Tests and Type Checking"
+    name: "${{ matrix.os }} :: ${{ matrix.version }} :: Unit Tests and Type Checking"
     strategy:
       fail-fast: false
       matrix:
--- a/tests/kernel/wave/wave_e2e_test.py
+++ b/tests/kernel/wave/wave_e2e_test.py
@@ -80,7 +80,7 @@ def get_test_shapes(test_name: str) -> list[tuple[int]]:
 
 @require_e2e
 @pytest.mark.parametrize("shape", get_test_shapes("test_copy")[:1])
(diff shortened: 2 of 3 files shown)


raikonenfnu:
> Run CI on draft PRs

I was hoping we can keep draft PR as a place people can discuss about WIP or RFC PRs but without incurring cost of saturating the CI machines. Then, if you actually want to run CI, we can change it to ready for review. Then if you actually want people to start reviewing, you can put them as reviewer and/or ping. 

Since this may be more of a developer flow/ergonomics thing would like @harsh-nod, @nithinsubbiah, @xintin to probably chimei n as well. :)

Hardcode84:
> I was hoping we can keep draft PR as a place people can discuss about WIP or RFC PRs but without incurring cost of saturating the CI machines. 

We don't have nearly enough draft PRs for this to matter and I don't know know any project which do this (IREE doesn't). Also, the current approach makes CI to restart when you edit the PR description, which is much worse.

xintin:
I think in the past the number of draft PRs increased, which is why we set ci to not run on draft PRs. This was useful to reduce noise.

We also need to decide on a standard policy: Is anything not marked as a draft PR available for review? Because tagging with [WIP] label alone may not be useful, as it will still run CI, thus defeating the purpose of turning it off for the draft ones.

Alternatively, it is possible to set reduced or selective ci actions for draft PRs.

Hardcode84:
The standard policy is "everything which is not draft is up to review" and that one of the reasons I want to run CI on drafts.

xintin:
I second this std practice.
Let's have a quick discussion in the next meeting. Having `WIP` tag and draft is redundant. 
For the draft PRs, we can have `workflow_dispatch` to trigger it manually if required to test ci on a draft PR.

xintin (approved):

PR #56, opened 2025-07-21
Add scatter_add operation

adedespirlet:
The scatter_add  operation performs element-wise accumulation from a source register into shared memory (LDS) using dynamic indices along a specified dimension. It supports both integer and floating-point data types and is lowered to memref.atomic_rmw using either addi or addf, depending on the type.

Input arguments:
-	src: a register containing the values to scatter
-	index a register containing the indices along dim where values should be added
-	dim the dimension along which scattering is performed
-	memory: the target tensor in LDS where results are accumulated
-	mapping: the index mapping representing mapping between sets of indices
-	elements_per_thread: number of elements each thread loads at once in its VGPR

Currently scatter_add accepts only static one-to-one index mappings. The lowering logic internally performs dynamic indexing , combining index and dim with the provided mapping to compute final memory access indices. Potential Future Work: rewrite the operation handler and API to align with the write() implementation where in case of a scatter operation, index mapping expressions passed through are fully dynamic.

Limitations:
Only intra-workgroup scattering is supported (i.e. within LDS/ shared memory) with a single wave. Multi-wave execution is not guaranteed to be safe: synchronization issues may occur when threads write to the same index. Further investigation is needed. The operation handles multiple elements per thread, provided that the non-scatter dimension is large enough (> elements_per_thread)

### DIFF
--- a/wave_lang/kernel/wave/codegen/read_write.py
+++ b/wave_lang/kernel/wave/codegen/read_write.py
@@ -27,6 +27,9 @@ from wave_lang.support.ir_imports import (
     memref_d,
     vector_d,
 )
+from wave_lang.aot.support.ir_utils import (
+    _is_float_type,
+)
 
 from ..._support.indexing import IndexExpr, IndexingContext, IndexSequence, IndexSymbol
 from ...compiler.base import ValidationError
@@ -46,6 +49,7 @@ from ...ops.wave_ops import (
     get_custom,
     read,
     write,
+    scatter_add,
 )
 from ..utils.general_utils import get_fastest_index, infer_dim
 from ..utils.mapping_utils import transform_index_on_mapping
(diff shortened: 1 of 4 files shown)


wave_lang/kernel/wave/codegen/read_write.py
```diff
@@ -978,3 +982,97 @@ def handle_gather_to_lds(emitter: WaveEmitter, node: fx.Node):
...
+                # TODO: Ideally emit a vectorized atomic op instead of 4 scalar atomics that store to consecutive locations
+                fast_dim = other_dims[-1]
+                indices[fast_dim] = arith_d.addi(
+                    indices[fast_dim], arith_d.constant(IndexType.get(), i)
+                )
+        result = memref_d.atomic_rmw(rmw_kind, reg_elem, memory, indices)
+        results.append(result)
+
+    result_type = VectorType.get([elements_per_thread], register_src.type.element_type)
+    result_vector = vector_d.from_elements(result_type, results)
```
Hardcode84, line 1067: This code looks incomplete

PR #60, opened 2025-07-21
[Wave] Unaligned shapes support in `gather_to_shared` pass

Hardcode84:
* Check if fastest dimension mask is divisible by `GatherToLDS` vector size. 
* Switch codegen to buffer ops and use OOB logic to mask values.
* More fine control over which dimensions are dynamic in `get_gemm_kernel`


wave_lang/kernel/wave/gather_to_shared.py
```diff
@@ -307,10 +325,10 @@ def gather_to_shared(
         THREAD_2: THREAD_2 if waves_per_block[2] > 1 else 0,
     }
 
-    supported_load_widths = [32]
+    supported_load_widths = [4]
```
raikonenfnu, line 328: I kind of prefer the old 32, 96, 128 since it is more analogous to the instructions, any thoughts on doing the division inside `get_load_bitwidh` or maybe we can omit them altogether since we can change `vector_width = (elements_per_thread * bitwidth) // 8` to `vector_width = (elements_per_thread * bitwidth)`. Thoughts?

Hardcode84: In this case we will need to convert bounds to bitwidth, which is also weird.

PR #72, opened 2025-07-22
[WAVE] Updated wave speculativde decode as per the latest flashinfer kernel updates

xintin:
- updated as per the latest commits in the sglang/flashinfer speculative kernel. 
- this is without `top-p top-k`. That will be a part of the performance.


harsh-nod (approved):
Please remove explicit waves_per_block and then should be good to land.

PR #74, opened 2025-07-22
[Wave] add debug_log_write op

willghatch:
This is a first step to having a convenient print op for Wave. It implicitly adds global memory kernel arguments, while also adding those memory arguments automatically in the kernel's invoke method.

### DIFF
--- a/tests/kernel/wave/wave_e2e_test.py
+++ b/tests/kernel/wave/wave_e2e_test.py
@@ -2203,3 +2203,64 @@ def test_scatter_add(shape, elems_per_thread, request):
 
     torch_output = scatter_add_baseline(input, index)
     assert_close(output, torch_output)
+
+
+@require_e2e
+def test_debug_log_write():
+    M = tkl.sym.M
+    N = tkl.sym.N
+    ADDRESS_SPACE = tkl.sym.ADDRESS_SPACE
+
+    shape = (256, 128)
+
+    wave_size = 64
+    BLOCK_M = 1
+    BLOCK_N = sympy.Max(sympy.Min(shape[1], 256), wave_size)
+
+    constraints: list[tkw.Constraint] = [
+        tkw.HardwareConstraint(
+            threads_per_wave=wave_size,
+            vector_shapes={M: BLOCK_M, N: BLOCK_N},
+        )
+    ]
+
+    constraints += [tkw.WorkgroupConstraint(M, BLOCK_M, 1)]
+    constraints += [tkw.WorkgroupConstraint(N, BLOCK_N, 0)]
+    constraints += [tkw.WaveConstraint(M, BLOCK_M)]
+    constraints += [tkw.WaveConstraint(N, BLOCK_N)]
+
+    @tkw.wave(constraints)
+    def test(
+        a: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f16],
+        b: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f16],
+        c: tkl.Memory[M, N, ADDRESS_SPACE, tkl.f16],
+    ):
+        lhs = tkw.read(a)
+        rhs = tkw.read(b)
+        tkw.debug_log_write(lhs)
+        tkw.debug_log_write(rhs, log_name="rhslog")
+        res = lhs + rhs
+        tkw.debug_log_write(res)
+        tkw.write(res, c)
+
+    a = device_randn(shape, dtype=torch.float16)
+    b = device_randn(shape, dtype=torch.float16)
+    c = device_zeros(shape, dtype=torch.float16)
+    ref = a + b
+
+    options = WaveCompileOptions(
+        subs={
+            M: shape[0],
+            N: shape[1],
+            ADDRESS_SPACE: tkl.AddressSpace.GLOBAL_MEMORY.value,
+        },
+    )
+    options = set_default_run_config(options)
+
+    test = wave_compile(options, test)
+
+    debug_logs = {}
+    test(a, b, c, debug_logs=debug_logs)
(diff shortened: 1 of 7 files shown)


willghatch:
To reviewers, note the TODO about how it breaks caching.  I haven't yet plumbed the extra data through caching, but I wanted to get something up today.  Though, actually, now that I'm thinking about it more, I think that when there are debug log ops, it should just disable saving to the cache, because future things that I plan to build on top of this will maybe not be serializable (eg. taking python function/closure arguments for how to print things from the log won't be serializable).

willghatch:
Huh.  The line with the unpacking operator that CI is complaining about works on my machine.  I'll look into why that's breaking tomorrow, maybe some python version difference?

wave_lang/kernel/ops/wave_ops.py
```diff
@@ -778,6 +786,10 @@ def type(self, value: Any):
     def has_side_effects(self) -> bool:
         return False
 
+    @property
+    def debug_log_transformer(self) -> bool:
```
ftynse, line 790: Nit: can we have some doc on what is the contract that subclasses should follow when implementing this?

wave_lang/kernel/wave/debug_log_hoist.py
```diff
@@ -0,0 +1,99 @@
...
+
+class DebugArgInfo(TypedDict):
+    symbol_name: str
+    debug_output_arg_id: int
+    dtype: DataType
+    symbolic_shape: Tuple[IndexSymbol, ...]
+
+
+def is_debug_log_transformer(node):
+    return get_custom(node).debug_log_transformer
```
ftynse, line 26: Can't we just do `isinstance`?

wave_lang/kernel/wave/debug_log_hoist.py
```diff
@@ -0,0 +1,99 @@
...
+from .._support.tracing import CapturedTrace
+from ..lang.global_symbols import *
+from ..ops.wave_ops import (
+    get_custom,
+    Write,
+    Placeholder,
+)
+from .._support.dtype import DataType
+from .._support.indexing import IndexSymbol
+from typing import List, Tuple, TypedDict
```
ftynse, line 15: Nit: `List` and `Tuple` are deprecated since Python 3.9 https://docs.python.org/3/library/typing.html#typing.List. It looks like our support matrix starts with 3.10 so let's not use those maybe?

wave_lang/kernel/ops/wave_ops.py
```diff
@@ -2001,6 +2013,52 @@ def is_contiguous_vec(self) -> bool:
...
+    The memory can be accessed by passing an an extra keyword to kernel invokation "debug_logs" with an empty dictionary.
+    The dictionary will be mutated to contain the logs.
+
+    This is intended as a primitive for building more convenient debug logging tools.
+    """
+
+    register_: fx.Proxy
+    elements_per_thread: Optional[Any] = None
+    mapping: Optional[IndexMapping] = None
+    mapping_dynamic_vals: tuple[fx.Node, ...] = ()
```
ftynse, line 2032: This may be a bit too detailed for a debug log, leaking the fact that it is a memory write, but without knowing what memory we are writing into. It can create room for bugs, e.g., by having the user cargo-cult a mapping from another write that is not compatible with the sizes of the memory generated. And, in general, I don't think it's possible to infer the memory size sufficient to cover all addresses after mapping in presence of dynamic values. Do we really need these?

tests/kernel/wave/wave_e2e_test.py
```diff
@@ -2096,3 +2096,64 @@ def test(
...
+    )
+    options = set_default_run_config(options)
+
+    test = wave_compile(options, test)
+
+    debug_logs = {}
+    test(a, b, c, debug_logs=debug_logs)
+    assert_close(a, debug_logs["debug_log_output_0"])
+    assert_close(b, debug_logs["debug_log_output_1"])
+    assert_close(c, debug_logs["debug_log_output_2"])
```
ftynse, line 2159: Could we somehow add a string tag to the `debug_log_write` op and use it for indexing here instead of a positional key? It will be annoying to have to update all keys back and forth when someone adds or removes a debug statement.

wave_lang/kernel/ops/wave_ops.py
willghatch: Honestly, I'm not 100% sure yet what we need here.  I have written very little wave code, so I'm not yet sure about all the ways these get used.  I left them to be able to manually adjust the writing like you can with normal writes (since they ultimately are just syntax sugar for normal writes), but I'm not yet certain what would be best here.

wave_lang/kernel/wave/debug_log_hoist.py
willghatch: Yes.  I was thinking I would be writing a family of these, but I can just switch to doing a series of `isinstance` calls.

tests/kernel/wave/wave_e2e_test.py
ftynse: I actually think assertions as you written may be an important use case too. One would want to compare with a reference, potentially after moving the data to a different device, and find diverging elements rather than eyeballing a potentially large tensor.

wave_lang/kernel/ops/wave_ops.py
willghatch: I've removed the extra optional arguments, and they are easy enough to add back if they end up being necessary.

ftynse (approved):

ftynse:
You may need to `git commit --amend -s` on every commit to make the DCO check happy, or sqash-rebase and then `-s`.

PR #75, opened 2025-07-22
[WAVE] Merge two kernels into one in wave speculative decode

xintin:
- merged two kernels into one kernel
- updated lit test

### DIFF
--- a/lit_tests/kernel/wave/speculative_decoding.py
+++ b/lit_tests/kernel/wave/speculative_decoding.py
@@ -51,8 +51,15 @@ def test_speculative_decoding():
     # CHECK: scf.while
     # CHECK: vector.extractelement
     # CHECK: scf.condition
+    # CHECK: scf.while
     # CHECK: ^bb0
+    # CHECK: arith.divf
+    # CHECK: arith.cmpf
+    # CHECK: arith.ori
+    # CHECK: arith.xori
     # CHECK: vector.extract
+    # CHECK: scf.if
+    # CHECK: scf.if
     # CHECK: vector.load
     # CHECK: vector.store
     # CHECK: scf.yield
--- a/lit_tests/kernel/wave/speculative_decoding.py
+++ b/lit_tests/kernel/wave/speculative_decoding.py
@@ -32,21 +32,45 @@ def test_speculative_decoding():
     kernel = wave_compile(options, kernel)
     print(kernel.asm)
 
-    # CHECK: scf.while
-    # CHECK: scf.condition
-    # CHECK: scf.while
-    # CHECK: scf.condition
-    # CHECK: arith.divf
-    # CHECK: arith.cmpf
-    # CHECK: arith.xori
-    # CHECK: arith.select
-    # CHECK: scf.if
-    # CHECK: vector.store
-    # CHECK: scf.if
+    # CHECK: builtin.module {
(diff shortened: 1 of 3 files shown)


lit_tests/kernel/wave/speculative_decoding.py
```diff
@@ -32,21 +32,45 @@ def test_speculative_decoding():
...
+    # --- Typical operations (representative):
+    # CHECK: gpu.block_id  y upper_bound 10
+    # CHECK: gpu.thread_id  x upper_bound 64
+
+    # CHECK: vector.load
+    # CHECK: vector.store
     # CHECK: scf.while
+    # CHECK: vector.extractelement
     # CHECK: scf.condition
-    # CHECK: scf.while
```
harsh-nod, line 37: why did the scf.while go away from this function?

wave_lang/kernel/wave/templates/speculative_decoding.py
```diff
@@ -162,58 +23,84 @@ def get_speculative_sampling_kernel(
...
 
     dynamic_symbols = []
 
     constraints: list[tkw.Constraint] = [
         tkw.HardwareConstraint(
             threads_per_wave=64,
+            waves_per_block=(1, 1, 1),
             vector_shapes={
-                NUM_DRAFT_TOKENS: num_draft_tokens,
+                NUM_DRAFT_TOKENS: 0,
```
harsh-nod, line 62: why did these change?

wave_lang/kernel/wave/templates/speculative_decoding.py
```diff
@@ -543,12 +435,56 @@ def else_():
             outer_done,
         ) = outer_loop
 
-        tkw.write(
-            last_accepted_retrieve_idx,
-            last_accepted_retrieve_idx_vec,
-            elements_per_thread=1,
-        )
-        tkw.write(cur_prob_offset, cur_prob_offset_vec, elements_per_thread=1)
         tkw.write(num_accepted_tokens, accept_token_num, elements_per_thread=1)
```
harsh-nod, line 438: Do we need to write num_accepted_tokens ?

harsh-nod (changes requested):
Overall, looks good, there are just a couple of  questions about the changes.

wave_lang/kernel/wave/templates/speculative_decoding.py
xintin: There is an assertion in the test script. It is mutable, hence we need to write it.

lit_tests/kernel/wave/speculative_decoding.py
xintin: Oh, I restructured it, so I omitted it since it was already matched once above.
Do you suggest including it, or is it okay this way?
I just want to keep it minimal, but still include all representative cases, hence added a few more matches below.

harsh-nod: Can you keep the while and if to test the structure?

wave_lang/kernel/wave/templates/speculative_decoding.py
harsh-nod: Add a comment explaining why this is 0 and a TODO that this has to be modified when we switch to the block scan op.

PR #76, opened 2025-07-23
added XCD reordering to template and tweaked PingPong Schedule

bodhisaha:
Will soon modify the documentation about workgroup reordering to reflect the addition of XCD reordering.


raikonenfnu:
@bodhisaha Looks like LIT test is broken, overall code looks good! lets fix and land.

PR #77, opened 2025-07-23
SchedulingType.FOUR_STAGE: GEMM Full Software Pipelining with Initiation Interval 1, Via Multibuffering

SourishW:
Tested this new pipelined schedule on various tests in wave_gemm_test.py. Also modified wave_gemm_test.py correspondingly. 

Here is a representation of the type of schedule generated from a (GR, LW, LR, MMA) GEMM kernel:
```

a_0 = READ_GLOBAL 0

WRITE_SHARED a_0 SM[0]
a_1 = READ_GLOBAL 1
lds_barrier

b_0 = READ_SHARED SM[0]
WRITE_SHARED a_1 SM[1]
a_2 = READ_GLOBAL 2
lds_barrier

for i = 0 to N -3:
    COMPUTE b_i
    b_{i+1} = READ_SHARED SM[i+1 %2]
    // Work in progress is removing the immediate subsequent lds_barrier
    (lds_barrier)
    WRITE_SHARED a_{i+2} SM[i%2]
    a_{i+3} = READ_GLOBAL k+3
    lds_barrier

COMPUTE b_{n-2}
b_{n-1} = READ_SHARED SM[n-1 %2]
WRITE_SHARED a_{n} SM[n % 2]

COMPUTE b_{n-1}
b_{n} = READ_SHARED SM[n %2]

COMPUTE b_n
```

Thank you in advance for reviewing

### DIFF
--- a/lit_tests/kernel/wave/gemm.py
+++ b/lit_tests/kernel/wave/gemm.py
@@ -1274,7 +1274,7 @@ def test_gemm_prefetch():
 
 
 @run_test
-def test_gemm_gemm_four_stage():
+def test_gemm_four_stage():
     constraints: list[tkw.Constraint] = [tkw.WorkgroupConstraint(M, BLOCK_M, 0)]
     constraints += [tkw.WorkgroupConstraint(N, BLOCK_N, 1)]
     constraints += [tkw.TilingConstraint(K, BLOCK_K)]
@@ -1329,7 +1329,7 @@ def test_gemm_gemm_four_stage():
             SHUFFLE_UNITS: 8,
         },
         canonicalize=True,
-        schedule=SchedulingType.GEMM_FOUR_STAGE,
+        schedule=SchedulingType.FOUR_STAGE,
         use_scheduling_barriers=True,
         compile_to_mlir=True,
         use_multi_buffering=True,
--- a/lit_tests/kernel/wave/gemm.py
+++ b/lit_tests/kernel/wave/gemm.py
@@ -1274,6 +1274,102 @@ def test_gemm_prefetch():
 
 
 @run_test
+def test_gemm_gemm_four_stage():
+    constraints: list[tkw.Constraint] = [tkw.WorkgroupConstraint(M, BLOCK_M, 0)]
+    constraints += [tkw.WorkgroupConstraint(N, BLOCK_N, 1)]
+    constraints += [tkw.TilingConstraint(K, BLOCK_K)]
+    constraints += [tkw.WaveConstraint(M, BLOCK_M / 2)]
+    constraints += [tkw.WaveConstraint(N, BLOCK_N / 2)]
+
+    constraints += [
+        tkw.HardwareConstraint(
+            threads_per_wave=64,
+            mma_type=tkw.MMAType.F32_16x16x16_F16,
+        )
+    ]
+
+    @tkw.wave(constraints)
+    def gemm_four_stage(
+        a: tkl.Memory[M, K, ADDRESS_SPACE, tkl.f16],
+        b: tkl.Memory[N, K, ADDRESS_SPACE, tkl.f16],
+        c: tkl.Memory[M, N, ADDRESS_SPACE_0, tkl.f32],
+    ):
+        c_reg = tkl.Register[M, N, tkl.f32](0.0)
+
+        @tkw.iterate(K, init_args=[c_reg])
+        def repeat(acc: tkl.Register[M, N, tkl.f32]) -> tkl.Register[M, N, tkl.f32]:
+            a_reg = tkw.read(a)
+            b_reg = tkw.read(b)
+            acc = tkw.mma(a_reg, b_reg, acc)
+            return acc
+
+        tkw.write(repeat, c)
+
+    options = WaveCompileOptions(
+        subs={
+            M: 128,
+            N: 128,
+            K: 256,
+            BLOCK_M: 64,
+            BLOCK_N: 64,
+            BLOCK_K: 32,
+            ADDRESS_SPACE: SHARED_ADDRESS_SPACE,
+            ADDRESS_SPACE_0: GLOBAL_ADDRESS_SPACE,
+            READ_SHARED_DELAY: 1,
+            WRITE_SHARED_DELAY: 1,
+            READ_GLOBAL_DELAY: 2,
+            WRITE_GLOBAL_DELAY: 2,
+            MMA_DELAY: 1,
+            VALU_DELAY: 1,
+            SHUFFLE_DELAY: 1,
+            SHARED_MEMORY_UNITS: 4,
+            GLOBAL_MEMORY_UNITS: 4,
+            MMA_UNITS: 4,
+            VALU_UNITS: 8,
+            SHUFFLE_UNITS: 8,
+        },
+        canonicalize=True,
+        schedule=SchedulingType.GEMM_FOUR_STAGE,
+        use_scheduling_barriers=True,
+        compile_to_mlir=True,
+    )
+
+    gemm_four_stage = wave_compile(options, gemm_four_stage)
+    print(gemm_four_stage.asm)
+    # CHECK-LABEL: func.func @gemm_four_stage
+    # Test multibuffering: verify shared memory views are correctly allocated
+    # CHECK: %[[ALLOC:.*]] = memref.alloc() : memref<18432xi8
+    # CHECK: %[[VIEW0:.*]] = memref.view %[[ALLOC]][%c0][] : memref<18432xi8
+    # CHECK: %[[VIEW1:.*]] = memref.view %[[ALLOC]][%c9216][] : memref<18432xi8
+
+    # Prologue
+    # Verify prologue stores to shared memory
+    # CHECK: %[[STORE_IDX:.*]] = affine.apply #map[0-9]+()[%thread_id_x, %thread_id_y]
+    # CHECK: vector.store %{{.*}}, %[[VIEW1]][%[[STORE_IDX]], %{{.*}}] : memref<128x36xf16
+    # CHECK: vector.store %{{.*}}, %[[VIEW0]][%[[STORE_IDX]], %{{.*}}] : memref<128x36xf16
+
+    # Verify prologue loads from shared memory
+    # CHECK: %[[LOAD_IDX1:.*]] = affine.apply #map[0-9]+()[%thread_id_x, %thread_id_y]
+    # CHECK: %[[LOAD_IDX2:.*]] = affine.apply #map[0-9]+()[%thread_id_x]
+    # CHECK: vector.load %[[VIEW0]][%[[LOAD_IDX1]], %[[LOAD_IDX2]]] : memref<128x36xf16
+
+    # Main Loop:
+    # Verify Pipelined Loop, iter_args should contain vector values from prologue
+    # CHECK: scf.for %[[ARG3:.*]] = %c0 to %c5 step %c1 iter_args(%{{.*}} = %cst, %{{.*}} = %cst, %{{.*}} = %cst, %{{.*}} = %cst, %{{.*}} = %{{.*}}, %{{.*}} = %{{.*}}, %{{.*}} = %{{.*}}, %{{.*}} = %{{.*}}, %{{.*}} = %{{.*}}, %{{.*}} = %{{.*}}, %{{.*}} = %{{.*}}, %{{.*}} = %{{.*}}, %{{.*}} = %{{.*}}, %{{.*}} = %{{.*}})
+
+    # Verify MFMA exists
+    # CHECK: amdgpu.mfma %{{.*}} * %{{.*}} + %{{.*}}
+
+    # Verify that affine maps using iter_arg %[[ARG3]] are used for pipelined indexing
+    # CHECK: %[[PIPE_IDX1:.*]] = affine.apply #map[0-9]+()[%thread_id_x, %[[ARG3]], %thread_id_y]
+    # CHECK: vector.load %[[VIEW0]][%[[PIPE_IDX1]], %{{.*}}] : memref<128x36xf16
+
+    # Verify that stores use iter arg for indexing as well
+    # CHECK: %[[STORE_PIPE_IDX:.*]] = affine.apply #map[0-9]+()[%thread_id_x, %thread_id_y, %[[ARG3]]]
+    # CHECK: vector.store %{{.*}}, %[[VIEW1]][%[[STORE_PIPE_IDX]], %{{.*}}] : memref<128x36xf16
+
+
+@run_test
 def test_dynamic_gemm_pipelined():
     constraints: list[tkw.Constraint] = [tkw.WorkgroupConstraint(M, BLOCK_M, 0)]
     constraints += [tkw.WorkgroupConstraint(N, BLOCK_N, 1)]
--- a/lit_tests/kernel/wave/gemm.py
+++ b/lit_tests/kernel/wave/gemm.py
@@ -1332,7 +1332,6 @@ def test_gemm_four_stage():
         schedule=SchedulingType.FOUR_STAGE,
         use_scheduling_barriers=True,
         compile_to_mlir=True,
-        use_multi_buffering=True,
         multi_buffer_count=2,
     )
 
--- a/lit_tests/kernel/wave/gemm.py
+++ b/lit_tests/kernel/wave/gemm.py
@@ -1332,6 +1332,8 @@ def test_gemm_gemm_four_stage():
         schedule=SchedulingType.GEMM_FOUR_STAGE,
         use_scheduling_barriers=True,
         compile_to_mlir=True,
+        use_multi_buffering=True,
+        multi_buffer_count=2,
     )
 
     gemm_four_stage = wave_compile(options, gemm_four_stage)
@@ -1344,13 +1346,13 @@ def test_gemm_gemm_four_stage():
 
     # Prologue
     # Verify prologue stores to shared memory
-    # CHECK: %[[STORE_IDX:.*]] = affine.apply #map[0-9]+()[%thread_id_x, %thread_id_y]
+    # CHECK: %[[STORE_IDX:.*]] = affine.apply #[[MAP_STORE:.*]]()[%thread_id_x, %thread_id_y]
     # CHECK: vector.store %{{.*}}, %[[VIEW1]][%[[STORE_IDX]], %{{.*}}] : memref<128x36xf16
     # CHECK: vector.store %{{.*}}, %[[VIEW0]][%[[STORE_IDX]], %{{.*}}] : memref<128x36xf16
 
     # Verify prologue loads from shared memory
-    # CHECK: %[[LOAD_IDX1:.*]] = affine.apply #map[0-9]+()[%thread_id_x, %thread_id_y]
-    # CHECK: %[[LOAD_IDX2:.*]] = affine.apply #map[0-9]+()[%thread_id_x]
+    # CHECK: %[[LOAD_IDX1:.*]] = affine.apply #[[MAP_LOAD1:.*]]()[%thread_id_x, %thread_id_y]
+    # CHECK: %[[LOAD_IDX2:.*]] = affine.apply #[[MAP_LOAD2:.*]]()[%thread_id_x]
     # CHECK: vector.load %[[VIEW0]][%[[LOAD_IDX1]], %[[LOAD_IDX2]]] : memref<128x36xf16
 
     # Main Loop:
@@ -1361,11 +1363,11 @@ def test_gemm_gemm_four_stage():
     # CHECK: amdgpu.mfma %{{.*}} * %{{.*}} + %{{.*}}
 
     # Verify that affine maps using iter_arg %[[ARG3]] are used for pipelined indexing
-    # CHECK: %[[PIPE_IDX1:.*]] = affine.apply #map[0-9]+()[%thread_id_x, %[[ARG3]], %thread_id_y]
+    # CHECK: %[[PIPE_IDX1:.*]] = affine.apply #[[MAP_PIPE1:.*]]()[%thread_id_x, %[[ARG3]], %thread_id_y]
     # CHECK: vector.load %[[VIEW0]][%[[PIPE_IDX1]], %{{.*}}] : memref<128x36xf16
 
     # Verify that stores use iter arg for indexing as well
-    # CHECK: %[[STORE_PIPE_IDX:.*]] = affine.apply #map[0-9]+()[%thread_id_x, %thread_id_y, %[[ARG3]]]
+    # CHECK: %[[STORE_PIPE_IDX:.*]] = affine.apply #[[MAP_STORE_PIPE:.*]]()[%thread_id_x, %thread_id_y, %[[ARG3]]]
     # CHECK: vector.store %{{.*}}, %[[VIEW1]][%[[STORE_PIPE_IDX]], %{{.*}}] : memref<128x36xf16
 
 
--- a/wave_lang/kernel/wave/compile_options.py
+++ b/wave_lang/kernel/wave/compile_options.py
@@ -25,6 +25,8 @@ class WaveCompileOptions:
     # === Scheduling options ===
     schedule: bool = SchedulingType.NONE
     use_scheduling_barriers: bool = False
+    use_multi_buffering: bool = False
+    multi_buffer_count: int = 2  # if no buffer count specified do double buffer
 
     # === Runtime options ===
     kernel_launch_info: KernelLaunchInfo = field(default_factory=KernelLaunchInfo)
--- a/wave_lang/kernel/wave/compile_options.py
+++ b/wave_lang/kernel/wave/compile_options.py
@@ -25,9 +25,8 @@ class WaveCompileOptions:
     # === Scheduling options ===
     schedule: bool = SchedulingType.NONE
     use_scheduling_barriers: bool = False
-    multi_buffer_count: Optional[int] = (
-        None  # None if no buffer count specified else 2 and up
-    )
+    # None if no buffer count specified else 2 and up
+    multi_buffer_count: Optional[int] = None
 
     # === Runtime options ===
     kernel_launch_info: KernelLaunchInfo = field(default_factory=KernelLaunchInfo)
--- a/wave_lang/kernel/wave/compile_options.py
+++ b/wave_lang/kernel/wave/compile_options.py
@@ -25,8 +25,9 @@ class WaveCompileOptions:
     # === Scheduling options ===
     schedule: bool = SchedulingType.NONE
     use_scheduling_barriers: bool = False
-    use_multi_buffering: bool = False
-    multi_buffer_count: int = 2  # if no buffer count specified do double buffer
+    multi_buffer_count: Optional[int] = (
+        None  # None if no buffer count specified else 2 and up
+    )
 
     # === Runtime options ===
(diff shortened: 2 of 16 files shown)


wave_lang/kernel/wave/scheduling/gemm_four_stage_pipelined_scheduling.py
```diff
@@ -0,0 +1,144 @@
...
+from .resources import Operation, get_custom_operation_type
+from enum import Enum
+
+
+class GemmFourStageStage(Enum):
+    GLOBAL_LOAD = 0
+    LOCAL_STORE = 1
+    LOCAL_LOAD = 2
+    COMPUTE = 3
+    DOESNT_MATTER = -1
```
raikonenfnu, line 13: Any better name aside from DOESNT_MATTER? np if not

wave_lang/kernel/wave/scheduling/gemm_four_stage_pipelined_scheduling.py
```diff
@@ -0,0 +1,144 @@
...
+        Returns the initiation interval of the schedule.
+        """
+        return self._initiation_interval
+
+    @property
+    def num_stages(self) -> int:
+        """
+        Returns the number of stages in the kernel of the pipelined loop.
+        """
+        return 4
```
raikonenfnu, line 144: Why not use the mathematical formula? The math formula is important because it will only generate `4`, if we are able to do scheduling properly

wave_lang/kernel/wave/scheduling/multi_buffering.py
```diff
@@ -24,9 +24,36 @@
...
+            new_distributed_shape.append(original_buffer.distributed_shape[i])
+        else:
+            # Increase non-reduction dimensions
+            new_shape.append(dim * buffer_count)
+            new_distributed_shape.append(
+                original_buffer.distributed_shape[i] * buffer_count
+            )
+
+    original_buffer.update_arg(0, tuple(new_shape))
+    original_buffer.update_arg(1, tuple(new_distributed_shape))
```
raikonenfnu, line 53: I think this is fine for this PR, but in future PR, maybe would be good to have the expanded dimension the slowest/first dimension instead of the fastest/last dimension. CC: @harsh-nod

wave_lang/kernel/wave/scheduling/gemm_four_stage_pipelined_scheduling.py
SourishW: Oh I forgot about that, I was using a temporary name until I thought of a better one, I will fix it

wave_lang/kernel/wave/scheduling/gemm_four_stage_pipelined_scheduling.py
```diff
@@ -0,0 +1,144 @@
...
+
+    """
+
+    def __init__(
+        self,
+        graph: fx.Graph,
+        edges: list[Edge],
+        resources: list[int],
+    ) -> None:
+        # assert False
```
raikonenfnu, line 93: dead code here?

wave_lang/kernel/wave/scheduling/gemm_four_stage_pipelined_scheduling.py
SourishW: Sure, will fix

wave_lang/kernel/wave/scheduling/loop_reconstruction.py
```diff
@@ -42,7 +48,28 @@ class PipelineStage(Enum):
     EPILOGUE = 2
 
 
+def _heuristically_multi_buffer_shared_memory(
```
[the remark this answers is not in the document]
SourishW: The modulo case handles this multi_buffering of shared memory after all operations have been already scheduled. IE it applies this almost as a pass afterwards, to find all shared memory and multibuffer it. 

I originally tried to do a similar approach and put it next to MODULO_MULTI_BUFFERED code after the fact, but was having issues finding which loop iterations needed which induction variable, since they're all offset, etc. Including it in loop reconstruction enables me to more conveniently have access to induction_variable that is different for each operation. (ie the induction_variable in loop_reconstruction.py doesnt mean the same thing as induction_variable in multi_buffering.py)

Here since we know in advance which iteration we are on, I decided to update it while generating the operations themselves. It would have been a little trickier to do it the other way. 

Actually btw we can even move modulo code into loop_reconstruction.py later as well.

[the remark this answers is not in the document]
raikonenfnu: > which loop iterations needed which induction variable

Couldn't we just use `from ..utils.general_utils import get_induction_variable; induction_variable = get_induction_variable(reduction, constraints)`?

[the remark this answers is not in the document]
SourishW: Those current_induction_variables are a little bit more complicated than just the one you get from reduction and constraints I think. Its different for prologue kernel epilogue. 

If the function add_nodes_by_schedule is called in prologue code section, then current_induction_variables become the prologue 0, 1, 2, etc etc. If called in kernel, then current_induction_variable becomes ARGK+0, 1 etc. If called in epilogue its K-3, K-2, etc

Actually if you trace up code in kernel it actually gets the original induction variable from the exact thing you mention. So we do use that.

raikonenfnu:
you have git conflicts btw, @SourishW

lit_tests/kernel/wave/gemm.py
```diff
@@ -1273,6 +1273,101 @@ def repeat(acc: tkl.Register[M, N, tkl.f32]) -> tkl.Register[M, N, tkl.f32]:
...
+    # CHECK: %[[VIEW1:.*]] = memref.view %[[ALLOC]][%c9216][] : memref<18432xi8
+
+    # Prologue
+    # Verify prologue stores to shared memory
+    # CHECK: %[[STORE_IDX:.*]] = affine.apply #map2()[%thread_id_x, %thread_id_y]
+    # CHECK: vector.store %{{.*}}, %[[VIEW1]][%[[STORE_IDX]], %{{.*}}] : memref<128x36xf16
+    # CHECK: vector.store %{{.*}}, %[[VIEW0]][%[[STORE_IDX]], %{{.*}}] : memref<128x36xf16
+
+    # Verify prologue loads from shared memory
+    # CHECK: %[[LOAD_IDX1:.*]] = affine.apply #map4()[%thread_id_x, %thread_id_y]
```
harsh-nod, line 1352: I don't think you should match to the map number exactly. The ordering of the maps and hence the map numbering is arbitrary so it could change.

wave_lang/kernel/wave/scheduling/gemm_four_stage_pipelined_scheduling.py
```diff
@@ -0,0 +1,148 @@
...
+        b_0 = READ_SHARED SM[0]
+        WRITE_SHARED a_1 SM[1]
+        a_2 = READ_GLOBAL 2
+
+
+        for i = 0 to N -3:
+            COMPUTE b_i
+            b_{i+1} = READ_SHARED SM[i+1 %2]
+            WRITE_SHARED a_{i+2} SM[i%2]
+            a_{i+3} = READ_GLOBAL k+3
```
harsh-nod, line 74: Should this be READ_GLOBAL i  + 3 ? With this schedule, there should be no barriers inside the loop right?

wave_lang/kernel/wave/scheduling/gemm_four_stage_pipelined_scheduling.py
```diff
@@ -0,0 +1,148 @@
...
+    Operation.READ_GLOBAL: GemmFourStageStage.GLOBAL_LOAD,
+    Operation.MMA: GemmFourStageStage.COMPUTE,
+    Operation.NOOP: GemmFourStageStage.SCHEDULING_NOOP,
+    Operation.VALU: GemmFourStageStage.COMPUTE,
+    Operation.SHUFFLE: GemmFourStageStage.COMPUTE,
+    Operation.WRITE_GLOBAL: GemmFourStageStage.COMPUTE,
+}
+
+
+def get_scheduling_stage(op: fx.Node) -> GemmFourStageStage:
```
harsh-nod, line 39: Should the return type be Operation? Also seems like you could reuse this function from prefetch_scheduling.py?

wave_lang/kernel/wave/scheduling/loop_reconstruction.py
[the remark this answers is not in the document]
harsh-nod: So just trying to understand why is this different from the code that already exists in multi_buffering.py because there the induction variables can be remapped without doing this inside loop reconstruction?

wave_lang/kernel/wave/scheduling/multi_buffering.py
harsh-nod: Yes, I would agree that it should be the slowest dimension.

harsh-nod (changes requested):

wave_lang/kernel/wave/scheduling/loop_reconstruction.py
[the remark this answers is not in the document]
SourishW: I forgot to mention about prologue and epilogue things, thats actually a bigger reason really:

The problem is you can't easily find the prologue and epilogue sections to multibuffer it in multi_buffering.py without adding some really weird attributes or some graph analysis to help identification of these segments later. 

Actually its hard so the current MODULO_MULTI_BUFFERED code doesn't really bother to look at prologue or epilogue either. 

It just finds an iterate block and multibuffers it, not caring about prologue and epilogue pipelining

image

You can see evidence of this in the IR too:

image

This whole approach of doing some pipelining then trying tracing your graph to reverse engineer your own pipeline to find prologue kernel and epilogue didn't make sense to me, so I thought why not do your multibuffering while you know what you're producing? Let me know if you have any comments on this @harsh-nod 

Will try to move code for MODULO_MULTI_BUFFERING too soon if thats ok.

lit_tests/kernel/wave/gemm.py
SourishW: I'm sorry for making this silly mistake, I forgot to replace this matching to affine map with matching to anything when I was replacing with regex. Will try to be more careful with code next time

wave_lang/kernel/wave/scheduling/gemm_four_stage_pipelined_scheduling.py
SourishW: You still need a barrier in the loop, just not two barriers and not in between the READ_SHARED and WRITE_SHARED segment. 

With double buffering Loop iteration i's write to SM[i%2] has to complete before loop iteration i+1's read from SM[(i+1)+1 %2] ie SM [i%2].

wave_lang/kernel/wave/scheduling/scheduler_utils.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def get_scheduling_stage(
+    op: fx.Node, operation_stage_table: dict[ScheduleStage:int]
+) -> ScheduleStage:
+    op_ty = get_custom_operation_type(get_custom(op))
+    assert op_ty is not None, f"get_custom_operation_type returned None for {op}"
+    if op_ty not in operation_stage_table:
+        raise NotImplementedError(f"Cannot find {op_ty} in operation_stage_table")
+    return operation_stage_table[op_ty]
```
martin-luecke, line 18: The type annotation of `operation_stage_table` is wrong. In [gemm_four_stage_pipelined_scheduling.py](https://github.com/iree-org/wave/pull/77/files#diff-1faec07204fed5bf51618ea52dcf5dd6916eb1bbdf967ec5265285e6cb2d31e6) it is `dict[Operation, GemmFourStageStage]`.
If you want to support different types of `operation_stage_table`s in this utils file you could use the following approach:
```python
from typing import TypeVar
from enum import Enum

ScheduleStage = TypeVar('ScheduleStage', bound=Enum)

def get_scheduling_stage(
    op: fx.Node, 
    operation_stage_table: dict[Operation, ScheduleStage]
) -> ScheduleStage:
```
We know the key is always the `Operation` Enum so we can directly use that. By using `TypeVar` this specifies that we expect a subclass of Enum as value in the dict and, crucially, the same subclass as return type of this function.

This also let's you drop your alias `ScheduleStage = Enum  # alias`

wave_lang/kernel/wave/scheduling/gemm_four_stage_pipelined_scheduling.py
```diff
@@ -0,0 +1,112 @@
...
+        for node in sorted_nodes:
+            node_stage = get_scheduling_stage(node, operation_stage_table)
+            next_stage = current_stage.next()
+            if node_stage in [current_stage, GemmFourStageStage.SCHEDULING_NOOP]:
+                schedule[node] = current_stage.value
+            elif node_stage == next_stage:
+                schedule[node] = next_stage.value
+                current_stage = next_stage
+            else:
+                # Node do not move contigously through stages.
```
martin-luecke, line 100:
```suggestion
                # Node does not move contiguously through stages.
```

wave_lang/kernel/wave/scheduling/gemm_four_stage_pipelined_scheduling.py
```diff
@@ -0,0 +1,112 @@
...
+        COMPUTE b_n
+
+    """
+
+    def mega_pipelined_scheduling(self, graph: fx.Graph, edges: list[Edge]):
+        """
+        Classify node to different stages. Based on it's stage,
+        program schedules clock for each node. This function also checks
+        that sorted node "contiguously" move between stages.
+        """
```
martin-luecke, line 86: Why is this called "mega_pipelined"? What do you mean with "Based on it's stage, program schedules clock for each node." 
Could you please rework the name and comment to reflect better what is happening here.
This function is missing a return type annotation.

wave_lang/kernel/wave/scheduling/gemm_four_stage_pipelined_scheduling.py
```diff
@@ -0,0 +1,112 @@
...
+    """
+    GEMM Four Stage Pipelined Scheduler
+
+    Convert vanilla schedule of:
+        for i = 0 to N:
+            a = READ_GLOBAL i
+            WRITE_SHARED a
+            barrier
+            b = READ_SHARED
+            COMPUTE b
```
martin-luecke, line 48: Is this specific to GEMMs only? Does it make sense to simply call it `FourStageScheduler`?

wave_lang/kernel/wave/scheduling/gemm_four_stage_pipelined_scheduling.py
```diff
@@ -0,0 +1,112 @@
...
+    COMPUTE = 3
+    SCHEDULING_NOOP = -1
+
+    # Helper function to get next stage from the current.
+    # If at stage 3 returns itself to prevent crash
+    # since it is final stage.
+    def next(self):
+
+        if self.value == 3:
+            return GemmFourStageStage(3)
```
martin-luecke, line 21:
```suggestion
            return self
```

wave_lang/kernel/wave/scheduling/gemm_four_stage_pipelined_scheduling.py
```diff
@@ -0,0 +1,112 @@
...
+    GLOBAL_LOAD = 0
+    LOCAL_STORE = 1
+    LOCAL_LOAD = 2
+    COMPUTE = 3
+    SCHEDULING_NOOP = -1
+
+    # Helper function to get next stage from the current.
+    # If at stage 3 returns itself to prevent crash
+    # since it is final stage.
+    def next(self):
```
martin-luecke, line 18: This needs type annotations. 
For Python3.11+ we could use `def next(self) → Self:`, otherwise  `def next(self) → "GemmFourStageStage":`

martin-luecke (changes requested):
Thank you for the PR!

Could you explain whether this is really only specific to Gemms or whether it might be usable for other kernels. In the case it is, I'd advocate for simply calling it `four_stage` everywhere.

Could you also explain which parts of the current multi-buffering implementation you think we could/should move to `loop_reconstruction`?
I think it might be worthwhile thinking about moving your `_heuristically_multi_buffer_shared_memory` to multi_buffering.py and simply calling it from loop_reconstruction (which already is a huge file and does more than "reconstructing" a loop). 


Otherwise, I left a few questions and suggestions.

wave_lang/kernel/wave/scheduling/scheduler_utils.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class BaseScheduler:
+    def __init__(
+        self,
+        graph: fx.Graph,
+        edges: list[Edge],
+        resources: list[int],
+    ) -> None:
+        # assert False
```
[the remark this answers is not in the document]
SourishW: Removed

wave_lang/kernel/wave/scheduling/multi_buffering.py
```diff
@@ -75,6 +121,7 @@ def _multi_buffer_memory_location(
     write_nodes: list[Write],
     reduction_axis: IndexSymbol,
     buffer_count: int,
+    scheduling_type: SchedulingType,
```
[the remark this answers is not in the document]
SourishW: Just removed it thanks

wave_lang/kernel/wave/scheduling/loop_reconstruction.py
```diff
@@ -63,7 +68,28 @@ def update_node_index(
         node.update_arg("dst_index", update_index(dst_idx, subs))
 
 
+def _heuristically_multi_buffer_shared_memory(
+    new_node: CustomOp,
+    induction_variable: IndexSymbol,
+    reduction: Iterate,
+    node: fx.Node,
+):
```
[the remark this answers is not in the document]
SourishW: Done

wave_lang/kernel/wave/scheduling/scheduler_utils.py
SourishW: Thanks for your suggestion, I did this with the TypeVar as you suggested

wave_lang/kernel/wave/scheduling/gemm_four_stage_pipelined_scheduling.py
SourishW: It was the old name I gave to scheduler type, I cleaned up the language to 

Classify node to different stages. Based on its stage,
        program schedules the node to a specific cycle for  each node.
        This function also checks that the sorted nodes move contiguously through
        expected stages.

wave_lang/kernel/wave/scheduling/gemm_four_stage_pipelined_scheduling.py
SourishW: Done

wave_lang/kernel/wave/scheduling/gemm_four_stage_pipelined_scheduling.py
SourishW: I used the latter

wave_lang/kernel/wave/scheduling/gemm_four_stage_pipelined_scheduling.py
SourishW: I fixed this btw forgot to reply

lit_tests/kernel/wave/gemm.py
SourishW: Fixed this and matched to anything. 
` # CHECK: vector.store %{{.*}}, %[[VIEW0]][%[[STORE_IDX]], %{{.*}}] : memref<128x36xf16

    # Verify prologue loads from shared memory
    # CHECK: %[[LOAD_IDX1:.*]] = affine.apply #[[MAP_LOAD1:.*]]()[%thread_id_x, %thread_id_y]`

SourishW:
@raikonenfnu @harsh-nod @martin-luecke Thank you so much for your reviews, 
I have addressed the comments and made appropriate changes. 

@martin-luecke I was also able to make the multibuffering strategy scheduling type agnostic so that we don't need different logic for each and every single new scheduling strategy we introduce. 

This is ported your original code for multibuffering out, and this makes it so that we multibuffer only in one place in a convenient location for all scheduling type, not in 5 uniquely for each and every future scheduling type. It should work seamlessly for any new scheduling type introduced. 

Now instead of specifying `SchedulingType.MODULO_MULTI_BUFFERED`, you specify `SchedulingType.MODULO` and turn on the wave compile flag for `use_multi_buffering=True`, optionally specifying `multi_buffer_count` (default of 2)

@raikonenfnu let me know if you want me to move the buffer indexing based on fastest dimension to this PR or a future PR. 

I will work on a PR for indexing analysis now. If you want me to work on fastest dimension, I will work on that simultaneously and make a combined PR

martin-luecke (approved):
LGTM after implementing a check for the pre-conditions we have for this scheduling aproach

Thank you for the changes.

wave_lang/kernel/wave/scheduling/gemm_four_stage_pipelined_scheduling.py
raikonenfnu: That's good, we actually have that check already implemented in schedule_reordering. But probably better to move it into this and also prefetch scheduler.

https://github.com/iree-org/wave/blob/main/wave_lang/kernel/wave/schedule_reordering.py#L164-L168

SourishW: I will rename to FourStageScheduler and add the precondition

wave_lang/kernel/wave/scheduling/scheduler_utils.py
```diff
@@ -0,0 +1,49 @@
...
+        Returns the initiation interval of the schedule.
+        """
+        return self._initiation_interval
+
+    @property
+    def num_stages(self) -> int:
+        """
+        Returns the number of stages in the kernel of the pipelined loop.
+        """
+        max_cycle = max([t + 1 for t in self.schedule.values()])
```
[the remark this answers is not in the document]
SourishW: Done (in 10 mins)

wave_lang/kernel/wave/scheduling/gemm_four_stage_pipelined_scheduling.py
SourishW: Done

wave_lang/kernel/wave/scheduling/multi_buffering.py
```diff
@@ -41,104 +88,33 @@ def multi_buffer(trace: CapturedTrace):
...
 
 
-def _multi_buffer_memory_location(
-    trace: CapturedTrace,
-    original_buffer: CustomOp,
-    read_nodes: list[Read],
-    write_nodes: list[Write],
-    reduction_axis: IndexSymbol,
-    buffer_count: int,
+def _increase_dimensions_for_multibuffering(
```
Hardcode84, line 111: `_increase_dimensions_for_multibuffering` currently multiplies each non-reduction dim by `buffer_count` which will cause issues if we have more than 1, I think we should multiply only one dim.

martin-luecke:
@Hardcode84 @SourishW I advocate for merging this PR with the current approach. We can investigate possibly switching to different multi-buffering buffer management afterwards. I remember looking at your suggestion of increasing the loop carried vars but had some problems I was not happy with. Possibly increased register pressure 🤔
The best hipblaslt kernels use this strategy so I went with it.

raikonenfnu:
> @Hardcode84 @SourishW I advocate for merging this PR with the current approach. We can investigate possibly switching to different multi-buffering buffer management afterwards. I remember looking at your suggestion of increasing the loop carried vars but had some problems I was not happy with. Possibly increased register pressure 🤔 The best hipblaslt kernels use this strategy so I went with it.

Agreed, but right now it still has failed schedule with the fully pipelined w/ global load to LDS

wave_lang/kernel/wave/compile_options.py
```diff
@@ -25,6 +25,8 @@ class WaveCompileOptions:
     # === Scheduling options ===
     schedule: bool = SchedulingType.NONE
     use_scheduling_barriers: bool = False
+    use_multi_buffering: bool = False
+    multi_buffer_count: int = 2  # if no buffer count specified do double buffer
```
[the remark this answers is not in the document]
SourishW: Done

SourishW:
> > @Hardcode84 @SourishW I advocate for merging this PR with the current approach. We can investigate possibly switching to different multi-buffering buffer management afterwards. I remember looking at your suggestion of increasing the loop carried vars but had some problems I was not happy with. Possibly increased register pressure 🤔 The best hipblaslt kernels use this strategy so I went with it.
> 
> Agreed, but right now it still has failed schedule with the fully pipelined w/ global load to LDS

No more failed schedule with global load to LDS!

SourishW:
> Please fix the DCO

I just fixed it, sorry I missed this

wave_lang/kernel/wave/scheduling/four_stage_pipelined_scheduling.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+class FourStageStage(Enum):
+    GLOBAL_LOAD = auto()
+    LOCAL_STORE = auto()
+    LOCAL_LOAD = auto()
+    COMPUTE = auto()
+    SCHEDULING_NOOP = -1
+
+    # Helper function to get next stage from the current.
```
Hardcode84, line 23: this comment is no longer relevant

wave_lang/kernel/wave/scheduling/multi_buffering.py
```diff
@@ -41,114 +107,45 @@ def multi_buffer(trace: CapturedTrace):
     reduction_axis = get_custom(reductions[0]).axis
 
     # Find reads and writes operating on shared memory
-    reads = []
-    writes = []
+    reads_and_writes = list()
```
Hardcode84, line 110: `reads_and_writes = []`

wave_lang/kernel/wave/compile_options.py
```diff
@@ -25,6 +25,9 @@ class WaveCompileOptions:
     # === Scheduling options ===
     schedule: bool = SchedulingType.NONE
     use_scheduling_barriers: bool = False
+    multi_buffer_count: Optional[int] = (
```
Hardcode84, line 28: weird formatting, move comment to separate line

Hardcode84 (approved):

wave_lang/kernel/wave/compile_options.py
SourishW: Just added to the kernel hash

wave_lang/kernel/wave/scheduling/multi_buffering.py
```diff
@@ -23,10 +21,78 @@
...
+
+        # operation reads from / writes to, which we change below.
+        new_node_idx = get_dict_with_updated_key(
+            new_node_idx, dim, dim * multi_buffer_count
+        )
+        if is_read_write:
+            new_node.index = new_node_idx
+        else:
+            new_node.dst_index = new_node_idx
+            return
```
rohan-tan-bhowmik, line 78: Tiny tiny suggestion for clarity and brevity

```suggestion
        if not is_read_write:
            new_node.dst_index = new_node_idx
            return
            
        new_node.index = new_node_idx
```

paulzzy: LGTM

rohan-tan-bhowmik (approved):

PR #87, opened 2025-07-24
Adding scatter_max operation

adedespirlet:
This PR adds a new Wave operation: scatter_max. Its implementation mirrors that of scatter_add. It performs an element-wise maximum reduction from a source register into shared memory (LDS) using dynamic indices along a specified dimension. It supports both integer and floating-point data types and is lowered to memref.atomic_rmw using either maxs or maximumf, depending on the element type. 

Input arguments:
- src: a register containing the values to scatter
- index: a register containing the indices along dim where values should be written
- dim: the dimension along which scattering is performed
- memory: the target tensor in LDS where results are accumulated
- mapping: the index mapping representing mapping between sets of indices
- elements_per_thread: number of elements each thread loads

The mapping argument defaults to an identity IndexMapping, removing the need to pass it explicitly in common cases . Custom static (non-identity) index mappings (e.g transpositions) must be passed explicitly.

Implementation notes:
This op does not use the dynamic_val_mapping field of IndexMapping, unlike the read/write operations. The op iternally performs dynamic indexing, combining the index,dim and (optional) mapping to compute the final memory access coordinates. Potential Future Work: rewrite the operation handler and API to align with wave's write() operation implementation where in case of a scatter operation, index mapping expressions passed through are fully dynamic.
Tests are included in wave_e2e_test.py to validate the correctness of scatter_max.

Limitations:
Currently, scatter_max only works with a single wave. Using 2 or 4 waves results in incorrect behavior, especially when thread index to the same location. The operation supports multiple elements per thread provided that the non-scatter dimension is large enough (> elements_per_thread)

### DIFF
--- a/tests/kernel/wave/wave_e2e_test.py
+++ b/tests/kernel/wave/wave_e2e_test.py
@@ -2133,14 +2133,6 @@ def test_scatter_add(shape, elems_per_thread, request):
         tkw.WaveConstraint(N, BLOCK_N),
     ]
 
-    i = tkw.IndexMapping.iterator(0)
-    j = tkw.IndexMapping.iterator(1)
-    mapping = tkw.IndexMapping(
-        num_iterators=2,
-        inputs={M: i, N: j},
-        outputs={M: i, N: j},
-    )
-
     @tkw.wave(constraints)
     def test(
         a: tkl.Memory[M, N, GLOBAL_ADDRESS_SPACE, tkl.f32],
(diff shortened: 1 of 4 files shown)


tests/kernel/wave/wave_e2e_test.py
```diff
@@ -2264,3 +2264,107 @@ def test(
...
+        ),
+        tkw.WorkgroupConstraint(M, BLOCK_M, 0),
+        tkw.WorkgroupConstraint(N, BLOCK_N, 1),
+        tkw.WaveConstraint(M, BLOCK_M / 1),
+        tkw.WaveConstraint(N, BLOCK_N),
+    ]
+
+    i = tkw.IndexMapping.iterator(0)
+    j = tkw.IndexMapping.iterator(1)
+    mapping1 = tkw.IndexMapping(
```
ftynse, line 2305: Why not just `mapping`?

ftynse: Actually, do we even need this mapping?

ftynse (commented):
Could we add a test to lit_tests/codegen.py checking the IR we generate?

tests/kernel/wave/wave_e2e_test.py
adedespirlet: In the previous implementation it was required, but you're right , it's cleaner not to have to define an IndexMapping when it's just the identity. I’ve updated the implementation so that if no mapping is provided, it now defaults internally to an identity IndexMapping

PR #89, opened 2025-07-24
Python 3.13 Support

tgymnich:


ftynse (approved):
The change LGTM. We may want to be more careful with increasing the CI matrix, how long do these tests take to run

PR #91, opened 2025-07-24
add constraints to the cache key

ashay:
The `anonymize_constraints()` function has a bug, in that it updates the
constraints as desired, but it doesn't return the updated constraints.
At the same time, in the caller, we use the return value of the function
to create the cache key.  Consequently, prior to this patch, we always
used `None` as the constraint portion of the cache key, so even after
updating the constraints, Wave still retrieved a stale compiled kernel
from the cache.

Additionally, prior to this change, the uncached compilation path
initialized/updated the constraints, thus causing a cache miss even
though the kernel and the programmer-specified constraints didn't
change.  Luckily, the updates to the constraints seem independent of the
trace, so this patch moves the constraint updates _before_ the cache
lookup.

### DIFF
--- a/wave_lang/kernel/wave/cache.py
+++ b/wave_lang/kernel/wave/cache.py
@@ -117,7 +117,7 @@ def get_nested_functions(root_fn: Callable):
     return fn_list
 
 
-def anonymize_constraints(input_constraints: list[Constraint]):
+def anonymize_constraints(input_constraints: list[Constraint]) -> list[Constraint]:
     """
     Helper function to anonymize constraint S.T we can have the same generate
     hash before and after initializing constraints and induction variables.
@@ -137,6 +137,7 @@ def anonymize_constraints(input_constraints: list[Constraint]):
             constraint.wave_id = None
         else:
             continue
+    return processed_constraints
 
 
 class WaveCacheManager(object):
--- a/wave_lang/kernel/wave/compile.py
+++ b/wave_lang/kernel/wave/compile.py
@@ -215,6 +215,20 @@ def wave_compile(options: WaveCompileOptions, kernel: "LaunchableWave") -> WaveK
     cache_manager = None
     binary_path = None
 
+    # Create an indexing context and populate substitutions.
+    push(IndexingContext, IndexingContext())
+    idxc = IndexingContext.current()
+
+    # Make a copy of the substitutions to avoid mutating the original
+    # options.subs.
+    idxc.subs = copy(options.subs)
+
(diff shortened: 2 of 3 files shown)


ashay:
@harsh-nod I am unsure if these changes are correct, so if you can take a look as well, that'd be great.  Thanks!

wave_lang/kernel/wave/cache.py
```diff
@@ -137,6 +137,7 @@ def anonymize_constraints(input_constraints: list[Constraint]):
             constraint.wave_id = None
         else:
             continue
```
harsh-nod, line 139: Good find. Could you update the type signature of this function to include the return type?

wave_lang/kernel/wave/compile.py
```diff
@@ -215,6 +215,12 @@ def wave_compile(options: WaveCompileOptions, kernel: "LaunchableWave") -> WaveK
     cache_manager = None
     binary_path = None
 
+    # Since constraints are used to lookup the compiled kernel in the cache,
+    # we initialize/update the constraints _before_ the cache lookup.
+    kernel.initialize_wave_constraints()
```
harsh-nod, line 220: Why move these from the pass pipeline to here? Aren't the constraints always guaranteed to be initialized by the time we arrive here?

wave_lang/kernel/wave/wave.py
```diff
@@ -332,7 +332,7 @@ def initialize_wave_constraints(self, trace: CapturedTrace) -> None:
         if hardware_constraint.waves_per_block is None:
             waves_per_block = [1, 1, 1]
             for wave_constraint in self.wave_constraints:
-                count = subs_idxc(wave_constraint.waves_per_block)
```
harsh-nod, line 335: Are we essentially forcing waves_per_block to be non-symbolic?

harsh-nod (changes requested):

wave_lang/kernel/wave/cache.py
ashay: Thanks!  Fixed in the most recent update.

wave_lang/kernel/wave/compile.py
ashay: They aren't.  Without this change, these calls (and thus the constraint initialization) happen after the cache lookup, not before, and this patch instead moves these calls to happen before the cache lookup.

harsh-nod: makes sense, thanks!

harsh-nod (approved):

PR #94, opened 2025-07-24
Support unaligned and unconstrained shapes in expansion

nithinsubbiah:
This is the first of two PR's that fixes numerical correctness for cases when a kernel dimension is non-aligned and is not present in any constraint. This current patch will produce correct results when `minimize_global_loads` is turned off and a follow-up patch will fix the bug in `minimize_global_loads`.

### DIFF
--- a/wave_lang/kernel/ops/wave_ops.py
+++ b/wave_lang/kernel/ops/wave_ops.py
@@ -1161,6 +1161,15 @@ class Placeholder(CustomOp):
 
         get_custom(var).index = value
 
+    # This method is created for parity with `Allocate` op and is used
+    # when calculating bound expressions.
+    @property
+    def get_unpadded_dims(self) -> dict[IndexSymbol, IndexExpr]:
+        unpadded_dim = {}
+        for sym_type in self.type.symbolic_shape:
+            unpadded_dim[sym_type] = sym_type
+        return unpadded_dim
+
 
 @dataclass
 class IterArg(Placeholder):


wave_lang/kernel/wave/generate_bounds_exprs.py
```diff
@@ -52,20 +64,22 @@ def is_read_write(node: fx.Node):
...
             # global mem, but we still need to handle masking against vector
             # size during shared mem access.
-            # A bound expression for this case is `min(global_bound, vector_size)`.
-            # Replace global bound with `max(tile_size, vector_size)` so the entire
-            # expression `min(max(tile_size, vector_size), vector_size)` can be
-            # simplified to just vector size.
             bounds = {
-                k: safe_subs(v, {k: _get_max_tile_size(k, constraints, vector_shapes)})
+                k: (
+                    _get_unconstrained_tile_size(node)
```
Hardcode84, line 77: I think we can just always replace `_get_max_tile_size` with `_get_unconstrained_tile_size` and don't need `unconstrained_dim`.

wave_lang/kernel/wave/generate_bounds_exprs.py
```diff
@@ -33,6 +37,14 @@ def _get_max_tile_size(
     return ret
 
 
+# For non-constrained, unalgined shapes we don't require padding as the vector shape will
+# be the tile size.
+def _get_unconstrained_tile_size(node):
+    return (
+        get_custom(node.memory).distributed_shape[-1] - get_custom(node.memory).padding
```
Hardcode84, line 44: Can you make this `Allocate` op method?

nithinsubbiah: done

wave_lang/kernel/wave/generate_bounds_exprs.py
nithinsubbiah: Yeah that fails the test too

wave_lang/kernel/ops/wave_ops.py
```diff
@@ -1209,6 +1209,10 @@ def indexing_dims(self) -> list[IndexSymbol]:
     def type(self) -> "Memory":
         return Memory[(*self.shape, self.address_space, self.dtype)]
 
+    @property
+    def get_tile_size(self) -> int:
```
Hardcode84, line 1213: This is not very informative name, something like `@property def unpadded_shape()` (which return full shape) will be better.

wave_lang/kernel/wave/generate_bounds_exprs.py
Hardcode84: Something like
```
            bounds = {
                k: safe_subs(v, {k: get_custom(node.memory).unpadded_shape[k]})
                for k, v in bounds.items()
            }
```

nithinsubbiah: I'm still not certain that constrained dimension needs to have unpadded shape as the bound

wave_lang/kernel/ops/wave_ops.py
nithinsubbiah: That makes sense, changed it to reflect the final dimension unpadded shape

Hardcode84 (approved):

PR #95, opened 2025-07-24
Add debug_print_global op

willghatch:
This is like debug_log_write, except it automatically prints, so you can have a one-line print in your kernel which prints the whole tensor.

Also add some minor changes to debug_log_write handling.

### DIFF
--- a/wave_lang/kernel/ops/wave_ops.py
+++ b/wave_lang/kernel/ops/wave_ops.py
@@ -130,9 +130,13 @@ def write(
 
 def debug_log_write(
     register_: "Register",
-    elements_per_thread: Optional[IndexExpr | int] = None,
-    mapping: Optional[IndexMapping] = None,
-    mapping_dynamic_vals: "Register" | tuple["Register", ...] = (),
+    label: Optional[str],
+): ...
+
+
+def debug_print_global(
+    register_: "Register",
+    label: Optional[str],
 ): ...
 
 
@@ -2035,7 +2039,7 @@ class DebugLogWrite(CustomOp):
     """
 
     register_: fx.Proxy
-    log_name: Optional[str] = None
+    label: Optional[str] = None
 
     @property
     def memory(self) -> Optional[fx.Proxy]:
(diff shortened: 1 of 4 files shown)


wave_lang/kernel/wave/debug_log_hoist.py
```diff
@@ -93,4 +96,5 @@ def is_arg_placeholder(node):
         custom = get_custom(debug_placeholder_op)
         debug_placeholder_op.meta["dtype"] = custom.type.dtype
         debug_placeholder_op.meta["symbolic_shape"] = custom.type.symbolic_shape
-        debug_arg_info.append(debug_placeholder_op.meta)
+        # insert since we find them in reverse order of the log statements.
```
martin-luecke, line 99:
```suggestion
        # insert in front since the log statements are in reverse order.
```

wave_lang/kernel/ops/wave_ops.py
```diff
@@ -130,9 +130,13 @@ def write(
 
 def debug_log_write(
     register_: "Register",
-    elements_per_thread: Optional[IndexExpr | int] = None,
-    mapping: Optional[IndexMapping] = None,
-    mapping_dynamic_vals: "Register" | tuple["Register", ...] = (),
```
martin-luecke, line 135: Why are these not needed anymore?

martin-luecke (changes requested):
Thank you for the PR! 

Why does this operation have "global" in the name? To me that suggests that it is somehow concerned with global memory.  

Could this new operation also be an option on the existing `debug_log_write`? 
Especially since this funcionality only depends on `placeholder.meta["debug_print_style"]` AFAICS.

Possibly the existing op could simply take an optional argument `print_func` where users could pass the builtin print function to print to the shell or e.g. print the value to a file. (like here: https://github.com/iree-org/wave/blob/6e3dcc544b8d38d8c97d2932872e74655d3db1b3/wave_lang/kernel/wave/utils/print_utils.py#L724-L726)

Ideally, we try to keep the set of operations in wave to the strictly required set and avoid having multiple operations accomplishing the same behavior.
Additionally, each new operation requires proper tests written for it. If you go down the route of adding a new argument for the existing `debug_log_write` op you could simply add to its existing test in `tests/kernel/wave/wave_e2e_test.py`: 
`tkw.debug_log_write(rhs, print_func=print)`.

wave_lang/kernel/ops/wave_ops.py
willghatch: They were actually removed from the actual operation in the previous commit before it landed, but I missed updating the signature.

willghatch:
I think I agree that we should have one op for doing this rather than multiple.  I wrote this PR thinking it would be a quick feature add for easier debugging for the team, but I think I'll just skip this and go for the more general approach anyway, where printing is one of several options.  I'm going to close this PR, but split out some of the extra stuff that I put into it into this PR into separate PRs.

PR #96, opened 2025-07-25
Temporary fix for math library loading while using cache

yichiche:
This error causes unit test failures on subsequent runs if cache_on is not set to 0. The issue stems from missing library context (e.g., math) when deserializing cached lambda functions.

With this workaround in place, we can proceed with submitting the PR for SGLang review.


nithinsubbiah (approved):

nithinsubbiah:
Please sign off the commit before merging

PR #100, opened 2025-07-25
Add initial RMSNorm kernel

adedespirlet:
This PR implements a first working version of the RMSNorm kernel for inputs of shape [M, N], where M is the number of tokens and N is the embedding dimension. The kernel normalizes each token across its embedding vector using the RMS formulation, followed by elementwise scaling with the learnable parameter gamma.

This kernel targets the naive decode step (M = 1) used in LLM inference, but it is written to support multiple tokens (M > 1) as well. It uses an embedding vector of size 5120, matching the shape of the RMSNorm operator used in LLaMA 4.

The kernel distributes M tokens across workgroups, with the number of tokens per workgroup set via TOKENS_PER_WK. For each token, up to 4 waves are used to compute the RMSNorm over N. Hence Intra-workgroup reduction is used by setting (block=True). Elements per thread are automatically inferred based on N and number of waves.

This is the first functional draft. Performance tuning and tiling strategies will be addressed in a follow-up.

### DIFF
--- a/tests/kernel/wave/rmsnorm_test.py
+++ b/tests/kernel/wave/rmsnorm_test.py
@@ -1,0 +1,106 @@
+# Copyright 2025 The IREE Authors
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+import pytest
+import torch
+from torch.nn import functional as F
+import wave_lang.kernel as tk
+import wave_lang.kernel.lang as tkl
+import wave_lang.kernel.wave as tkw
+from wave_lang.kernel.lang.global_symbols import *
+from wave_lang.kernel.wave.utils.run_utils import (
+    set_default_run_config,
+)
+from wave_lang.kernel.wave.utils.torch_utils import (
+    device_randn,
+    device_zeros,
+    device_ones,
+    device_arange,
+    device_randint,
+)
+from wave_lang.kernel.wave.compile import WaveCompileOptions, wave_compile
+from .common.utils import (
+    require_e2e,
+)
+from wave_lang.kernel.wave.compile import wave_compile, WaveCompileOptions
+
+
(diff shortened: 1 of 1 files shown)


tests/kernel/wave/rmsnorm_test.py
```diff
@@ -0,0 +1,106 @@
...
+        output = a_scaled * gamma_broad
+        tkw.write(output, c, elements_per_thread=ELEMS_PER_THREAD)
+
+    options = WaveCompileOptions(
+        subs={
+            M: shape[0],
+            N: shape[1],
+            TOKENS_PER_WK: 1,
+            EMB_SIZE: shape[1],
+            ADDRESS_SPACE: tkl.AddressSpace.GLOBAL_MEMORY.value,
```
tgymnich, line 89:
```suggestion
            ADDRESS_SPACE: GLOBAL_ADDRESS_SPACE,
```

tests/kernel/wave/rmsnorm_test.py
```diff
@@ -0,0 +1,106 @@
...
+        },
+        canonicalize=True,
+    )
+    options = set_default_run_config(options)
+    test = wave_compile(options, test)
+    print(test.asm)
```
tgymnich, line 95: is this left over from debugging and could this be removed?

adedespirlet: yes indeed thanks!;)

tests/kernel/wave/rmsnorm_test.py
```diff
@@ -0,0 +1,105 @@
...
+            ADDRESS_SPACE: GLOBAL_ADDRESS_SPACE,
+        },
+        canonicalize=True,
+    )
+    options = set_default_run_config(options)
+    test = wave_compile(options, test)
+
+    torch.manual_seed(1)
+    randint_hi = 4
+    a = device_randint(randint_hi, (shape[0], shape[1]), dtype=torch.float16)
```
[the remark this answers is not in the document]
adedespirlet: yep true

PR #101, opened 2025-07-25
[Wait] Waitcnt counter

Hardcode84:


wave_lang/kernel/wave/barriers.py
```diff
@@ -113,33 +134,93 @@ def add_shared_memory_barriers(
...
-            )
+            subgraph = trace.get_subgraph(custom.subgraph_name)
+            if not any(i.async_deps for i in info.values()):
+                add_shared_memory_barriers(trace, subgraph, info)
+                continue
+
+            first = get_first(subgraph.nodes)
+
+            # Convert dependencies to placeholders.
+            with subgraph.inserting_before(first):
```
Hardcode84, line 167: add tp util

wave_lang/kernel/wave/codegen/handlers.py
```diff
@@ -1395,33 +1398,159 @@ def handle_set_wave_prio(emitter: WaveEmitter, node: fx.Node):
...
+        Read,
+        Write,
+    )
+
+    def get_edges(node: AsyncDepNode) -> list[AsyncDepNode]:
+        if (
+            node.read_count > _MAX_ASYNC_READ_WRITE_COUNT
+            or node.write_count > _MAX_ASYNC_READ_WRITE_COUNT
+        ):
+            return []
```
Hardcode84, line 1439: return 0 wait count?

wave_lang/kernel/wave/codegen/handlers.py
```diff
@@ -1395,33 +1398,159 @@ def handle_set_wave_prio(emitter: WaveEmitter, node: fx.Node):
...
+
+        read_count = node.read_count
+        write_count = node.write_count
+        if (
+            isinstance(custom, Read)
+            and custom.memory_type.address_space == GLOBAL_ADDRESS_SPACE
+        ):
+            read_count += ceildiv(
+                custom.elements_per_thread * custom.type.dtype.bitwidth(), 32
```
Hardcode84, line 1452: comment for DWORD size

wave_lang/kernel/wave/barriers.py
Hardcode84: I've thought on it, but it too intertwined with the pass logic as we need to insert new ops and update the `SharedMemoryBarrierInfo` at the same time.

PR #102, opened 2025-07-25
support dynamic shapes in debug_log_write

willghatch:


tests/kernel/wave/wave_e2e_test.py
```diff
@@ -2251,9 +2251,9 @@ def test(
     options = WaveCompileOptions(
```
rohan-tan-bhowmik, line 2251: Would it be beneficial to separate only static-dim arrays and dynamic-dims into two separate tests?

rohan-tan-bhowmik (approved):

tests/kernel/wave/wave_e2e_test.py
willghatch: Perhaps, I'll update it if you want.

PR #104, opened 2025-07-25
Bump minimum torch version to 2.6

paulzzy:
`torch.fx.graph.output_node` was [added in PyTorch 2.6](https://github.com/pytorch/pytorch/releases/tag/v2.6.0#:~:text=Add-,output_node,-util%20function%20to) and is used by `wave_lang.kernel.wave.schedule_reordering.insert_prefetch_loop_barriers`. That function is used by the test 
`lit_tests/kernel/wave/gemm.py:test_gemm_two_cluster_pingpong` (which also had a typo that I fixed).

And since ROCm 6.2 only supports up to PyTorch 2.5.1, the minimum ROCm version is now 6.3.

### DIFF
--- a/lit_tests/kernel/wave/gemm.py
+++ b/lit_tests/kernel/wave/gemm.py
@@ -1366,7 +1366,7 @@ def test_dynamic_gemm_pipelined():
 
 
 @run_test
-def test_gemm_two_cluter_pingpong():
+def test_gemm_two_cluster_pingpong():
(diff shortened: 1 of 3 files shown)


paulzzy:
I discovered this issue after using `uv pip install -r pytorch-rocm-requirements.txt`, which installed `torch==2.5` given the original version constraint of `torch>=2.5,<2.8` (as compared to `pip`, which installed `torch==2.7`). I'm guessing their version resolvers are different.

tgymnich (approved):

PR #110, opened 2025-07-28
Rename Turbine to Wave

tgymnich:
- **remove turbine name from CI**
- **remove turbine name from docs and examples**
- **remove turbine name from logger**
- **remove turbine name from tests**


harsh-nod (approved):
Thanks for the cleanup!

PR #112, opened 2025-07-28
[Wave] Add sinh op

panditsa:


raikonenfnu (approved):

tgymnich (commented):
Please add a test case to `lit_tests/kernel/wave/codegen.py`

tgymnich (approved):

PR #114, opened 2025-07-29
Build with `uv` in GitHub Actions

paulzzy:
Should produce functionally identical builds while being much faster and requiring fewer scripts we need to maintain. There's an existing issue with a [race condition](https://github.com/ncipollo/release-action/issues/448) I haven't had time to fix, but otherwise this should [work fine](https://github.com/paulzzy/wave/actions/runs/16587525596).

This is a prerequisite for supporting Windows builds.

### DIFF
--- a/.github/workflows/build_packages.yml
+++ b/.github/workflows/build_packages.yml
@@ -50,7 +50,7 @@ jobs:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v6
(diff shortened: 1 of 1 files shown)


.github/workflows/build_packages.yml
```diff
@@ -28,59 +28,51 @@ jobs:
...
-            python-version: cp313-cp313
+            python-version: "3.13"
 
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
-      - name: "Setting up Python"
-        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
```
tgymnich, line 53: Please pin the extension to a commit hash.

tgymnich (approved):

PR #118, opened 2025-07-29
fix signature of debug_log_write

willghatch:
In the original PR adding debug_log_write it originally had more arguments, and after removing them from the implementation I forgot to remove them from the type annotation.


willghatch:
FYI @martin-luecke this is split out from https://github.com/iree-org/wave/pull/95

martin-luecke (approved):

PR #119, opened 2025-07-29
NFC move some reordering logic for debug logs

willghatch:
This will be helpful when there is more processing.  I originally did this in https://github.com/iree-org/wave/pull/95 but I have since decided to approach the main aspect of it in a different way, I'm pulling out this small NFC that will still be helpful for other ways to write printing / visualizingg features.


willghatch:
FYI @martin-luecke this is split out from https://github.com/iree-org/wave/pull/95, thanks for your PR comment improving my comment in the code, I've further edited it just a smidge for clarity.

tgymnich (approved):

PR #123, opened 2025-07-30
Re-enable `manylinux` builds with `cibuildwheel`

paulzzy:
It would also be straightforward to add `musllinux` builds.

See https://github.com/paulzzy/wave/actions/runs/16614171122 for an example build-and-release run.

### DIFF
--- a/.github/workflows/build_and_release.yml
+++ b/.github/workflows/build_and_release.yml
@@ -30,44 +30,36 @@ jobs:
         include:
           # Linux packages
           - runs-on: ubuntu-24.04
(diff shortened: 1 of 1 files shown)


harsh-nod:
Where will the build artifacts be stored? And how will they be uploaded to pypi? I see they are being stored to a wheelhouse directory but how do we get it from there to push to pypi?

paulzzy:
Depends on https://github.com/iree-org/wave/pull/130

harsh-nod (approved):

PR #126, opened 2025-07-30
Migrate existing mi300 runners to new mi325 capacity.

deedongala:


.github/workflows/ci.yaml
```diff
@@ -35,7 +35,7 @@ jobs:
         include:
           - name: linux-mi300
```
saienduri, line 36: change name

PR #131, opened 2025-07-30
Migrate MI300 Capacity to new MI325 Capacity.

deedongala:
Runner capacity is shifting from MI300 to MI325 machines. These machines are compatible enough for testing, though benchmarks will need new baselines.

### DIFF
--- a/.github/workflows/perf.yaml
+++ b/.github/workflows/perf.yaml
@@ -31,7 +31,7 @@ jobs:
       fail-fast: false
       matrix:
         version: [3.11]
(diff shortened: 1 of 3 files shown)


.github/workflows/perf.yaml
```diff
@@ -64,7 +64,7 @@ jobs:
           pip install -r requirements.txt -e .
 
       - name: Run e2e tests on MI300
```
saienduri, line 66: update name

Hardcode84 (approved):

PR #133, opened 2025-07-30
[Wave] Added bool to float casting

badgerbroch:
When tkw.cast is used with a bool -> float the resulting cast will now be:

- true -> 1.0
- false -> 0.0

This matches the semantics for python, c++.

Previously when doing this it would have the behavior of:

- true -> -1.0
- false -> 0.0


Hardcode84:
Can you add a lit test for it?

badgerbroch:
> Can you add a lit test for it?

Yes will do

lit_tests/kernel/wave/casting.py
```diff
@@ -81,3 +83,4 @@ def test(
     # CHECK:  %[[D4:.*]] = arith.extsi %[[D3]] : vector<16xi16> to vector<16xi32>
     # CHECK:  %[[D5:.*]] = arith.sitofp %[[D4]] : vector<16xi32> to vector<16xf32>
     # CHECK:  %[[D6:.*]] = arith.truncf %[[D5]] : vector<16xf32> to vector<16xf16>
+    # CHECK: arith.select %{{.*}}, %{{.*}}, %{{.*}} : vector<16xi1>, vector<16xf16>
```
Hardcode84, line 86: please match the actual constants values.

Hardcode84 (approved):

PR #144, opened 2025-08-01
[Wave] Optimize `subs_idxc`

Hardcode84:
Cache subs queries in the `IndexingContext`, to help sympy-heavy passes like `generate_bounds_exprs`

### DIFF
--- a/wave_lang/kernel/_support/indexing.py
+++ b/wave_lang/kernel/_support/indexing.py
@@ -82,7 +82,7 @@ def safe_subs(
     input: IndexExpr | int | IndexSequence,
     subs: dict[IndexSymbol, int | IndexSymbol],
     simultaneous: bool = False,
-) -> IndexSymbol | int:
+) -> IndexSymbol | int | IndexSequence:
     """
     Substitute input using provided `subs` list if input is sympy object.
     Otherwise return input unchanged.


wave_lang/kernel/wave/codegen/emitter.py
```diff
@@ -54,8 +54,8 @@
 
 
 def _get_upper_bound(expr: Any) -> Optional[Attribute]:
-    res = subs_idxc(sympy.sympify(expr))
```
martin-luecke, line 57: Does removing the `sympify` call possibly change some simplifications performend on expr or is this a non functional change?

i.e. could this impact the code we generate?

wave_lang/kernel/wave/utils/general_utils.py
```diff
@@ -311,7 +311,7 @@ def evaluate_with_assumptions(constraints: list[Constraint], expr: IndexExpr) ->
     # Solve returns a false if the inequalities are not consistent.
     if isinstance(result, sympy.logic.boolalg.BooleanAtom):
         return False
-    return True if any([result.equals(x) for x in facts]) else None
```
martin-luecke, line 314: Why this change? Was this incorrect before?

wave_lang/kernel/wave/utils/symbol_utils.py
```diff
@@ -3,35 +3,19 @@
...
-    """
-    Substitute input using IndexingContext if input is sympy object.
-    Otherwise return input unchanged.
-    """
-    idxc = IndexingContext.current()
-    return safe_subs(input, idxc.subs)
+# Reexport symbols from indexing.py
+from ..._support.indexing import (
+    IndexExpr,
+    IndexingContext,  # noqa
```
martin-luecke, line 13: Why do these lines need to be marked?

martin-luecke (approved):
LGTM, only a few questions + more precise type annotations.

Do you see a good way of adding a test that the caching (or indeed the mapping) actually works as expected?

wave_lang/kernel/wave/codegen/emitter.py
```diff
@@ -54,8 +54,8 @@
...
-    if res.is_number:
+    res = subs_idxc(expr)
+    if isinstance(res, int) or res.is_number:
```
raikonenfnu, line 58: Why would there be a change here?

wave_lang/kernel/wave/codegen/emitter.py
Hardcode84: `sYmpify` just converts basic types to sympy types, it should have no effect.

wave_lang/kernel/wave/codegen/emitter.py
Hardcode84: Fun fact `1` and `sympy.core.numbers.One` will occupy same slot in the dict, so we can end up here with actual integers or symbol literals from the cache. Regardless, refactored it to the standalone func.

PR #149, opened 2025-08-01
[Wave] Gather to lds swizzling

Hardcode84:
Add `GatherToLDS` swizzling. Swizzling is applied to shared memory reads index and then the inverse swizzling is applied the the `GatherToLDS` global read index (as its write index is always contiguous), "Inverse swizzling" is in fact the same formula as applying `xor` second time produces the original result.

### DIFF
--- a/wave_lang/kernel/wave/gather_to_shared.py
+++ b/wave_lang/kernel/wave/gather_to_shared.py
@@ -38,10 +38,11 @@ from .minimize_global_loads import (
 from .utils.general_utils import (
     ceildiv,
     delinearize_index,
+    find_index_bounds,
     get_hardware_constraint,
     infer_dim,
     remove_thread_indexing,
-    find_index_bounds,
+    remove_global_indexing,
 )
 from .utils.graph_utils import DCE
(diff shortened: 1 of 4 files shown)


wave_lang/kernel/wave/gather_to_shared.py
```diff
@@ -438,3 +448,98 @@ def gather_to_shared(
...
+            logger.info(f"shape={shape} must be at least 2D")
+            continue
+
+        col_dim = infer_dim(shape[-1])
+        row_dim = infer_dim(shape[-2])
+
+        max_phase = 8
+
+        for read in reads:
+            index = remove_global_indexing(read.index, constraints)
```
raikonenfnu, line 521: NIT/thought exercise: Thoughts on moving this pass after `apply_shared_memory_indexing_corrections` so we may not need to remove the global indexing?

Hardcode84: I would prefer to just move `apply_shared_memory_indexing_corrections ` earlier (or just do it as part of promotion)

PR #150, opened 2025-08-01
[Wave] Fix tail-padded memref allocs when `minimize_shared_allocs` is disabled

Hardcode84:
`minimize_shared_allocs` handles memref tail padding automatically, but  when it's disabled we need to handle it in lowering.
Tail-padded memrefs cannot be represented as normal allocs so allocate them as flat i8 array and take a view.

### DIFF
--- a/lit_tests/kernel/wave/gather_to_shared.py
+++ b/lit_tests/kernel/wave/gather_to_shared.py
@@ -269,3 +269,76 @@ def test_gather_to_shared_scaled_dims():
     # CHECK-COUNT-8:      vector.load
     # CHECK-COUNT-2:      amdgpu.scaled_mfma
     # CHECK-COUNT-4:    vector.store
+
+
+@run_test
+def test_gather_to_shared_not_minimize_shared_allocs():
+    constraints: list[tkw.Constraint] = [tkw.WorkgroupConstraint(M, BLOCK_M, 0)]
+    constraints += [tkw.WorkgroupConstraint(N, BLOCK_N, 1)]
+    constraints += [tkw.TilingConstraint(K, BLOCK_K)]
+    constraints += [tkw.WaveConstraint(M, BLOCK_M / 2)]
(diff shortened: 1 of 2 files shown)


wave_lang/kernel/wave/codegen/handlers.py
```diff
@@ -197,17 +197,24 @@ def handle_allocate(emitter: WaveEmitter, node: fx.Node):
...
-    ), "Tail padding must be handled by the minimize_shared_allocs pass"
-
     memref_shape = cast_py_literal(emitter, distributed_shape)
     element_type = IrType.parse(dtype.ir_type_asm())
     address_space = Attribute.parse("#gpu.address_space<workgroup>")
     memref_type = MemRefType.get(memref_shape, element_type, None, address_space)
 
     if parent is not None:
         parent = cast_py_value(emitter, parent).ir_value
+    elif tail_padding != 0:
```
raikonenfnu, line 207: It's unfortunate there is no easy way we can have a common handling for tail padding, but since I have no better suggestions going to approve for now, we should in parallel keep thinking of better ideas :)

lit_tests/kernel/wave/gather_to_shared.py
```diff
@@ -269,3 +269,76 @@ def repeat(acc: tkl.Register[M, N, tkl.f32]) -> tkl.Register[M, N, tkl.f32]:
...
+
+    # CHECK-LABEL:    test_gather_to_shared_not_minimize_shared_allocs
+    # CHECK:          func.func @scaled_gemm
+    # CHECK:            memref.alloc()
+    # CHECK:            memref.view
+    # CHECK:            memref.alloc()
+    # CHECK:            memref.view
+    # CHECK:            memref.alloc()
+    # CHECK:            memref.view
+    # CHECK:            memref.alloc()
```
raikonenfnu, line 342: can we just add a CHECK for the padded tail shape and view?

PR #151, opened 2025-08-01
Add support for dynamic variables in new IndexMapping API

harsh-nod:
This PR adds support for dynamic mappings with the new API. The extend attention template is also modified to work with the new API.


wave_lang/kernel/wave/templates/extend_attention.py
```diff
@@ -260,23 +208,26 @@ def first_loop(
             q_reg = tkw.read(
                 q,
                 elements_per_thread=LOAD_ELEMS_PER_THREAD_QK,
-                mapping=q_mapping,
+                source=(n_q + EXT_IDX, h, d_q),
```
nithinsubbiah, line 211: Do the position of iterators matter? In the mask mapping `N_Q` was the second dimension but here it's the first

wave_lang/kernel/wave/templates/extend_attention.py
```diff
@@ -130,72 +130,20 @@ def get_extend_attention_kernel(
...
+    h_kv = tkl.sym.h_kv
+    d_kv = tkl.sym.d_kv
+    n_kv = tkl.sym.n_kv
+    h = tkl.sym.h
+    n_q = tkl.sym.n_q
+    d_q = tkl.sym.d_q
+    s = tkl.sym.s
+
+    bindings = {h_kv: H_KV, d_kv: D_KV, n_kv: N_KV}
+    bindings.update({h: H, n_q: N_Q, d_q: D_Q, s: S})
```
nithinsubbiah, line 143: Can all these variables be defined as in line 142? Is `update` being used just for completeness/cleaner code?

nithinsubbiah (approved):

wave_lang/kernel/wave/templates/extend_attention.py
harsh-nod: yes, just wanted to separate variables for *_KV and *_Q, but can all be done in one line.

wave_lang/kernel/wave/templates/extend_attention.py
harsh-nod: yes, the order of the iterators should match the shape. So here the shape of the source would be (N_Q, H, D_Q).

PR #153, opened 2025-08-01
[debugging] add printer and handler args to debug_log

willghatch:
This provides an easy in-line way to print debug logs without modifying code outside of their use site. It also modifies the debug_logs dictionary to be able to hold extra metadata that will be useful for a viewer.


wave_lang/kernel/wave/compile.py
```diff
@@ -169,6 +175,15 @@ def get_dynamic_dimension_actual(sym):
                     benchmark_results, self.options.benchmark_results_file
                 )
 
+        if self.debug_outputs:
+            for info_dict, (label, debug_log) in zip(
+                self.debug_outputs, debug_logs.items()
+            ):
+                if info_dict.get("printer", None):
```
Hardcode84, line 182: nit:
```
if printer := info_dict.get("printer", None)
    printer(label, debug_log["value"])
```

Hardcode84 (approved):

PR #155, opened 2025-08-02
Support unaligned and unconstrained shapes in expansion (#94)

nithinsubbiah:
This is the first of two PR's that fixes numerical correctness for cases when a kernel dimension is non-aligned and is not present in any constraint. This current patch will produce correct results when `minimize_global_loads` is turned off and a follow-up patch will fix the bug in `minimize_global_loads`.

---------


Hardcode84:
Can we add a lit test to what was broken in scaled tests?

raikonenfnu:
+1 on the LIT test, in theory https://github.com/iree-org/wave/blob/main/lit_tests/kernel/wave/scaled_gemm.py#L524-L622 should be protecting it, but may be missing some piece there

PR #193, opened 2025-08-12
[debugging][docs] add doc explaining how to use debug_log, html_viewer

willghatch:
It would be somewhat better with some screenshots of the viewer.  But since I expect the viewer to still change quickly at this point, I would rather not bloat the git repository with images that are soon going to be outdated.

This PR also documents stuff that's still pending in https://github.com/iree-org/wave/pull/180

### DIFF
--- a/docs/wave/debugging.rst
+++ b/docs/wave/debugging.rst
@@ -1,0 +1,80 @@
+Debugging in Wave
+=================
+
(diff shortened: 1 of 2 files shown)


harsh-nod (approved):
lgtm! thanks!

PR #201, opened 2025-08-14
[compiler] Apply padding in `materialized_shapes`

nithinsubbiah:
This patch fixes numerical error when the problem shape is unaligned and not constrained. During `minimize_global_loads` `materialized_shape` is used to calculate the new optimized access pattern and the issue is solved by applying padding to the unaligned dimension. We haven't encountered this issue before because the unaligned shape has always been part of the constraints.

Follow-up to https://github.com/iree-org/wave/pull/155

### DIFF
--- a/lit_tests/kernel/wave/minimize_global_loads.py
+++ b/lit_tests/kernel/wave/minimize_global_loads.py
@@ -13,6 +13,7 @@ from wave_lang.kernel.wave.analysis.index_sequence_analysis import (
     set_post_expansion_indices,
 )
 from wave_lang.kernel.wave.barriers import add_shared_memory_barriers
+from wave_lang.kernel.wave.compile import WaveCompileOptions, wave_compile
 from wave_lang.kernel.wave.expansion.expansion import add_get_results, expand_graph
 from wave_lang.kernel.wave.hoisting import hoist_loop_invariant_ops
 from wave_lang.kernel.wave.minimize_global_loads import minimize_global_loads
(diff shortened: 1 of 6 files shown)


lit_tests/kernel/wave/minimize_global_loads.py
```diff
@@ -271,5 +272,72 @@ def test_gemm():
...
+    gemm = wave_compile(options, gemm)
+    print(gemm.asm)
+
+    # CHECK-LABEL:    test_materialized_shape_padding
+    # CHECK-DAG:        #{{.*}} =  affine_map<()[s0] -> ((s0 floordiv 4) mod 16)>
+    # CHECK-DAG:        #{{.*}} =  affine_map<()[s0] -> (s0 * 8 - (s0 floordiv 4) * 32)>
+    # CHECK:            func.func @gemm
+    # CHECK:                %{{.*}} = arith.constant dense<17> : vector<8xindex>
+    # CHECK:                %{{.*}} = arith.constant dense<[0, 1, 2, 3, 4, 5, 6, 7]> : vector<8xindex>
+    # CHECK:                %{{.*}} = vector.load %0[%1, %2] : memref<16x17xf16, strided<[17, 1], offset: ?>>, vector<8xf16>
```
raikonenfnu, line 339: can we make this regex variables?

lit_tests/kernel/wave/minimize_global_loads.py
```diff
@@ -271,5 +272,72 @@ def test_gemm():
...
+        K: 17,
+    }
+    options = WaveCompileOptions(
+        subs=hyperparams,
+        canonicalize=True,
+    )
+    gemm = wave_compile(options, gemm)
+    print(gemm.asm)
+
+    # CHECK-LABEL:    test_materialized_shape_padding
```
raikonenfnu, line 333: can you write what to look out for in the LIT test, and why this matters

lit_tests/kernel/wave/minimize_global_loads.py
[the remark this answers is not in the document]
nithinsubbiah: done

lit_tests/kernel/wave/minimize_global_loads.py
nithinsubbiah: done

raikonenfnu:
rebase please

Hardcode84 (approved):

lit_tests/kernel/wave/minimize_global_loads.py
nithinsubbiah: Added a comment on what exactly this tests. I don't think regex is needed for these variables as we're just testing the affine maps and the number of elements during load

raikonenfnu (approved):
LGTM

PR #202, opened 2025-08-15
Add support for modulo scheduling in bshd kernels

harsh-nod:
This PR adds support for modulo scheduling in the bshd kernels. In order to do this, it implements support for hoisting binarypyops out of iterate nodes. This was previously handled in the downstream compiler stages but we do it here to enable scheduling.


raikonenfnu (commented):
Overall code looks good, just a QQ

PR #215, opened 2025-08-19
[Compiler] Combine all buffer ops option into a single option.

raikonenfnu:
- Refactor codebase to use a single use_buffer_ops options
- Refactor read_write.py for better legibility


wave_lang/kernel/wave/codegen/read_write.py
```diff
@@ -444,6 +445,19 @@ def _get_out_of_bounds_index(element_type: IrType) -> int:
     return oob_index_value
 
 
+def _get_constant_value(candidate_op: Any):
```
Hardcode84, line 448: nit: `-> Value`

Hardcode84 (approved):

PR #217, opened 2025-08-19
Extend prefill attention

rohan-tan-bhowmik:
Some minor (not affecting normal usage) things that were fixed in order to enable extend prefill attention for Sarathi Prefill + else on SharkTank:

- Support for dynamic dims on k_cache/v_cache were not enabled, causing interesting but unfortunate crashes
- Fixed test to support num_kv_heads being different from num_q_heads


Hardcode84 (commented):
Can we add some tests for it?

rohan-tan-bhowmik:
> Can we add some tests for it?

Yup, added some shapes that the PR enables

Hardcode84 (approved):

PR #219, opened 2025-08-20
[Compiler] Teach compiler to handle scaling in BroadcastOp.

raikonenfnu:
In order to support fused dynamic quantization, we would need to teach our compiler to handle broadcast to scaled shapes/dimensions. We do this by modifying Broadcast's indexing_dim to actually use the dim(K) as oppose to the shape (K/32) + handle said case in determining what dims are actually being broadcasted in codegen.py


lit_tests/kernel/wave/codegen.py
```diff
@@ -1930,6 +1930,65 @@ def broadcast_add(
...
+            vector_shapes={M: BLOCK_M, N: BLOCK_N},
+        )
+    ]
+
+    constraints += [tkw.WorkgroupConstraint(M, BLOCK_M, 1)]
+    constraints += [tkw.WorkgroupConstraint(N, BLOCK_N, 0)]
+    constraints += [tkw.WaveConstraint(M, BLOCK_M)]
+    constraints += [tkw.WaveConstraint(N, BLOCK_N)]
+
+    @tkw.wave(constraints)
```
harsh-nod, line 1956: Can we add these kernels to templates to avoid duplicating them?

wave_lang/kernel/ops/wave_ops.py
```diff
@@ -2360,7 +2360,10 @@ def __post_init__(self):
 
     @property
     def indexing_dims(self) -> list[IndexSymbol]:
-        return self.target_shape
+        from ..wave.utils.general_utils import infer_dim
+
+        dims = [infer_dim(expr) for expr in self.target_shape]
```
harsh-nod, line 2365: Can you add a comment explaining why we use infer_dim here? (specifically so that it allows broadcast ops to correctly process symbolic dims of the form binary_op(K, x) where K is a symbol and x is an integer.

harsh-nod (approved):
Just a request for refactor and comment, but should be good to land after that.

lit_tests/kernel/wave/codegen.py
raikonenfnu: Probably need your guidance on this one, I was thinking of reserving templates for kernels that are actually useful for end users not unlike https://github.com/triton-lang/triton/tree/main/python/triton_kernels/triton_kernels. 

Since these kernels sole purpose right now is for testing compiler feature I was thinking of keeping them in the test itself. I don't mind either ways, but would love your thoughts on it. :)

harsh-nod: Maybe we can have a test templates for things that are not our production kernels, but that keep getting reused for tests etc. I think we have a lot of code duplication because of this.

raikonenfnu: makes sense, I will move for now, and then factor away the performant/deployment kernels in future PR

PR #221, opened 2025-08-21
[Dist.] Update compile_utils default options

panditsa:


wave_lang/kernel/wave/compile_options.py
```diff
@@ -37,6 +37,7 @@ class WaveCompileOptions:
     backend: str = "rocm"
     target: str = "gfx942"
     iree_preprocessing_pass_pipeline: str = None
+    num_devices: int = 1
```
Hardcode84, line 40: Please add it to kernel hash.

wave_lang/kernel/wave/utils/run_utils.py
```diff
@@ -144,5 +144,6 @@ def set_default_run_config(options: WaveCompileOptions) -> WaveCompileOptions:
         options.use_scheduling_barriers = bool(int(enable_scheduling_barriers))
 
     options.backend = "rocm"
+    options.device = "hip"
```
Hardcode84, line 147: `rocm` backend is already implies `hip` device, don't see much reason to keep it in options

panditsa:
> I'm still concerned about using device as compilation parameter. Currently, we are compiling for the specified architecture but the actual device to run kernel on is picked from the input tensor device. This allows to run kernels on any device on multi-gpu system and I would prefer to keep it working (also, sglang is actually using it to run wave on multiple gpus, I think).

Well, at runtime we figure out the actual device from the input tensors but when compiling through iree-compile we still need to provide the `--iree-hal-target-device` or `--iree-hal-target-backends` but not both.

The number of devices and specifically providing them for compiler is a little strange to me but that's how we've done it in sharktank:
```
    iree-compile \
    8b_instruct_tp8.mlir \
    --iree-hip-target=gfx942 \
    -o=8b_instruct_tp8.vmfb \
    --iree-hal-target-device=hip[0] \
    --iree-hal-target-device=hip[1] \
    --iree-hal-target-device=hip[2] \
    --iree-hal-target-device=hip[3] \
    --iree-hal-target-device=hip[4] \
    --iree-hal-target-device=hip[5] \
    --iree-hal-target-device=hip[6] \
    --iree-hal-target-device=hip[7] \
    --iree-opt-level=O3 \
    --iree-stream-resource-memory-model=discrete \
    --iree-hal-indirect-command-buffers=true \
    --iree-hal-memoization=true \
```

Hardcode84 (approved):

PR #223, opened 2025-08-21
Add conditional barriers to attention schedule

harsh-nod:
This PR adds the ability to add conditional barriers for attention when the PREFETCH_ATTENTION schedule is used and 8 waves are specified. Based on annotations during scheduling, operations are clustered into groups and then reordered to enable ping pong.

### DIFF
--- a/wave_lang/kernel/ops/wave_ops.py
+++ b/wave_lang/kernel/ops/wave_ops.py
@@ -1194,11 +1194,10 @@ class IterArg(Placeholder):
 
     @property
     def distributed_shape(self):
-        return self.fx_node.distributed_shape
-
-    @distributed_shape.setter
-    def distributed_shape(self, value):
-        self.fx_node.distributed_shape = value
+        init_arg = self.parent_op().init_args[self.iter_idx]
+        allocate = get_custom(init_arg)
+        assert isinstance(allocate, Allocate)
+        return allocate.distributed_shape
 
     def infer_type(self, *args):
         parent_op = self.parent_op()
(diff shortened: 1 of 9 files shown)


wave_lang/kernel/ops/wave_ops.py
```diff
@@ -1189,6 +1192,14 @@ def iter_idx(self):
     def iter_idx(self, value):
         self.fx_node.iter_idx = value
 
+    @property
+    def distributed_shape(self):
+        return self.fx_node.distributed_shape
```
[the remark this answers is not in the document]
harsh-nod: good point, will fix.

wave_lang/kernel/wave/templates/vanilla_attention.py
```diff
@@ -408,7 +409,7 @@ def base_attention_custom_mask(
         ADDRESS_SPACE: SHARED_ADDRESS_SPACE,
         BLOCK_B: 1,
         BLOCK_H: 1,
-        BLOCK_M: 128,
+        BLOCK_M: 128 * (num_waves // 4),
```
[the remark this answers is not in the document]
harsh-nod: I wanted to keep the shape the same for num_waves = 4 and just scale it linearly from there. Its just a heuristic, open to changing it.

wave_lang/kernel/wave/schedule_reordering.py
```diff
@@ -786,3 +846,309 @@ def schedule_reordering(
...
+
+    for node in graph.nodes:
+        prefetch_stage = node.meta.get("prefetch_stage", None)
+        if prefetch_stage:
+            try:
+                op_type = AttentionOperationType.from_string(prefetch_stage)
+                operation_groups[op_type].append(node)
+            except ValueError:
+                # Skip unknown prefetch stages
+                continue
```
raikonenfnu, line 1031: Shouldn't we return error or skip reordering for unknown prefetch stage?

wave_lang/kernel/wave/scheduling/prefetch_scheduling.py
```diff
@@ -7,6 +7,8 @@
 from enum import Enum
 from typing import Sequence
 
+from ..schedule_reordering import AttentionOperationType
```
raikonenfnu, line 10: Thoughts on moving attentionOperationType to https://github.com/iree-org/wave/blob/main/wave_lang/kernel/wave/utils/classes.py for a cleaner dep chain?

harsh-nod: sure, I can move it there.

PR #224, opened 2025-08-21
Fix issues with multi-buffering

harsh-nod:
In the current multi-buffering code, when doing N buffers along the B dimension, the index was being modified from

{B: ..., K: ... }

to

{N*B: ..., K ... }

But this does not make sense as the index keys are the distribution symbols / indexing dims and not the actual shape of memory allocated.


harsh-nod:
No longer needed

PR #232, opened 2025-08-25
Unit test for MoE GEMM kernel

tyb0807:


tgymnich (approved):

nithinsubbiah:
What's the utility of this test given we already have a test checked in?

PR #239, opened 2025-08-26
`print` by default in `debug_log` op

paulzzy:
Reduces the number of required arguments, makes `debug_log` a little easier to use and faster to write.


willghatch (approved):
This is probably a more convenient default.  Let's do it.

PR #240, opened 2025-08-27
[Compiler Infra] Speed up compile time with LRUCache on simplify

raikonenfnu:
Per @Hardcode84's suggestion, we can make Wave compile go from ~ 30s -> 10s for reordered gemm using lru cache.

### DIFF
--- a/wave_lang/kernel/wave/codegen/read_write.py
+++ b/wave_lang/kernel/wave/codegen/read_write.py
@@ -86,6 +86,11 @@ def _get_start_indices(
     return start_indices
 
 
+@functools.lru_cache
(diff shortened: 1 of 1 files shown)


wave_lang/kernel/wave/codegen/read_write.py
```diff
@@ -86,6 +86,11 @@ def _get_start_indices(
     return start_indices
 
 
+@functools.lru_cache
+def _simplify(expr):
```
Hardcode84, line 90: add some comment?

Hardcode84 (approved):

PR #243, opened 2025-08-27
add water sources

Hardcode84:


ftynse (commented):
Looks reasonable. Let me give at a run first, and then I can help staging this so we don't lose most of the commit history, which tends to be useful in the longer run and has educational power as well.

ftynse:
This currently doesn't run python binding tests. They are skipped by lit for some reason, despite flags.

ftynse (changes requested):
We need to figure out the python bindings story... Wave depends on iREE that vendors MLIR bindings, we cannot have water do the same and have two sets of incompatible bindings. To unblock progress, we may temporarily have two build processes each with its own set of bindings that are tested and look at vendoring our own build separately.

ftynse:
Superceded by #313, #288

PR #250, opened 2025-08-28
Multibuffer liveness analysis

Hardcode84:
Remove `multi_buffer_count` compiler option and use liveness analysis instead to determine how many buffers needed during scheduling to avoid data races during overlapping iterations execution.

`_write_dependency` between shared write and shared read is used to determine how long buffer contents is alive. The total buffer count is `ceildiv(cycles_alive, initiation_interval)`.

Some additional fixes:
* Fix `update_write_dependencies` to properly support `GatherToLDS`, previously they were ignored and corresponding `_write_dependency` weren't populated
* Fix `minimize_shared_allocs` to properly handle situation when buffers are being used as iterargs.

### DIFF
--- a/wave_lang/kernel/wave/scheduling/loop_reconstruction_utils.py
+++ b/wave_lang/kernel/wave/scheduling/loop_reconstruction_utils.py
@@ -315,6 +315,11 @@ def compute_multi_buffer_count(
         assert node in lifetime, f"Node {node} not found in lifetime"
         # Lifetime returns 0 if node result only used on same clock, 1 if it used on next clock, etc,
         # so we need to add 1 to the lifetime to get the number of clocks the result is live.
+        # Ceildiv is required for cases like (lifetime=3, initiation_interval=2) which would otherwise
+        # result in buffer_count=1:
+        # 000
+        #   111
+        #     222
         buffer_count = ceildiv(lifetime[node] + 1, initiation_interval)
         logger.debug(f"Node: {node}, Buffer count: {buffer_count}")
         if buffer_count < 2:


lit_tests/kernel/wave/scheduling.py
```diff
@@ -162,7 +161,7 @@ def test_gemm_pipelined():
     # CHECK-NEXT: %read_2_shared_M:0_N:0_K:1
     # CHECK-NEXT: %scheduling_group_barrier_4
     # CHECK-SAME: ({Operation.MMA: 2, Operation.READ_SHARED: 2}, 0)
-    # CHECK-NEXT: [mma_M:0_N:0_K:1_mapped_1_1, mma_M:0_N:1_K:1_mapped_1_1, mma_M:1_N:0_K:1_mapped_1_1, mma_M:1_N:1_K:1_mapped_1_1, read_4_shared_M:0_N:0_K:1_mapped_2_0, read_2_shared_M:0_N:0_K:1_mapped_2_0, read_2_shared_M:1_N:0_K:0_mapped_2_0, read_2_shared_M:1_N:0_K:1_mapped_2_0, outer_rotating_reg_9, outer_rotating_reg_8, outer_rotating_reg_11, outer_rotating_reg_10]
+    # CHECK-NEXT: [mma_M:0_N:0_K:1_mapped_1_1, mma_M:0_N:1_K:1_mapped_1_1, mma_M:1_N:0_K:1_mapped_1_1, mma_M:1_N:1_K:1_mapped_1_1, read_4_shared_M:0_N:0_K:1_mapped_2_0, read_2_shared_M:0_N:0_K:1_mapped_2_0, read_2_shared_M:1_N:0_K:0_mapped_2_0, read_2_shared_M:1_N:0_K:1_mapped_2_0]
```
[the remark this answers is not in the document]
Hardcode84: Yes, the way `MODULO` is generated currently it doesn't need the multibuffering.

harsh-nod:
Overall looks good to me, will let @raikonenfnu give the +2 after his comment is addressed

tests/kernel/wave/attention/vanilla_attention_test.py
```diff
@@ -1083,6 +1083,7 @@ def testAttentionBSHD_Prefetch_MultiBuffer(
         benchmark_batch_size=10,
         benchmark_repetitions=3,
         benchmark_results_file=perf_filename_tk,
+        multi_buffer_count=2,  # TODO: Hack as schedule reordering doesn't respect lifetimes computed during the main scheduling algorithm.
```
harsh-nod, line 1086: Does that mean that if we specify our own schedule we need to manually set the multi_buffer_count?

Hardcode84: `override_schedule` is consumed before the loop reconstruction so it should work fine without specifying multibuffer count manually.

PR #251, opened 2025-08-28
implement broadcasting for SelectOp

willghatch:
Motivated by my topk implementation which needs broadcasted selection.

### DIFF
--- a/wave_lang/kernel/lang/wave_types.py
+++ b/wave_lang/kernel/lang/wave_types.py
@@ -138,6 +138,9 @@ class Register(metaclass=KernelBufferMeta):
             raise TypeError(f"Expected at least 2 arguments, got: {shape_and_dtype}")
 
         shape = shape_and_dtype[:-1]
+        # allow shape to be provided as a tuple instead of as individual elements, to work around lack of unpacking in subscripts for Python 3.10
+        if len(shape) == 1 and isinstance(shape[0], tuple):
+            shape = shape[0]
         dtype = shape_and_dtype[-1]
 
         # Allow constant int expressions in shape
--- a/wave_lang/kernel/ops/wave_ops.py
+++ b/wave_lang/kernel/ops/wave_ops.py
@@ -20,6 +20,7 @@ from typing import (
 import numpy as np
 import torch.fx as fx
 from typing_extensions import Self
+from itertools import combinations
 
 from .._support.dtype import DataType, i1
 from .._support.indexing import IndexExpr, IndexSequence, IndexSymbol
@@ -884,7 +885,7 @@ class BinaryOpBase(CustomOp, ABC):
 
         lhs_dim_set = set(lhs_type.symbolic_shape)
         rhs_dim_set = set(rhs_type.symbolic_shape)
-        if lhs_dim_set.isdisjoint(rhs_dim_set):
+        if lhs_dim_set and rhs_dim_set and lhs_dim_set.isdisjoint(rhs_dim_set):
             raise ValueError(
                 "BinaryPyOp requires lhs and rhs shape to be at least broadcastable."
                 f" got {lhs_type.symbolic_shape} vs {rhs_type.symbolic_shape}"
@@ -1008,6 +1009,30 @@ class SelectOp(CustomOp):
         combined_dims += get_custom(self.if_false).indexing_dims
         return list(dict.fromkeys(combined_dims))
 
+    def infer_broadcast_shape(
+        self, cond_type: Register, t_type: Register, f_type: Register
+    ):
+        c_shape = getattr(cond_type, "symbolic_shape", ())
+        t_shape = getattr(t_type, "symbolic_shape", ())
+        f_shape = getattr(f_type, "symbolic_shape", ())
+
+        # Check if shapes are broadcastable - they should not be disjoint
+        non_empty_shapes = [s for s in [c_shape, t_shape, f_shape] if s]
+        if any(
+            set(i_shape).isdisjoint(set(j_shape))
+            for i_shape, j_shape in combinations(non_empty_shapes, 2)
+        ):
+            raise ValueError(
+                f"SelectOp requires operand shapes to be broadcastable. "
+                f"Got condition: {c_shape}, if_true: {t_shape}, if_false: {f_shape}"
+            )
+
+        most_general_shape = ()
+        for s in [c_shape, t_shape, f_shape]:
+            if len(s) >= len(most_general_shape):
+                most_general_shape = s
+        return most_general_shape
+
     def infer_type(self, *args):
         cond_type = get_custom(self.cond).type
         if_true_type = get_custom(self.if_true).type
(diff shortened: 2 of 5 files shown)


wave_lang/kernel/wave/analysis/index_sequence_analysis.py
```diff
@@ -217,7 +217,7 @@ def verify_nodes(trace: CapturedTrace, constraints: list[Constraint]):
             continue
         if isinstance(custom.type, DataType):
             continue
-        assert custom.index, f"Index not set for node {custom.fx_node}: {custom}"
+        assert custom.index != None, f"Index not set for node {custom.fx_node}: {custom}"
```
willghatch, line 220: I'm concerned about this change.  There is a lit test that's hitting this with a SelectOp that wasn't before.  The index is an empty dictionary, which makes sense because in the case of that test it's a scalar.  But I worry that this could now be missing something useful that it used to catch.

wave_lang/kernel/wave/codegen/handlers.py
```diff
@@ -557,11 +556,12 @@ def handle_generic_atomic(emitter: WaveEmitter, node: fx.Node):
 
 
 def get_rank(mlir_type):
-    if not isinstance(mlir_type, ShapedType):
+    if hasattr(mlir_type, "rank"):
```
willghatch, line 559: This is another questionable change.  The vector types that I'm generating are not caught here.  Maybe I should find another type to test.  But this was also a mildly frustrating problem to track down to having a whitelisted set of values that didn't catch mine, so honestly I think I prefer the duck-type version.

wave_lang/kernel/lang/wave_types.py
```diff
@@ -138,6 +138,9 @@ def __class_getitem__(
             raise TypeError(f"Expected at least 2 arguments, got: {shape_and_dtype}")
 
         shape = shape_and_dtype[:-1]
+        # allow shape to be provided as a tuple instead of as individual elements, to work around lack of unpacking in subscripts for Python 3.10
+        if len(shape) == 1 and isinstance(shape[0], tuple):
+            shape = shape[0]
```
raikonenfnu, line 143: Is there a LIT test that show case this use case?

wave_lang/kernel/ops/wave_ops.py
```diff
@@ -1006,6 +1006,29 @@ def indexing_dims(self) -> list[IndexSymbol]:
...
+                        f"SelectOp requires operand shapes to be broadcastable. "
+                        f"Got condition: {c_shape}, if_true: {t_shape}, if_false: {f_shape}"
+                    )
+
+        # Find the most general shape
+        result_shape = ()
+        for t in [cond_type, t_type, f_type]:
+            if hasattr(t, "symbolic_shape"):
+                if len(t.symbolic_shape) >= len(result_shape):
+                    result_shape = t.symbolic_shape
```
raikonenfnu, line 1029: Thoughts on simplifying this by using previously defined `c_shape , t_shape, and f_shape`

raikonenfnu: or perhaps even the `non_empty_shapes`, something like:

```python
most_general_shape = max(non_empty_shapes, key=len, default=())
```

raikonenfnu (commented):
In the PR comment, can you add:

1. Motivation on why you created this PR
2. List of changes and it's intuition

wave_lang/kernel/lang/wave_types.py
willghatch: This is used in the new `infer_broadcast_shape`.  IE the type is `Register[broadcast_shape, dtype]`, because old versions of python still don't support `Register[*broadcast_shape, dtype]`.

wave_lang/kernel/ops/wave_ops.py
willghatch: Thanks, I missed this during my clean-up pass.

willghatch:
@raikonenfnu Thanks for the comments, and especially about a better place to do broadcasting.  Here is... well, it's basically a full rewrite to now do the broadcasting in index_sequencing_analysis.py.

wave_lang/kernel/wave/analysis/index_sequence_analysis.py
```diff
@@ -946,26 +949,109 @@ def apply_offset(node: fx.Node):
...
+        for operand in operands:
+            if operand["size"] < target["size"]:
+                if operand["size"] > 1 and target["size"] > 1:
+                    raise NotImplementedError(
+                        f"Currently only support resolving discrepancies when one of the shapes is 1."
+                        f"{generate_error_context()}"
+                    )
+
+                if operand["dim"] != target["dim"]:
+                    missing_dims = set(target["custom"].type.symbolic_shape).difference(
```
raikonenfnu, line 1034: Can we add back the comment
```
            # If the dimensions don't agree, we can still do this broadcast only if
            # the two nodes differ in shape along the broadcasting dimension and the
            # broadcasting dimension is the innermost dimension.
```

For better context for the reader on what's happening

wave_lang/kernel/wave/analysis/index_sequence_analysis.py
```diff
@@ -946,26 +949,109 @@ def apply_offset(node: fx.Node):
...
+            dim, size = get_largest_index_and_size(index, operand_custom)
+            operands.append(
+                {"custom": operand_custom, "dim": dim, "size": size, "id": identifier}
+            )
+
+        sizes = [operand["size"] for operand in operands]
+        if len(set(sizes)) <= 1 or all(size == 1 for size in sizes):
+            continue
+
+        target = max(operands, key=lambda x: x["size"])
```
raikonenfnu, line 1009: Would also be nice to add back some of the context on when/why we deem some cases done :)

wave_lang/kernel/ops/wave_ops.py
```diff
@@ -1008,6 +1009,30 @@ def indexing_dims(self) -> list[IndexSymbol]:
...
+        ):
+            raise ValueError(
+                f"SelectOp requires operand shapes to be broadcastable. "
+                f"Got condition: {c_shape}, if_true: {t_shape}, if_false: {f_shape}"
+            )
+
+        most_general_shape = ()
+        for s in [c_shape, t_shape, f_shape]:
+            if len(s) >= len(most_general_shape):
+                most_general_shape = s
```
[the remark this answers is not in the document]
willghatch: This was in your original feedback, sorry for not doing it then.  I didn't do it at first because it was failing, and I was confused why.  So I just left my loop in to make progress with everything else.  Ultimately looking again, I realized that the issue was due to name shadowing -- I was calling the locally defined Wave max function instead of the builtin max function.  So I have now put in the version using max.

willghatch:
Thanks @raikonenfnu , I've pushed an update addressing your comments.

wave_lang/kernel/wave/analysis/index_sequence_analysis.py
```diff
@@ -946,26 +949,110 @@ def apply_offset(node: fx.Node):
...
     broadcast_dim: IndexSymbol,
     broadcast_size: int,
     target_node: CustomOp,
 ):
     """
-    Create a broadcast node for the given binary operator.
+    Create a broadcast node for any multi-operand operator.
     """
-    with binary_op.graph.inserting_before(binary_op.fx_node):
+    to_broadcast = get_custom(getattr(op, to_broadcast_arg_name))
```
raikonenfnu, line 971: One final thing, can we use to_broadcast directly instead of trying to find it again with `get_custom(getattr(op, to_broadcast_arg_name))`. I was thinking we can change the call to `to_broadcast` with operand["custom"] instead of operand["id"], for e.g:
```
create_broadcast(
                    custom,
                    operand["custom"],
                    target["dim"],
                    target["size"],
                    target["custom"],
                )
```

willghatch: We do need the name for when we update later, IE `op.update_arg(to_broadcast_arg_name, custom.fx_node)`.  So I think I was trying to go for fewer arguments.  We could get the argument name from the custom op's position in the parent, but that seems harder.  I'll push a commit that adds another argument, and you can decide either way.

PR #255, opened 2025-09-02
[Wave][Reorder] Add more early exit checks for instr reordering.

raikonenfnu:
GEMM with direct global load to LDS is currently unsupported in schedule reordering, but was still getting through and having some pieces being reordered which cause numerical issue and slowness. 

This commit fixes it by adding extra checks.


harsh-nod (approved):
lgtm!

PR #257, opened 2025-09-02
Fix handling of empty KV splits in paged decode attention

paulzzy:
I think this approach can be improved since initialization with `device_full` is expensive. Should probably be done in the kernel itself.

### DIFF
--- a/wave_lang/kernel/wave/templates/paged_decode_attention.py
+++ b/wave_lang/kernel/wave/templates/paged_decode_attention.py
@@ -403,6 +403,9 @@ def get_paged_decode_attention_kernels(
             res = tkw.broadcast(res, target_shape=[U, S, N, B])
             tkw.write(res, output, mapping=logits_mapping)
 
+        # `sympy.Eq` is used because `==` tests for exact structural equality rather than symbolic equality
+        #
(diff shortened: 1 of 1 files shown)


paulzzy:
Initializing `phase_0_output` and `phase_0_output_max` with `torch.empty` is now safe. With empty splits, the kernel will write all `0` or all `-1e6` to `phase_0_output` and `phase_0_output_max`, respectively.

wave_lang/kernel/wave/templates/paged_decode_attention.py
```diff
@@ -403,6 +403,13 @@ def then():
             res = tkw.broadcast(res, target_shape=[U, S, N, B])
             tkw.write(res, output, mapping=logits_mapping)
 
+        @tkw.conditional(sympy.Eq(SPLIT_LEN, 0))
```
Hardcode84, line 406: Can we `SPLIT_LEN == 0`?

paulzzy: Unfortunately not, because SymPy treats `==` as [exact structural equality](https://docs.sympy.org/dev/tutorials/intro-tutorial/gotchas.html#equals-signs) rather than symbolic equality.

PR #259, opened 2025-09-02
Add a dedicated pass to partition gathers/scatters

Hardcode84:
Move gather/scatter decomposition into dedicated pass. This allows to cleanup a lot of code in read/write handlers and get rid of last traces of `vector.gather`/`scatter` there.

### DIFF
--- a/wave_lang/kernel/wave/analysis/partition_strided_operators.py
+++ b/wave_lang/kernel/wave/analysis/partition_strided_operators.py
@@ -439,6 +439,7 @@ def partition_gather_like_ops(trace: CapturedTrace, constraints: list[Constraint
                     raise NotImplementedError(f"Unsupported op type: {custom}")
 
                 # Update new_node information
+                custom.infer_type()
                 new_node.index = new_index
                 new_node.vector_shapes = custom.vector_shapes
                 ops_to_combine.append(new_node)
--- a/wave_lang/kernel/wave/analysis/partition_strided_operators.py
+++ b/wave_lang/kernel/wave/analysis/partition_strided_operators.py
@@ -202,6 +202,7 @@ def partition_strided_operators(trace: CapturedTrace, constraints: list[Constrai
                     )
                     for j, dim in enumerate(symbolic_shape)
                 }
+                extract.index = write.index
                 write.vector_shapes = vector_shapes
                 ops_to_combine.append(write)
 
@@ -329,6 +330,7 @@ def partition_ops_with_gpr_offsets(trace: CapturedTrace, constraints: list[Const
                             extract = ExtractSlice(
                                 dyn_val, [cur_gpr_start_id], [gpr_size], [1]
                             ).add_to_graph(custom.graph)
+                            extract.index = updated_index_with_gpr_offset
                             new_dynamic_vals.append(extract)
                         else:
                             new_dynamic_vals.append(dyn_val)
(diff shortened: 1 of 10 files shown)


wave_lang/kernel/wave/analysis/partition_strided_operators.py
```diff
@@ -380,3 +387,117 @@ def has_gpr_offsets(node: fx.Node) -> bool:
...
+    For example, if we have a write operation that writes elements with stride 2:
+    write([0,2,4,6]) -> write([0]), write([2]), write([4]), write([6])
+
+    The pass:
+    1. Identifies reads/writes with non-contiguous access patterns
+    2. For each such operation:
+       - Creates multiple single-element reads/writes
+       - Updates indices to access the correct elements
+       - Handles dynamic values in mappings
+       - Combines results back together if needed (for reads)
```
harsh-nod, line 406: Nice documentation!

wave_lang/kernel/wave/analysis/partition_strided_operators.py
```diff
@@ -380,3 +387,117 @@ def has_gpr_offsets(node: fx.Node) -> bool:
...
+            for i in range(elements_per_thread):
+                new_index = deepcopy(index)
+                for j, v in enumerate(new_index.values()):
+                    if j == fastest_index:
+                        v.start = v.start + i
+                    v.size = 1
+                    v.stride = 1
+
+                new_dynamic_vals = []
+                for dynamic_val in custom.mapping_dynamic_vals:
```
harsh-nod, line 439: When you partition an operator, what does that mean for dynamic values that are used by the operator? There are a couple of cases: 1) Single dynamic value in which case you can just use that, 2) N dynamic values where N is the the same as the elements_per_thread and in this case you need to extract the right element and anything else is unclear what we should do. If this is correct, can you add a comment describing that here?

wave_lang/kernel/wave/analysis/partition_strided_operators.py
```diff
@@ -380,3 +387,117 @@ def has_gpr_offsets(node: fx.Node) -> bool:
...
+                    extract.index = new_index
+                    new_node = Write(
+                        extract,
+                        custom.memory,
+                        mapping=custom.mapping,
+                        mapping_dynamic_vals=new_dynamic_vals,
+                        elements_per_thread=1,
+                    ).add_to_graph(custom.graph)
+                elif isinstance(custom, Read):
+                    # TODO: Add support on how to handle strided reads.
```
harsh-nod, line 469: Why won't the existing code handle strided reads? And what about strides writes?

wave_lang/kernel/wave/utils/mapping_utils.py
```diff
@@ -260,19 +279,62 @@ def check_is_mapping_contiguous(
...
+    new_index = transform_index_on_mapping(
+        mapping, symbolic_shape, index, is_read=is_read
+    )
+    prev_offset = _compute_offset(
+        [new_index[infer_dim(d)] for d in symbolic_shape], strides
+    )
+    for i in range(1, elements_per_thread):
+        new_index = deepcopy(index)
+        new_index[fastest_dim].start += i
+        new_index = transform_index_on_mapping(
```
harsh-nod, line 309: Instead of transforming the index multiple times, can we not do this just once with an index like: OLD_INDEX + V -> TRANSFORMED_INDEX and then substitute 1, 2, ..., elements_per_thread into this TRANSFORMED_INDEX for V?

harsh-nod (changes requested):
this should be a nice cleanup. thanks!

wave_lang/kernel/wave/utils/mapping_utils.py
Hardcode84: I don't think it will work, e.g. for the igemm case we are transforming index `m, n, k` -> `N, C, H, W` so no original symbols remains and we no longer know what symbol corresponds to the vector fastest dim. We need to substitute pre-transform.

PR #265, opened 2025-09-04
Always create `stream.binding.subspan` on func level

Hardcode84:
To avoid dominance errors when kernel arg is used in multiple scf constructs.


paulzzy (commented):
Would [`ThreadEmitter.lookup_node_values`](https://github.com/iree-org/iree-turbine/blob/4b8338591c9f5e63f2c789104b9301a517845bc3/iree/turbine/kernel/compiler/vector_codegen.py#L108-L114) also benefit from changing its insertion point?

Hardcode84:
> Would [`ThreadEmitter.lookup_node_values`](https://github.com/iree-org/iree-turbine/blob/4b8338591c9f5e63f2c789104b9301a517845bc3/iree/turbine/kernel/compiler/vector_codegen.py#L108-L114) also benefit from changing its insertion point?

Maybe, but we are not using it in Wave.

Hardcode84:
Also, need to remove remaining unused TK parts from Wave repo.

PR #266, opened 2025-09-04
Remove unused vector_codegen.py

paulzzy:
Remnants of the migration from iree-org/iree-turbine.

See https://github.com/iree-org/wave/pull/265#issuecomment-3254625629


Hardcode84:
Well, looks like, there are still some deps (for casting)

PR #269, opened 2025-09-04
Use dynamic values for AtomicOps

panditsa:
Add dynamic indexing support for atomic operations in MoE alignment

For the MoE align block, we need support for dynamically indexed values in atomic operations. This change introduces `mapping_dynamic_vals` parameter to AtomicOp and updates the atomic operation handler to support dynamic indexing.

Key changes:
- Add `mapping_dynamic_vals` parameter to atomic operations 
- Update `handle_atomic_op` decorator to process dynamic values
- Convert dynamic registers to index-typed vectors and extract scalar indices
- Pass dynamic value dictionary to `_build_start_indices` for proper index computation

This enables atomic operations like `atomic_add` to use runtime-computed indices from input data

### DIFF
--- a/wave_lang/kernel/ops/wave_ops.py
+++ b/wave_lang/kernel/ops/wave_ops.py
@@ -270,6 +270,16 @@ def atomic_min(
     rhs: "Memory",
     elements_per_thread: Optional[IndexExpr | int] = None,
     mapping: Optional[IndexMapping] = None,
+    mapping_dynamic_vals: "Register" | tuple["Register", ...] = (),
+) -> "Register": ...
+
+
+def atomic_add(
+    lhs: "Register",
+    rhs: "Memory",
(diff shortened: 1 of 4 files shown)


wave_lang/kernel/ops/wave_ops.py
```diff
@@ -1390,6 +1400,7 @@ class AtomicOp(BinaryOpBase, ABC):
 
     elements_per_thread: Optional[Any] = None
     mapping: Optional[IndexMapping] = None
+    mapping_dynamic_vals: tuple[fx.Node, ...] = ()
```
willghatch, line 1403: Unrelated to this PR exactly, but why does AtomicOp need to inherit from both BinaryOpBase and ABC, when BinaryOpBase has ABC in its inheritance chain?  Just curious.

PR #274, opened 2025-09-05
Scalarize packed math

Hardcode84:
Packed math can have a performance penalty near mfma ops on some hardware. Add an option to scalarize `addf`s/`mulf`s if there is an mfma in the same block during the codegen. AMDGPU backend ideally should be doing this for us but alas.

### DIFF
--- a/wave_lang/kernel/wave/codegen/handlers.py
+++ b/wave_lang/kernel/wave/codegen/handlers.py
@@ -576,6 +576,8 @@ _ops_to_scalarize = [
 
 
 def handle_binary_op(op):
+    captured_op = op
+
     def decorator(binary_fn: Callable[[Value, Value], OpResult]):
         @handle_op(op)
         def handle_generic_binary(emitter: WaveEmitter, node: fx.Node):
@@ -624,7 +626,7 @@ def handle_binary_op(op):
             #  `v_add_u32_e32 v1, s16, v12`, just not the "packed" variants.
             if (
                 emitter.options.scalarize_packed_math
-                and op in _ops_to_scalarize
+                and captured_op in _ops_to_scalarize
                 and _near_mma(node)
             ):
                 lhs_values = []


wave_lang/kernel/wave/codegen/handlers.py
```diff
@@ -596,7 +605,33 @@ def handle_generic_binary(emitter: WaveEmitter, node: fx.Node):
...
+                rhs_values = []
+                for i in range(rhs.type.shape[0]):
+                    rhs_elem = vector_d.extract(
+                        rhs, static_position=[i], dynamic_position=[]
+                    )
+                    rhs_values.append(rhs_elem)
+                values = []
+                for lhs_elem, rhs_elem in zip(lhs_values, rhs_values):
+                    result = binary_fn(lhs_elem, rhs_elem, emitter.options)
+                    values.append(result)
```
raikonenfnu, line 628: Can we apply loop fusion here?

wave_lang/kernel/wave/codegen/handlers.py
```diff
@@ -1669,7 +1704,20 @@ def handle_cast(emitter: WaveEmitter, node: fx.Node):
     conversion_op = get_conversion_op(
         src_elem_type, dst_elem_type, fastmath=get_fast_math_flags(emitter.options)
     )
-    casted_vector = conversion_op(dst_vector_type, vector_src)
+    if emitter.options.scalarize_packed_math and _near_mma(node):
```
raikonenfnu, line 1707: Thoughts of getting rid of `_near_mma` since we already giving kernel author the option over it by letting them set `scalarize_packed_math`?

wave_lang/kernel/wave/codegen/handlers.py
```diff
@@ -612,7 +647,7 @@ def get_fast_math_flags(options: WaveCompileOptions) -> int | None:
     return arith_d.FastMathFlags.fast if options.use_fast_math else None
 
 
-@handle_binary_op(operator.add)
+@handle_binary_op(operator.add, maybe_scalarize=True)
```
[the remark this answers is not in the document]
Hardcode84: Any bi op can scalarize, but not all will conflict with mfma (and triton only scalarize those)

wave_lang/kernel/wave/codegen/handlers.py
Hardcode84: I specifically dit it this way so in IR we have nice
```
vector.extract
vector.extract
vector.extract
vector.extract
vector.extract
vector.extract
vector.extract
vector.extract
arith.fadd
arith.fadd
arith.fadd
arith.fadd
```
Instead of mix of them

wave_lang/kernel/wave/codegen/handlers.py
Hardcode84: No point to scalarize if we don't have mfma

wave_lang/kernel/wave/codegen/handlers.py
raikonenfnu: how about we only loop fusion the lhs and rhs? haha

wave_lang/kernel/wave/codegen/handlers.py
```diff
@@ -596,7 +605,33 @@ def handle_generic_binary(emitter: WaveEmitter, node: fx.Node):
                     f"rhs={get_custom(op.rhs)}"
                 )
 
-            result = binary_fn(lhs, rhs, emitter.options)
```
raikonenfnu, line 599: It's a little tricky for me to track all these conditions to enable scalarize. I think it'd help if we make centralize all the conditions for scalarize together. 

Any thoughts on doing something like:
```
scalarizable_ops = {operator.add, operator.mul, operator.sub}
should_scalarize = emitter.options.scalarize_packed_math and op in scalarizable_op and _near_mma(node)
```
Then, we won't need maybe_scalarize in the templates :)

wave_lang/kernel/wave/codegen/handlers.py
raikonenfnu: cool cool, it's interesting to see that we also need scalarize for cast haha

wave_lang/kernel/wave/codegen/handlers.py
Hardcode84: Done (although, I had some annoyances with python lambdas)

PR #275, opened 2025-09-05
[Wave] Linearize shared memory accesses

raikonenfnu:
To promote better register usage, we can linearize shared memory accesses S.T the generated affine_map for local reads will be 1D which makes it easier for compiler to find a common subexpression between different vector.loads. 

This helps `ds_read` and `ds_write` to use more of the same base register and rely more on `offset:` field.

Changes brought from this PR/commit:

1. Introduce a `linearize_shared_access` option in WaveCompile

2. Implement a `linearize_index` util function that takes in IndexSequence and their strides and return a linearized IndexExpression.

3.  Implement a `linearize_shared_mem` in `read_write.py` to reinterpret N-d memref of shared memory into 1D. This is different than linearize_memref, because this is static, and since the input memref doesn't have a dynamic offset, we also do not need to specify offset.

4. Modified `read_write.py`'s `_create_vec_read_write` to generate linearized reads/writes iff memory is shared and has no mask and is not a scatter/gather.

### DIFF
--- a/lit_tests/kernel/wave/scaled_mma.py
+++ b/lit_tests/kernel/wave/scaled_mma.py
@@ -430,7 +430,7 @@ def test_mxfp4_scaled_mma_linearize_shrared_access():
     options = WaveCompileOptions(
         subs=hyperparams,
         canonicalize=True,
-        backend="rocm",
+        device="hip",
         target="gfx950",
         linearize_shared_access=True,
         compile_to_mlir=True,


wave_lang/kernel/wave/codegen/read_write.py
```diff
@@ -408,6 +409,25 @@ def _linearize_memref(
     )
 
 
+def _linearize_shared_mem(memory: CustomOp):
```
Hardcode84, line 412: This won't work with the dynamic symbols. Can you also add a test for the dynamic dim?

wave_lang/kernel/wave/utils/general_utils.py
```diff
@@ -71,6 +71,20 @@ def get_default_scheduling_params() -> dict[IndexSymbol, Any]:
     }
 
 
+def linearize_index(expr_list: dict[IndexExpr, IndexSequence], strides: list[int]):
+    assert len(expr_list) == len(strides)
+    linear_expr = None
+    for expr, stride in zip(expr_list.values(), strides):
+        if not isinstance(expr, IndexSequence):
+            return None
```
Hardcode84, line 79: We should probably raise an error here instead of returning `None`.

wave_lang/kernel/wave/utils/general_utils.py
```diff
@@ -71,6 +71,20 @@ def get_default_scheduling_params() -> dict[IndexSymbol, Any]:
     }
 
 
+def linearize_index(expr_list: dict[IndexExpr, IndexSequence], strides: list[int]):
```
Hardcode84, line 74: Missing return type.

Hardcode84:
Also, why linearize only for shared mem?

PR #277, opened 2025-09-09
implement TopkOp

willghatch:
This implements a TopkOp with a similar shape to the ReduceOp.

The algorithm it uses is similar to https://github.com/sgl-project/sglang/blob/6337d9057c78fc29b31d15bec892bb9013201f6d/sgl-kernel/csrc/moe/moe_topk_softmax_kernels.cu#L185 -- it does a reduction to find the maximum value and index, then masks the value at that index and repeats.

There are some things cargo-culted from the implementation of ReduceOp.  I don't yet understand the placeholder stuff that is done when creating a conditional region during the rewriting.

This PR depends on https://github.com/iree-org/wave/pull/251


willghatch:
@harsh-nod ping, I've made all of the requested changes, please take another look.  One of the requested changes maybe doesn't matter anymore, since another change took it out.  IE I wrote some functions to simplify making subgraphs, but after taking out writing to shared memory topk also doesn't use subgraphs anymore.  We can keep the new subgraph-related functions and changes to ReduceOp, or we can drop them.

wave_lang/kernel/wave/analysis/index_sequence_analysis.py
```diff
@@ -983,78 +989,85 @@ def create_broadcast(
...
+        dim, size = get_largest_index_and_size(index, operand_custom)
+        operands.append(
+            {"custom": operand_custom, "dim": dim, "size": size, "id": identifier}
+        )
 
-            index = get_index(operand_custom)
-            dim, size = get_largest_index_and_size(index, operand_custom)
-            operands.append(
-                {"custom": operand_custom, "dim": dim, "size": size, "id": identifier}
+    sizes = [operand["size"] for operand in operands]
```
harsh-nod, line 1019: Is sizes being used?

wave_lang/kernel/wave/codegen/handlers.py
```diff
@@ -1684,7 +1684,21 @@ def handle_select(emitter: WaveEmitter, node: fx.Node):
...
 
     unwrap = lambda x: cast_py_value(emitter, x).ir_value
-    selected = arith_d.select(unwrap(cond), unwrap(if_true), unwrap(if_false))
+    cond = unwrap(cond)
+    if_true = unwrap(if_true)
+    if_false = unwrap(if_false)
+
+    # Handle special scalar/rank-0 cases where if_true/if_false may be
+    # Dtype, vector<Dtype>, or vector<1xDtype>.
+    true_false_ranks = [get_rank(arg.type) for arg in (if_true, if_false)]
```
harsh-nod, line 1693: This code is duplicated in several places in the handlers (see also in handle_generic_binary). Could you pull this into a common function that can be reused?

wave_lang/kernel/wave/decompose_topk_ops.py
```diff
@@ -0,0 +1,615 @@
...
+) -> tuple[fx.Node, fx.Node]:
+    """
+    Reduce data across threads in a wave by doing butterfly shuffle for topk.
+    """
+    init_val = src_val
+    init_idx = src_idx
+    num_steps = int(math.log2(float(cluster_size)))
+
+    for _ in range(num_steps):
+        # Shuffle both values and indices
```
harsh-nod, line 180: No action required, but maybe you can add a TODO - if the values are f16 and the indices are <= f16, we could pack them together and to a single shuffle with both and this would reduce one extra shuffle required.

harsh-nod (approved):
Most of my comments are refactors, so after addressing those and adding a lit test, please feel free to merge. Thanks this was a big first commit but a good one!

PR #285, opened 2025-09-09
Add support for hardware transpose operation

ashay:
This PR contains five patches for adding transpose support on gfx950 machines:

> Use string instead of index for updating the write dependency
> 
> Prior to this patch, adding new fields at the end of the ReadOp type
> introduced problems because the code subtly changed the newly introduced
> field instead of the write dependency even though it was asked to update
> just the write dependency.  This patch fixes the problem by using the
> name of the field instead of the index.

and

> Update discovery of optimizable nodes
> 
> Instead of sharing the analysis that discovers the nodes to optimize in
> the global-to-shared-gathers pass with the minimize-global-loads pass
> (which requires a more restrictive analysis), this patch introduces a
> separate function for the global-to-shared-gathers.

and

> Add support for debug-dumping i8 types

and

> Skip the in-thread-transpose pass when compiling for gfx95X chips
> 
> We do this because the hardware-transpose pass (that we'll add in the
> next commit) will supersede the compilation on MI350+ machines.

and

> Add support for hardware transpose operation
> 
> This patch enables the emission of `amdgpu.transpose_load` operations on
> gfx950 machines based on the underlying read operation.  We deliberately
> decouple the transpose load op generation from the use of MFMA
> instructions so that regardless of whether the results of the read
> operation are fed to MFMA operations, they are rewritten to use the
> transpose operation.

Credit to Eric Feng for the initial changes that were key to constructing these patches.

### DIFF
--- a/wave_lang/kernel/ops/wave_ops.py
+++ b/wave_lang/kernel/ops/wave_ops.py
@@ -31,15 +31,57 @@ from ..lang.global_symbols import *
 from ..lang.kernel_buffer import AddressSpace
 from ..lang.wave_types import IndexMapping, Memory, Register
 from .base import OpDispatcher
+from ..wave.constraints import Constraint
 
 if TYPE_CHECKING:
-    from ..wave.constraints import Constraint
     from ..wave.scheduling.resources import Operation
 
 T = TypeVar("T", bound=Type[Any])
 AccT = TypeVar("AccT")
 CustomOpT = TypeVar("CustomOpT", bound="CustomOp")
 PlaceholderT = TypeVar("PlaceholderT", bound="Placeholder")
+
+
+def read_meets_hw_transpose_requirements(
+    read: Read, constraints: list[Constraint]
+) -> bool:
+    from ..wave.minimize_global_loads import is_transposed_read
+    from ..wave.utils.general_utils import find_index_bounds
+
+    if read.bounds is not None:
+        return False
+
+    if len(read.type.symbolic_shape) <= 1:
+        return False
+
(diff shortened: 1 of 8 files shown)


wave_lang/kernel/ops/wave_ops.py
```diff
@@ -1653,6 +1653,7 @@ class Read(CustomOp):
     source: Optional[tuple[IndexExpr]] = None
     target: Optional[tuple[IndexExpr]] = None
     _write_dependency: Optional[list[fx.Node]] = None
+    transpose: bool = False
```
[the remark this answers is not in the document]
Hardcode84: Our read/write ops are already overloaded with different semantics, including mapping, which can technically encode the same information.

wave_lang/kernel/ops/wave_ops.py
```diff
@@ -40,6 +40,41 @@
...
+        constraints, read.index, read.vector_shapes, read.type.symbolic_shape
+    )
+
+    if bounds is not None:
+        return False
+
+    if read.has_identity_mapping():
+        return False
+
+    if len(list(read.index.keys())) != 2:
```
Hardcode84, line 61: This check looks incomplete, should we check mapping corresponds to the transpose?

Hardcode84: Now I'm thinking separate op or ReadOp flag op weren't the bad ideas :). Anyway, lets merge this in current form to unblock things and see what we can improve later.

Hardcode84 (approved):

PR #291, opened 2025-09-10
add locations to placeholder and output nodes based on kernel location

willghatch:
This allows the initial trace of a kernel to have locations on 100% of nodes, at least for examples that I'm looking at so far.

### DIFF
--- a/wave_lang/kernel/_support/location.py
+++ b/wave_lang/kernel/_support/location.py
@@ -7,7 +7,7 @@
 import inspect
 import sys
 from dataclasses import dataclass
-from typing import List, Optional, Union
+from typing import List, Optional, Union, Callable
 
 from iree.compiler.ir import Context, Location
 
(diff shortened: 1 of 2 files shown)


harsh-amd (changes requested):
Could you add a test for this?

wave_lang/kernel/_support/location.py
```diff
@@ -107,3 +107,85 @@ def capture_location(
...
+    if kernel_location is None:
+        return trace
+
+    for node in root_graph.nodes:
+        custom_op = get_custom(node)
+        if hasattr(custom_op, "location") and custom_op.location is not None:
+            continue
+
+        if isinstance(custom_op, (Placeholder, Output)):
+            custom_op.fx_node.location = kernel_location
```
ftynse, line 189: I'm confused by the check above looking at `custom_op.location` while this is setting a `custom_op.fx_node.location`.

PR #292, opened 2025-09-10
add location_check_pass

willghatch:
This adds a pass that makes no changes to the trace, but that logs information about debug locations.

Note that this PR doesn't actually add the pass to the pipeline, it just adds the pass so that it can be used.  In working on locations I'm using this pass, and I figure I should just check it in.

### DIFF
--- a/wave_lang/kernel/wave/location_check_pass.py
+++ b/wave_lang/kernel/wave/location_check_pass.py
@@ -1,0 +1,60 @@
+# Copyright 2025 The IREE Authors
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+from typing import List
+from wave_lang.support.logging import get_logger
+from .._support.tracing import CapturedTrace
+from ..ops.wave_ops import get_custom, CustomOp
+from typing import Optional
+
+logger = get_logger("wave.ops_location_check")
+
+
+def location_check_pass(
+    trace: CapturedTrace,
+    pass_name: str = "unnamed",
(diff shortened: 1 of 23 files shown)


wave_lang/kernel/wave/location_check_pass.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    ops_with_location: List[CustomOp] = []
+    ops_without_location: List[CustomOp] = []
+    log_messages = []
+
+    for node in trace.get_root_graph().nodes:
+        custom_op = get_custom(node)
+        location = getattr(custom_op, "location", None)
+        if print_locations == "all":
+            log_messages.append(f"  op: {node} location: {location}")
```
harsh-nod, line 43: Why not just do logger.info here?

willghatch: For nicer log formatting, I wanted the log to put the summary at the top then the details below.

wave_lang/kernel/wave/wave.py
```diff
@@ -813,6 +814,16 @@ def _trace_and_get_kernel_signature(
             partial(generate_bounds_exprs, trace, self.constraints),
         ]
 
+        graph_passes.append(
+            partial(
+                location_check_pass,
+                trace,
+                "enforce-locations",
+                log=False,
+                enforce_100=options.enforce_locations,
```
xintin, line 823: Question: what is enforce_100?

willghatch: Enforce 100% locations.  I can rename that -- the name makes sense in the context of the pass, but for external users, probably `enforce_locations` is better.

xintin:
LGTM! Left a couple of comments. Should be good to merge after that.
Thanks :)

PR #295, opened 2025-09-12
[CI] Add GitHub Actions workflow: self-hosted RDNA4

Megan0704-1:
### Summary of changes:
- Add GitHub Actions workflow for self-hosted RDNA device. The workflow is currently manually triggered because GEMM tests for this device are not yet supported.
- Require CDNA for end-to-end tests. Note: WaveCompileOptions now defaults to emitting ISA gfx942 (see wave_lang/kernel/wave/compile_options.py), so tests expecting MFMA must be gated or updated.


.github/workflows/ci-gpu-rdna.yaml
```diff
@@ -0,0 +1,123 @@
...
+      fail-fast: false
+      matrix:
+        version: ["3.11"]
+        os: [Shark49]
+    runs-on: [self-hosted, Linux, X64, rdna4, shark49]
+    timeout-minutes: 60
+    if: github.event_name != 'pull_request' || github.event.pull_request.draft == false
+
+    container:
+      image: 'ghcr.io/rocm/no_rocm_image_ubuntu24_04@sha256:405945a40deaff9db90b9839c0f41d4cba4a383c1a7459b28627047bf6302a26'
```
harsh-nod, line 39: why do you need a new rocm image? can we not use one of the existing images?

harsh-nod (approved):
thanks for setting this up!

.github/workflows/ci-gpu-rdna.yaml
Megan0704-1: Thanks for the review! 
This uses rocm's published `no-rocm` Ubuntu 24.04 (existing images are Jammy, 22.04) base pinned by digest. I chose it so I can install the RDNA4 userspace via the pinned pip indexes and avoid a driver userspace mismatch on the self-hosted runner. It also keeps the surface reproducible so easier to debug.

PR #296, opened 2025-09-12
[debugging] add location propagation to replace_all_uses_with

willghatch:
This prevents the location loss that previously happened with this method.


willghatch:
This now depends on https://github.com/iree-org/wave/pull/297 for the custom op location setter.

ftynse (approved):
Would be nice to have a test, but probably difficult without proper pass management and a quick way to create test passes.

PR #297, opened 2025-09-12
[debugging] propagate locations

willghatch:
This patch fixes many places where locations were dropped during compilation. It's not necessarily exhaustive, but it gets large test cases like the codegen.py test to 100% location coverage at the end of the Wave pipeline.

### DIFF
--- a/lit_tests/kernel/wave/specific_location.py
+++ b/lit_tests/kernel/wave/specific_location.py
@@ -1,0 +1,92 @@
+# RUN: python %s | FileCheck %s
+
+# This file contains location tests that check exact source line and column numbers.
+# These tests are sensitive to minor edits and line number changes.
+
+import wave_lang.kernel.lang as tkl
+import wave_lang.kernel.wave as tkw
+from wave_lang.support.location_config import (
+    LocationCaptureConfig,
+    LocationCaptureLevel,
+)
+from wave_lang.kernel.wave.compile import WaveCompileOptions, wave_compile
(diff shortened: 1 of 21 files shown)


willghatch:
I've increased the scope of this PR to include all of the other location propagation that I've done since.

xintin:
Do we need a separate file for `specific_location` test? Are we assuming this file will not evolve over time? 
Just a thought: Is it possible to include the test in `location.py` and use FileCheck numeric variable to express locations relative to the check line?

willghatch:
We discussed adding an optional location field to the base class, but to do that we also need to add a location field to every op constructor, since our op construction is sort of indirect and weird.  Instead, I propose adding a location argument to the `add_to_graph` method, since that is another method that we always use when making ops during the compiler.  Thoughts @harsh-nod @xintin ?

lit_tests/kernel/wave/location.py
```diff
@@ -222,3 +222,68 @@ def test_stack_trace_dedup(
...
+        )
+    ]
+    constraints += [tkw.WorkgroupConstraint(M, BLOCK_M, 1)]
+    constraints += [tkw.WaveConstraint(M, BLOCK_M)]
+
+    subs = {
+        M: 256,
+        K: 128,
+        BLOCK_M: 1,
+        ELEMS_PER_THREAD: 2,
```
xintin, line 244: Just curious, if we do not pass elems per thread here, is the compiler able to deduce it implicitly?

xintin (approved):

PR #302, opened 2025-09-16
Distributed GEMM

panditsa:
Enable distributed matrix multiplication operations across multiple GPU devices in wave, where the distribution factor for each dimension is managed through DeviceConstraint

### Key Changes

- In the host_codegen.py, we split the input tensors across devices based on device constraint, dispatch computation to each device, and merge results back into the original tensor shape
- We now use a `HostSignature` class to manage full problem-size buffers (e.g., 1024x8192 matrix), while `KernelSignature` handles per-device tile buffers (e.g., 512x4096 tiles for 2x2 distribution). Non-distributed workloads have no impact, because in absence of `DeviceConstraint`, `HostSignature` is the same as` KernelSignature`
- In the `host_utils.py`, added functions that help split input and output tensors per device, manage tensor distribution across devices using `device_constraint_map`.
- Updated distributed GEMM template to accept` device_m` and `device_n` parameters, with test coverage for distribution factors from 1x1 up to 4x2 across problem sizes up to 4096x20480x2560
- Use MultiDeviceLaunchable class to orchestrate execution across multiple GPU devices


Hardcode84 (commented):
Any lit test for codegen changes? Also, please fix merge conflicts.

Hardcode84:
Please fix the DCO and merge conflicts.

wave_lang/kernel/compiler/host_utils.py
```diff
@@ -0,0 +1,369 @@
+from typing import Optional
```
Hardcode84, line 1: Missing license header

Hardcode84 (approved):

PR #308, opened 2025-09-19
Run external MLIR module through Wave

tyb0807:
Currently we still miss some corner cases (that I don't see from the examples that I've run so far) regarding the changes we have to make to the external, purely upstream MLIR code. These cases are handled here https://github.com/iree-org/wave/blob/0a596e2caaaa51175ddc1b52847ac17341a60d46/wave_lang/kernel/compiler/dispatch_codegen.py#L313-L336. Currently we only update the code to use the `stream.binding` arguments.

### DIFF
--- a/tests/kernel/wave/attention/chained_gemm_test.py
+++ b/tests/kernel/wave/attention/chained_gemm_test.py
@@ -124,7 +124,7 @@ def testChainedGemm(
 
     batch, q_seq_len, v_head_dim, qk_head_dim, kv_seq_len = shape
     hyperparams = {
-        ADDRESS_SPACE: SHARED_ADDRESS_SPACE,
+        ADDRESS_SPACE: GLOBAL_ADDRESS_SPACE,
         LOAD_ELEMS_PER_THREAD: get_mfma_load_elems_per_thread(mfma_variant),
         STORE_ELEMS_PER_THREAD: get_mfma_store_elems_per_thread(mfma_variant),
         BLOCK_B: 1,
--- a/wave_lang/kernel/wave/compile.py
+++ b/wave_lang/kernel/wave/compile.py
@@ -5,6 +5,7 @@ from typing import Any, Optional, Callable, Sequence
 import torch
 
 from wave_lang.kernel.lang import IndexSymbol
+from wave_lang.support.ir_imports import Module
 
 from .._support.indexing import IndexingContext
 from ...support.location_config import LocationCaptureLevel
@@ -24,7 +25,7 @@ from .utils.run_utils import (
     invoke_with_wave_runtime,
     get_benchmark_flags,
 )
-from .water import water_leak_in_bounds_check
+from .water import water_leak_in_bounds_check, _deiree
 from wave_lang.runtime.launch import Launchable
 from .profiling import benchmark_module
 from .debug_log_hoist import DebugArgInfo
@@ -323,6 +324,16 @@ def wave_compile(options: WaveCompileOptions, kernel: "LaunchableWave") -> WaveK
             debug_handlers,
             device_layout,
         ) = kernel._trace_and_get_kernel_signature(options)
+
+        module_op = Module.parse(_deiree(mb.module_op), context=mb.module_op.context)
+        (
+            mb,
+            _,
+            exe,
+            kernel_sig,
+            entrypoint_name,
+        ) = kernel.compile_to_mlir(trace=graph, context=None, module_op=module_op, options=options)
+
         options.kernel_sig = kernel_sig
 
         # Get the trace from the kernel. Since the trace contains complex objects
--- a/wave_lang/kernel/wave/wave.py
+++ b/wave_lang/kernel/wave/wave.py
@@ -17,7 +17,7 @@ import sympy
 import torch.fx as fx
 from sympy.utilities.lambdify import lambdastr
 
-from wave_lang.support.ir_imports import Context, Module, Operation
+from wave_lang.support.ir_imports import Context, Module, Operation, InsertionPoint, stream_d, IndexType, IntegerAttr, arith_d
 
 from .._support.indexing import IndexExpr, IndexingContext, index_symbol
 from ...support.location_config import LocationCaptureConfig
@@ -534,6 +534,7 @@ class LaunchableWave(Launchable):
         entrypoint_name = self._name
         root_graph = trace.get_root_graph()
 
+        print(module_op)
         # pass device constraint to kernel signature
         # so that we can set the dimensions of the tensors per device
         kernel_sig = kernel_codegen.KernelSignature(self.device_constraints)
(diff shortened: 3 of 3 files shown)


tests/kernel/wave/attention/chained_gemm_test.py
```diff
@@ -124,7 +124,7 @@ def repeat(
 
     batch, q_seq_len, v_head_dim, qk_head_dim, kv_seq_len = shape
     hyperparams = {
-        ADDRESS_SPACE: SHARED_ADDRESS_SPACE,
+        ADDRESS_SPACE: GLOBAL_ADDRESS_SPACE,
```
ftynse, line 127: We probably don't want to change this

wave_lang/kernel/wave/wave.py
```diff
@@ -534,6 +534,7 @@ def compile_to_mlir(
         entrypoint_name = self._name
         root_graph = trace.get_root_graph()
 
+        print(module_op)
```
ftynse, line 537: Lefover debug

wave_lang/kernel/wave/wave.py
```diff
@@ -557,6 +558,7 @@ def compile_to_mlir(
         if options.waves_per_eu:
             llvm_func_config["amdgpu-waves-per-eu"] = options.waves_per_eu
 
+        print("BEFORE ", mb.module_op.get_asm())
```
ftynse, line 561:
```suggestion
```

wave_lang/kernel/wave/wave.py
```diff
@@ -567,18 +569,64 @@ def compile_to_mlir(
...
-        )
-        try:
-            emitter.emit(trace.get_root_graph())
-        except:
-            logger.info("Error in emitter")
-            asm = mb.module_op.get_asm()
-            logger.info(asm)
-            raise
-        emitter.finish()
+        print("AFTER ", mb.module_op.get_asm())
```
ftynse, line 572:
```suggestion
```

wave_lang/kernel/wave/compile.py
```diff
@@ -323,6 +324,16 @@ def get_binary_path():
             debug_handlers,
             device_layout,
         ) = kernel._trace_and_get_kernel_signature(options)
+
+        module_op = Module.parse(_deiree(mb.module_op), context=mb.module_op.context)
```
ftynse, line 328: Could you add a comment as to why we need to re-parse a module here?

wave_lang/kernel/wave/wave.py
```diff
@@ -567,18 +569,64 @@ def compile_to_mlir(
...
+                for op in ops_to_move:
+                    op.detach_from_parent()
+                    target_block.append(op)
+
+                # Replace all uses of old arguments with new subspan results
+                for old_arg, new_value in arg_mapping.items():
+                    old_arg.replace_all_uses_with(new_value)
+
 
+        print("FINAL ", mb.module_op.get_asm())
```
ftynse, line 629:
```suggestion
```

(maybe we can use a specialized logger for these)

wave_lang/kernel/wave/wave.py
```diff
@@ -567,18 +569,64 @@ def compile_to_mlir(
...
+            except:
+                logger.info("Error in emitter")
+                asm = mb.module_op.get_asm()
+                logger.info(asm)
+                raise
+            emitter.finish()
+        else:
+            with exe._loc, InsertionPoint.at_block_begin(dispatch_entrypoint.entry_block):
+                target_block = dispatch_entrypoint.entry_block
+                source_func_op = list(module_op.operation.regions[0].blocks[0])[0]
```
ftynse, line 588: Nit: isn't there something like `.operartions` on a block that avoids materializing the entire list?

wave_lang/kernel/wave/wave.py
```diff
@@ -567,18 +569,64 @@ def compile_to_mlir(
...
+                source_block = source_func_op.regions[0].blocks[0]
+
+                source_args = list(source_block.arguments)
+                target_args = list(target_block.arguments)
+
+                def convert_memref_to_stream_binding(target_block, old_arg, new_arg, index):
```
ftynse, line 594: type annotations plz

wave_lang/kernel/wave/wave.py
```diff
@@ -567,18 +569,64 @@ def compile_to_mlir(
...
+                    result_type = IndexType.get()
+                    zero_value = arith_d.constant(result_type, IntegerAttr.get(result_type, 0))
+
+                    # Create subspan operation
+                    subspan_op = stream_d.binding_subspan(
+                        old_arg.type,  # The original memref type
+                        new_arg,       # The stream.binding argument
+                        byte_offset=zero_value,
+                      # dynamic_dims=dispatch_entrypoint.get_dynamic_dims(binding),
+                        dynamic_dims=[], # TODO: get dynamic dims
```
ftynse, line 606: Can we error out when dynamic dims are present instead of silently dropping them?

wave_lang/kernel/wave/wave.py
```diff
@@ -567,18 +569,64 @@ def compile_to_mlir(
...
+                    )
+
+                    return subspan_op
+
+                # Create argument mapping
+                arg_mapping = {}
+                for i, old_arg in enumerate(source_args):
+                    if i < len(target_args) and "memref" in str(old_arg.type):
```
ftynse, line 614: Wouldn't `isinstance(old_arg.type, MemRefType)` work? I'd rather avoid strings

wave_lang/kernel/wave/wave.py
```diff
@@ -567,18 +569,64 @@ def compile_to_mlir(
...
+                        new_subspan = convert_memref_to_stream_binding(target_block, old_arg, target_args[i], i)
+                        arg_mapping[old_arg] = new_subspan
+
+                # Move operations
+                ops_to_move = list(source_block)
```
ftynse, line 619: Isn't `source_block` iterable? If detaching ops messing with some internal state, please document.

wave_lang/kernel/wave/compile.py
```diff
@@ -366,8 +389,8 @@ def get_binary_path():
                 ),
             )
 
-        if options.override_mlir:
-            asm = options.override_mlir
+        # if options.override_mlir:
+        #     asm = options.override_mlir
```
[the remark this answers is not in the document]
tyb0807: No because we've created a new `mb` for the overriding MLIR, so `asm = mb.module_op.get_asm(...)` would have the right IR anyway.

wave_lang/kernel/wave/compile.py
```diff
@@ -323,6 +324,28 @@ def get_binary_path():
...
         ) = kernel._trace_and_get_kernel_signature(options)
+
+        if options.override_mlir:
+            overriding_module_op = Module.parse(
+                options.override_mlir, context=mb.module_op.context
+            )
+            if not isinstance(
+                overriding_module_op.operation.regions[0].blocks[0].operations[0],
+                stream_d.ExecutableOp,
+            ):
```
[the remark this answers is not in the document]
tyb0807: I'll do a sweep over all top-level ops of the module.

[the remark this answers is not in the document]
tyb0807: Maybe it's clearer to have an additional flag to tell if we want to iree-fy the module or not?

PR #338, opened 2025-10-03
import the second batch of the wave dialect commits

ftynse:
Please review per commit


water/lib/Dialect/Wave/IR/WaveAttrs.cpp
```diff
@@ -165,7 +165,7 @@ bool WaveHyperparameterAttr::hasSymbol(StringRef symbolName) const {
 }
 
 //===----------------------------------------------------------------------===//
-// WaveDistributedShapeAttr
+// DistributedShapeAttr
```
harsh-nod, line 168: The rest of the code references this as WaveDistributedShapeAttr so why this change?

PR #350, opened 2025-10-09
Remove aot code

Hardcode84:
Remove turbine code remains.


Hardcode84:
this code may still be used in sharktank repo, check with them before merging

PR #367, opened 2025-10-20
Standalone examples

panditsa:
This PR adds standalone Python examples to make Wave Lang more accessible for new users. Currently, most examples exist within the pytest framework or as lit tests, making them difficult to discover and use for quick experimentation. These focused, runnable examples demonstrate individual language features and can be easily run and modified without test framework overhead.

### DIFF
--- a/examples/python/5_gemm.py
+++ b/examples/python/5_gemm.py
@@ -67,7 +67,7 @@ def simple_gemm_test(is_debug=False):
     ]
 
     @tkw.wave(constraints)
-    def gemm(
+    def wave_kernel(
         a: Memory[M, K, ADDRESS_SPACE_A, f16],  # Input matrix A
         b: Memory[N, K, ADDRESS_SPACE_B, f16],  # Input matrix B
         c: Memory[M, N, ADDRESS_SPACE_C, f32],  # Output matrix C
@@ -117,7 +117,7 @@ def simple_gemm_test(is_debug=False):
         print_ir_after="all" if is_debug else [],
     )
     options = set_default_run_config(options)
-    compiled_gemm = wave_compile(options, gemm)
+    compiled_gemm = wave_compile(options, wave_kernel)
 
     if is_debug:
         print(compiled_gemm.asm)
@@ -162,7 +162,7 @@ def test_gemm_select_expert(is_debug=False):
     )
 
     @tkw.wave(constraints)
-    def gemm(
+    def wave_kernel(
         a: Memory[M, K, ADDRESS_SPACE_A, f16],  # Input matrix A
         b: Memory[E, N, K, ADDRESS_SPACE_B, f16],  # Input matrix B
         c: Memory[M, N, ADDRESS_SPACE_C, f32],  # Output matrix C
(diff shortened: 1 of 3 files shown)


examples/python/1_dynamic_mapping.py
```diff
@@ -0,0 +1,206 @@
...
+        subs=bindings,
+        canonicalize=canonicalize,
+        dynamic_symbols=dynamic_symbols,
+        print_ir_after="all" if is_debug else [],
+    )
+
+
+def test_dynamic_offset(is_debug=False):
+    """1D tensor read with a single dynamic offset value, showcasing simpler offset-based indexing."""
+    ONE = tkl.sym.ONE
```
tgymnich: does this need to be a symbol?

panditsa: Not necessarily, you can just pass a scalar `tkl.i32` type for the offset and pass that scalar directly to the dynamic value. I'll add that as another example to the list!

```
def test_dynamic_offset(is_debug=False):
    """1D tensor read with a single dynamic offset value, showcasing simpler offset-based indexing."""
    constraints: list[tkw.Constraint] = [
        tkw.HardwareConstraint(
            threads_per_wave=64,
            vector_shapes={N: 16},
        )
    ]
    constraints += [tkw.WorkgroupConstraint(N, BLOCK_N, 1)]
    constraints += [tkw.WaveConstraint(N, BLOCK_N)]

    j = tkw.IndexMapping.iterator(0)
    k = tkw.IndexMapping.dynamic_val(0)
    mapping = tkw.IndexMapping(
        num_iterators=1,
        inputs={N: (k + j) % 16},
        outputs={N: j},
        dynamic_val_mappings={N: j},
    )

    @tkw.wave(constraints)
    def kernel(
        a: tkl.Memory[N, ADDRESS_SPACE, tkl.i32],
        off: tkl.i32,
        b: tkl.Memory[N, ADDRESS_SPACE, tkl.i32],
    ):

        res = tkw.read(
            a,
            mapping=mapping,
            mapping_dynamic_vals=(off,),
        )
        tkw.write(res, b)

    compiled_kernel = wave_compile(
        get_wave_compile_options(
            canonicalize=True, is_debug=is_debug
        ),
        kernel,
    )
    if is_debug:
        print(compiled_kernel.asm)

    # create input tensors
    a = (
        torch.arange(0, 16, dtype=torch.int32)
        .reshape(
            16,
        )
        .cuda()
    )
    off = 1 
    b = torch.zeros((16,), dtype=torch.int32).cuda()

    compiled_kernel(a, off, b)
    print(a)
    print(off)
    print(b)
```

examples/python/5_gemm.py
```diff
@@ -0,0 +1,1779 @@
...
+        subs=hyperparams,
+        print_ir_after="all" if is_debug else [],
+    )
+    options = set_default_run_config(options)
+    compiled_gemm = wave_compile(options, gemm)
+
+    if is_debug:
+        print(compiled_gemm.asm)
+        with open("gemm.mlir", "w") as f:
+            f.write(compiled_gemm.asm)
```
ashay: Perhaps drop the file writes, so that they don't pollute the directory for folks running these examples.

PR #368, opened 2025-10-20
Keep all the dims for scan-op

panditsa:
The dimension over which we implement scan-op is not reduced-away. With this change, we always output in the same shape as input, regardless of which dimension we are implementing scan-op over. This also enables implementing cumsum over a vector.

### DIFF
--- a/lit_tests/kernel/wave/mlir_converter.py
+++ b/lit_tests/kernel/wave/mlir_converter.py
@@ -243,9 +243,21 @@ def mlir_converter_matmul():
     # CHECK-NEXT:     %[[READ_SHARED_B_2:.*]] = wave.read %[[ALLOCATE_2]]
     # CHECK-NEXT:     %[[READ_SHARED_B_3:.*]] = wave.read %[[ALLOCATE_2]]
     # CHECK-NEXT:     %[[MMA_0:.*]] = wave.mma %[[READ_SHARED_B_0]], %[[READ_SHARED_A_0]], %[[ARG3]]
+    # CHECK-COUNT-2:  {K : [
+    # CHECK-SAME:     {M : [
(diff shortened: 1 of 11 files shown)


wave_lang/kernel/wave/decompose_scan_ops.py
```diff
@@ -343,4 +342,5 @@ def decompose_scan_ops(
                     custom.fx_node, global_scan[user.expanded_dims[scan_dim]]
                 )
 
+        custom.graph.erase_node(custom.fx_node)
```
xintin: why are we erasing this node?

xintin (changes requested):

wave_lang/kernel/wave/decompose_scan_ops.py
```diff
@@ -118,8 +117,16 @@ def emit_global_scan(
...
-    target_shape.pop(target_shape.index(scan_dim))
-    scanop_result.index = {target_shape[0]: get_custom(src).index[target_shape[0]]}
+    src_index = get_custom(src).index
+
+    # Keep indices for all dimensions EXCEPT scan_dim
+    scanop_result.index = {
+        dim: src_index[dim] for dim in target_shape if dim != scan_dim
+    }
+
+    # If no indices are left, set the index to the scan_dim with size 1 (?)
```
panditsa: @xintin When dealing with 1D cumsum, removing the scan dimension from target_shape results in an empty index, which causes the codegen to fail when determining vector shapes. I've added a fallback to set `scanop_result.index = {scan_dim: IndexSequence(0, 1, 1)}` when the index would be empty, which resolves the issue. However, I'm not entirely sure if this is the correct approach or if there's a better way to handle the 1D case.

PR #370, opened 2025-10-21
Add GEMM template for C = transpose(A) * B

harsh-nod:
This PR adds a GEMM template for C = transpose(A) * B.  Here both A and B are read in a transposed fashion because we want the K dimension to be the contiguous dimension. We add a lit test and end to end test.


harsh-nod:
Not required.

PR #373, opened 2025-10-21
Combine WaveExprAttr and WaveExpressionAttr 

tgymnich:
- **rename WaveExprAttr to WaveExpressionListAttr**
- **replace WaveExpressionAttr with WaveExpressionListAttr**


martin-luecke (approved):
LGTM, but please update `ExpressionListAttr` to `WaveExpressListAttr` or `WaveExprListAttr` everywhere for consistency with the other attributes we define.

PR #377, opened 2025-10-22
Rewrite more reads and writes with gather-to-lds operations

ashay:
This patch makes three main changes.  First, it moves the
gather-to-shared passes to run after the minimize-global-loads pass.
Doing so uncovers new opportunities for rewriting read/write operations.
Second, this patch fixes a bug in the handling of dynamic values map
where we were incorrectly using the same map for both source and
destination values.  Third, this patch fixes a bug in the rewriting of
read and write index values, where instead of the `symbolic_shape`, we
should have used `indexing_dims`.

The net effect of these changes is that Wave discovers more
opportunities for optimizations in several attention kernels.

### DIFF
--- a/lit_tests/kernel/wave/attention/attention.py
+++ b/lit_tests/kernel/wave/attention/attention.py
@@ -542,9 +542,7 @@ def test_attention_bshd_gather_to_shared():
 
     # CHECK-LABEL:       func.func @base_attention
     # CHECK:                    {{.*}} = scf.for
-    # CHECK-COUNT-2:            amdgpu.gather_to_lds
-    # CHECK:                    vector.load
-    # CHECK:                    vector.store
-    # CHECK:                    vector.load
-    # CHECK:                    vector.store
+    # CHECK-NOT:                vector.store
+    # CHECK-COUNT-4:            amdgpu.gather_to_lds
+    # CHECK-NOT:                vector.store
     # CHECK-DAG:                scf.yield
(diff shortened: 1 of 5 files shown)


lit_tests/kernel/wave/attention/attention.py
```diff
@@ -542,9 +542,5 @@ def test_attention_bshd_gather_to_shared():
 
     # CHECK-LABEL:       func.func @base_attention
     # CHECK:                    {{.*}} = scf.for
-    # CHECK-COUNT-2:            amdgpu.gather_to_lds
-    # CHECK:                    vector.load
-    # CHECK:                    vector.store
-    # CHECK:                    vector.load
-    # CHECK:                    vector.store
+    # CHECK-COUNT-4:            amdgpu.gather_to_lds
```
harsh-nod: very nice! I assume a CHECK-NOT: vector.store would also work here?

lit_tests/kernel/wave/gather_to_shared.py
```diff
@@ -355,10 +355,13 @@ def repeat(acc: tkl.Register[M, N, tkl.f32]) -> tkl.Register[M, N, tkl.f32]:
 
     # CHECK-LABEL:    test_gather_to_shared_not_minimize_shared_allocs
     # CHECK:          func.func @scaled_gemm
-    # CHECK:            memref.alloc() : memref<1024xi8, #gpu.address_space<workgroup>>
-    # CHECK:            memref.view {{.*}} : memref<1024xi8, #gpu.address_space<workgroup>> to memref<32x8xi8, #gpu.address_space<workgroup>>
-    # CHECK:            memref.alloc() : memref<32x128xi8, #gpu.address_space<workgroup>>
-    # CHECK:            memref.alloc() : memref<1024xi8, #gpu.address_space<workgroup>>
-    # CHECK:            memref.view {{.*}} : memref<1024xi8, #gpu.address_space<workgroup>> to memref<32x8xi8, #gpu.address_space<workgroup>>
-    # CHECK:            memref.alloc() : memref<32x128xi8, #gpu.address_space<workgroup>>
```
ashay: Re-add these back.

ashay: Done

PR #386, opened 2025-10-28
Move wave index attribute representation to DictArrayAttr

martin-luecke:
The MMA index sequence represents different indexing expressions for the different operands. This is modeled using sympy.Piecewise expressions, to keep to a single indexing expression. This design decision prohibits conversion of the full expressions to Affine. 
For propagation in wave the Piecewise expression is specialized here: https://github.com/iree-org/wave/blob/07e5c25fc5a94a152418b97609f21f348ca5e6e5/wave_lang/kernel/wave/analysis/index_sequence_analysis.py#L414-418
To enable representation of this in water we use similar specialization logic and move representation of index sequence from a single dictionary to an array of dictionaries.
Currently, MMA ops are the only operation with more than a single entry in this array.

### DIFF
--- a/water/lib/Dialect/Wave/IR/WaveInterfaces.cpp
+++ b/water/lib/Dialect/Wave/IR/WaveInterfaces.cpp
@@ -30,7 +30,8 @@ LogicalResult wave::verifyWaveIndexMappings(Operation *op) {
   auto arr = dyn_cast<ArrayAttr>(attribute);
   if (!arr)
     return op->emitError("'index' attribute must be an array of dictionaries");
-  SmallVector<DictionaryAttr, 3> dicts;
+  SmallVector<DictionaryAttr> dicts;
+  dicts.reserve(arr.size());
   for (Attribute nestedAttr : arr) {
     auto dict = dyn_cast<DictionaryAttr>(nestedAttr);
     if (!dict)


tgymnich (approved):

water/lib/Dialect/Wave/IR/WaveInterfaces.cpp
```diff
@@ -27,46 +27,56 @@ LogicalResult wave::verifyWaveIndexMappings(Operation *op) {
...
-    AffineMap stepMap = mapping.getStep();
-    AffineMap strideMap = mapping.getStride();
-    if (failed(checkNoDims(startMap, "start")) ||
-        failed(checkNoDims(stepMap, "step")) ||
-        failed(checkNoDims(strideMap, "stride")))
-      return failure();
+  auto arr = dyn_cast<ArrayAttr>(attribute);
+  if (!arr)
+    return op->emitError("'index' attribute must be an array of dictionaries");
+  SmallVector<DictionaryAttr, 3> dicts;
```
ftynse: Nit: why 3 is hardcoded? I get that it is from the `mma` op having a max three of this, but should we care here / is there an obvious performance benefit?

water/lib/Dialect/Wave/Transforms/LowerReadWriteOps.cpp
```diff
@@ -371,7 +371,14 @@ static FailureOr<MemAccessInfo> createMemoryIndicesAndMask(
   // This logic operates after the "expansion" pass that usually unrolls all
   // tensor dimensions but one so the extent along them is 1. The dimension
   // with non-unit extent is considered to be vectorized.
-  DictionaryAttr indexDict = op.getIndexAttr();
+
+  // Read/Write ops only carry a single index expression: the first (and only)
+  // dictionary inside the array attribute.
+  ArrayAttr indexArr = op.getIndexAttr();
+  if (!indexArr || indexArr.size() != 1)
```
ftynse: Normally, I'd expect a verifier to check that, when the index array is present, there's exactly one element in it. And the second check would turn into an assert here and a diagnostic on the op, not a silent failure-to-match.

water/lib/Dialect/Wave/Transforms/LowerReadWriteOps.cpp
```diff
@@ -371,7 +371,14 @@ static FailureOr<MemAccessInfo> createMemoryIndicesAndMask(
...
+    return rewriter.notifyMatchFailure(
+        op, "'index' must be an array with exactly one dictionary");
+  DictionaryAttr indexDict = dyn_cast<DictionaryAttr>(indexArr[0]);
```
ftynse: Ditto, this should be a verifier check.

ftynse (commented):
Design question: can we have two index dictionaries in the list where the same dimension is associated with different expressions?

water/lib/Dialect/Wave/IR/WaveInterfaces.cpp
martin-luecke: I see! Sounds good 👍

martin-luecke:
> can we have two index dictionaries in the list where the same dimension is associated with different expressions?

Indeed, this is precisely the reason for this PR.
For instance an MMA with this index: 
```
{M: $WG0*BLOCK_M + Piecewise((Mod($T0, 32), ~$MMA_ACC), (Mod($GPR_NUM, 4) + 8*(Mod(floor(... Piecewise((1, ~$MMA_ACC), (32, $MMA_ACC)), K: $ARGK*BLOCK_K + 4*floor((Mod($T0, 64))/32) : 4 : 1, N: $T1*floor(BLOCK_N/2) + $WG1*BLOCK_N + Mod($T0, 32) : 1 : 1}
```

Results in this list of indexing expressions:
```
[{M: Mod($T0, 32) : 1 : 1, N: Mod($T0, 32) : 1 : 1, K: 4*floor((Mod($T0, 64))/32) : 4 : 1},
{M: Mod($T0, 32) : 1 : 1, N: Mod($T0, 32) : 1 : 1, K: 4*floor((Mod($T0, 64))/32) : 4 : 1},
{M: Mod($GPR_NUM, 4) + 8*(Mod(floor($GPR_NUM/4), 4)) + 4*floor((Mod($T0, 64))/32) : 16 : 32, N: Mod($T0, 32) : 1 : 1, K: 4*floor((Mod($T0, 64))/32) : 4 : 1}]
```

PR #388, opened 2025-10-28
Use correct offset for base element of transposed-load operation

ashay:
Prior to this patch, we indiscriminately set `THREAD_0` in the Read
operation's index to obtain the address of the top-left corner (i.e. the
base element) of the matrix to be read in a transposed manner.  This
works well when the number of waves per block is (1, 1, 1), but not when
there are more than one waves per block, since it wipes away any offset
that was the result of shifting the window of the Read operation.

This patch fixes the problem by using a uniform index for all threads
_per wave_, instead of the same index for _all_ threads.

This patch also borrows an e2e test from PR #370 (credit to Harsh for
writing that test).


ashay:
Rebased with the main branch (and fixed the lit test that was failing after IREE update).  Also removed the redundancy between the gather-to-shared pass and the hardware-transpose pass, where both passes compute a wave-uniform index.

harsh-nod (approved):
lgtm!

PR #391, opened 2025-10-29
Manual shared-memory management in GEMM

panditsa:
This PR adds `test_explicit_shared_gemm` to validate GEMM kernels with explicit shared memory allocation and manual memory transfers. 

The key issue was that when users manually allocate shared memory and write to it (e.g.,` tkw.write(global_data, allocate)`), the compiler's Dead Code Elimination (DCE) incorrectly removed these writes because read nodes had no _write_dependency linking them to the writes showing up as `[num_users = 0]` in the IR . Fixed by adding `fix_manual_allocate_dependencies()` to set write dependencies for user-created allocates after the `promote_placeholders` pass.

Another issue fixed is the NoneType error (below) when using shared memory. Fixed by adding infer_type() method, and setting the fx_node.type to a memory type.
```
    def infer_type(self, *args):
        from ..wave.utils.general_utils import infer_dim
    
>       address_space = self.memory_type.address_space
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       AttributeError: 'NoneType' object has no attribute 'address_space'
```

### DIFF
--- a/lit_tests/kernel/wave/gemm.py
+++ b/lit_tests/kernel/wave/gemm.py
@@ -2128,3 +2128,100 @@ def test_gemm_with_transpose_a_b_gfx950():
     # Ensure that the transpose loads and mma are present.
     # CHECK-COUNT-8:    amdgpu.transpose_load {{.*}} : memref<{{.*}}xf16, #gpu.address_space<workgroup>> -> vector<{{.*}}xf16>
     # CHECK-COUNT-8:    amdgpu.mfma {{.*}} * {{.*}} + {{.*}} blgp = none : vector<{{.*}}xf16>, vector<{{.*}}xf16>, vector<{{.*}}xf32>
+
+
+@run_test
+def test_explicit_shared_gemm():
+    constraints: list[tkw.Constraint] = [tkw.WorkgroupConstraint(M, BLOCK_M, 0)]
+    constraints += [tkw.WorkgroupConstraint(N, BLOCK_N, 1)]
+    constraints += [tkw.TilingConstraint(K, BLOCK_K)]
+    constraints += [tkw.WaveConstraint(M, BLOCK_M / 2)]
+    constraints += [tkw.WaveConstraint(N, BLOCK_N / 2)]
+
+    constraints += [
+        tkw.HardwareConstraint(
+            threads_per_wave=64,
+            mma_type=tkw.MMAType.F32_16x16x16_F16,
+            vector_shapes={M: 16, N: 16, K: 16},
+        )
+    ]
+
+    @tkw.wave(constraints)
+    def gemm(
+        a: tkl.Memory[M, K, ADDRESS_SPACE, tkl.f16],
+        b: tkl.Memory[N, K, ADDRESS_SPACE, tkl.f16],
+        c: tkl.Memory[M, N, ADDRESS_SPACE_0, tkl.f32],
+    ):
+        # Allocate shared memory explicitly
+        a_shared = tkw.allocate(
+            (M, K), (BLOCK_M, BLOCK_K), tkl.f16, SHARED_ADDRESS_SPACE
+        )
+        b_shared = tkw.allocate(
+            (N, K), (BLOCK_N, BLOCK_K), tkl.f16, SHARED_ADDRESS_SPACE
+        )
+
+        c_acc = tkl.Register[M, N, tkl.f32](0.0)
+
+        @tkw.iterate(K, init_args=[c_acc])
+        def repeat(acc: tkl.Register[M, N, tkl.f32]) -> tkl.Register[M, N, tkl.f32]:
+            a_global = tkw.read(a)
+            b_global = tkw.read(b)
+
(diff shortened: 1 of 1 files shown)


wave_lang/kernel/wave/promotion.py
```diff
@@ -119,6 +119,61 @@ def apply_promotion_pattern(
...
+    When users manually allocate shared memory and write to it, the subsequent
+    reads need write dependencies set to prevent DCE from removing the writes.
+    """
+    root_graph = trace.get_root_graph()
+    for subgraph in root_graph.subgraphs.values():
+        node_list = list(subgraph.nodes)
+
+        # Group writes by which allocate they target
+        writes_by_allocate = {}
+        for node in node_list:
```
harsh-nod: Can you do a trace.walk to filter only writes to shared memory address space memory?

harsh-nod (changes requested):

wave_lang/kernel/wave/promotion.py
panditsa: Done

wave_lang/kernel/wave/promotion.py
```diff
@@ -119,6 +119,49 @@ def apply_promotion_pattern(
...
+        lambda node: isinstance(get_custom(node), Write)
+        and subs_idxc(get_custom(node).memory_type.address_space)
+        == SHARED_ADDRESS_SPACE
+    )
+
+    # Group writes by which allocate they target
+    writes_by_allocate = {}
+    for node in write_nodes:
+        custom = get_custom(node)
+        writes_by_allocate.setdefault(custom.memory, []).append(node)
```
raikonenfnu: Just curious, why do we need the setdefault here?

wave_lang/kernel/wave/analysis/index_sequence_analysis.py
```diff
@@ -124,6 +124,8 @@ def has_scaled_indices(node: fx.Node):
     """
     if isinstance(get_custom(node), (Iterate, Output)):
         return False
+    if isinstance(get_custom(node), Allocate):
```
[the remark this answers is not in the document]
panditsa: Earlier we didn't have a type on allocate so this check would return False when checking for allocate nodes. Now that we set a type for allocate nodes, we end up running `resolve_scaled_indices `pass on allocate node as well (previously it used to exit early).

I originally needed this because in `set_thread_dependent_index_from_mma` (and other variants) did not propagate indices to allocate, so for allocate the size and stride would always stay 1 even through it had a scaled dimension. So some kernels in the lit test would not even compile, they failed in `resolve_scaled_indices ` assert because despite the scaled factor the size was fixed to 1. 
```
                assert (
                    dim_index.size % scale_factor == 0
                ), f"Size({dim_index.size}) needs to be divisible by scale ({scale_factor})."
```

The fix is either to ignore allocate like we had done previously (when no type was set) or to propagate indices to allocate node.

The latest version now propagates indices to allocate such that the size of scaled dimension on allocate is now updated properly.

wave_lang/kernel/wave/promotion.py
panditsa: just to create a dictionary entry with an empty list if the key doesn't exist.

wave_lang/kernel/wave/promotion.py
```diff
@@ -119,6 +119,49 @@ def apply_promotion_pattern(
...
+
+    root_graph = trace.get_root_graph()
+    for subgraph in root_graph.subgraphs.values():
+        node_list = list(subgraph.nodes)
+        # set write dependencies on reads from those allocates
+        for read_idx, node in enumerate(node_list):
```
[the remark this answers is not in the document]
panditsa: If the allocate was called outside an `tkw.iterate` and used inside the `tkw.iterate` then the user for that allocate is not the read op but rather the iterate op. Example:

```
(Pdb) p get_custom(allocate_nodes[4]).users
[Iterate(graph=<torch.fx.graph.Graph object at 0x729f5d5a6030>, fx_node=iterate, tkw_op_name='iterate', _tracing_function=<bound method define_op.<locals>.decorator.<locals>.new_function of ...>, axis=K, init_args=[register], subgraph_name='region_0', implicit_captures=[a, b, allocate, allocate_1], step=1, start=None, condition=None)]
(Pdb) p get_custom(allocate_nodes[5]).users
[Iterate(graph=<torch.fx.graph.Graph object at 0x729f5d5a6030>, fx_node=iterate, tkw_op_name='iterate', _tracing_function=<bound method define_op.<locals>.decorator.<locals>.new_function of ...>, axis=K, init_args=[register], subgraph_name='region_0', implicit_captures=[a, b, allocate, allocate_1], step=1, start=None, condition=None)]
```

Additionally, I also want to find the index of the reads that occur on an allocate after a write to the same allocate. To do that I am looking up their index in the node_list for each subgraph. So it just made sense to use the full graph.

harsh-nod (approved):
lgtm!

PR #395, opened 2025-10-29
Updated requirements such that wave-lang picks the right version of iree dependencies

xintin:
- Updated build and release process to use stable version for release and nightly iree dependencies' version for the nightly wave-lang release.
- `requirements-iree-unpinned.txt` was unused. So, repurposed it to store stable release `iree-base-*` versions in it.

### DIFF
--- a/examples/python/1_dynamic_mapping.py
+++ b/examples/python/1_dynamic_mapping.py
@@ -1,0 +1,264 @@
+"""
+Dynamic Mapping Examples
+
+Demonstrates reading and writing with dynamic index mappings and offset-based access patterns.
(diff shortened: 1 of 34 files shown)


setup.py
```diff
@@ -18,6 +17,7 @@
 
 THIS_DIR = os.path.realpath(os.path.dirname(__file__))
 REPO_ROOT = THIS_DIR
+IS_STABLE_REL = int(os.getenv("IS_STABLE_REL", "0"))
```
Hardcode84: Let's prefix all project specific env vars with `WAVE_`

PR #399, opened 2025-10-30
Unaligned shapes support with tensor load op

Hardcode84:
The main difference is that we are now computing descriptor dims based on the actual bounds (which can be tensor dimensions or something else for unaligned case) and then to compute local bounds do `bound[dim] - tile_index[dim]`.

Also, moved strides calculations completely into handlers and did a general code cleanup.


Megan0704-1 (approved):
Thank you @Hardcode84 ! LGTM overal.

PR #400, opened 2025-10-30
[Compiler][NFC] Annotate GEMM operation on schedule

raikonenfnu:
In order to do ping-pong or schedule reordering we need to track down different components (e.g local read lhs, local read rhs, global read lhs, global read lhs, mma, etc). To make this more robust towards scheduling that would require double buffering/rotating registers in the future, we refactor the search for these components by tagging it before we apply scheduling, this makes looking for components much easier.

### DIFF
--- a/wave_lang/kernel/wave/scheduling/four_stage_pipelined_scheduling.py
+++ b/wave_lang/kernel/wave/scheduling/four_stage_pipelined_scheduling.py
@@ -4,7 +4,7 @@ from .resources import Operation
 from enum import Enum, auto
 from .scheduler_utils import (
     get_scheduling_stage,
-    BaseScheduler,
+    GemmScheduler,
     is_single_mma_source,
     is_mma_node,
 )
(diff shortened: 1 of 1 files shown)


wave_lang/kernel/wave/schedule_reordering.py
```diff
@@ -341,17 +354,40 @@ def slice_scale_mma(
...
+    )
+    sorted_rhs_scale = sorted(
+        rhs_scale_nodes, key=lambda x: x.expanded_dims[reduction_dim]
+    )
+
+    mma_per_k_dim = len(sorted_mma) // reduction_expand_size
+    lhs_per_k_dim = len(sorted_lhs) // reduction_expand_size
+    rhs_per_k_dim = len(sorted_rhs) // reduction_expand_size
+    lhs_scale_per_k_dim = len(sorted_lhs_scale) // reduction_expand_size
+    rhs_scale_per_k_dim = len(sorted_rhs_scale) // reduction_expand_size
```
harsh-amd: can we refactor into a common function?

harsh-nod (approved):
lgtm!

PR #401, opened 2025-10-30
Thread election condition

Megan0704-1:
A tensor.load should be emitted by only one representive thread per wave. 
Currently every thread emits the instruction, which is not ideal and has impact on tensor waitcnt placement. 
This PR adds a conditional barrier that compares the global thread ID to fix this.


Megan0704-1:
image

PR #402, opened 2025-10-30
[water] Revamp RegionBranchOpInterface

tgymnich:
backport changes from https://github.com/llvm/llvm-project/pull/165429


tgymnich:
This will need to wait until we bump mlir...

PR #403, opened 2025-10-30
[water] Bump LLVM

tgymnich:
- bump llvm
- update amdgpu.mfma op
- backport changes from https://github.com/llvm/llvm-project/pull/165429


tgymnich:
This will need to wait until we bump mlir...

PR #406, opened 2025-10-31
wait tensorcnt support

Megan0704-1:
we used to insert a `s.wait.tensorcnt` and `lds_barrier` right after tensor loads, which block and force the waves to wait for each global to memory copy finishes before proceeding.

There are 2 cases to consider for barrier placement.
| barID  | Placements | Before | After |
|--------|--------|--------|--------|
| -1 | - Signal after the last tensor load<br>- Wait before the first shared memory load | tensor.load MK<br>tensorcnt<br>dscnt<br>signal -1<br>wait -1<br>tensor.load NK<br>tensorcnt<br>dscnt<br>signal -1<br>wait -1 | tensor.load MK<br>tensor.load NK<br>tensorcnt<br>dscnt<br>signal -1<br>wait -1 |

note that 2 consecutive signals for the same barID is undefined and invalid.

### DIFF
--- a/lit_tests/kernel/wave/codegen.py
+++ b/lit_tests/kernel/wave/codegen.py
@@ -1099,7 +1099,7 @@ def test_scheduling_ops():
 @run_test
 def test_tensor_waitcnt():
     constraints: list[tkw.Constraint] = [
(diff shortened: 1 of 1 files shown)


Megan0704-1:
image

Hardcode84 (approved):

PR #407, opened 2025-10-31
Update wmma codegen for gfx1250

Megan0704-1:


lit_tests/kernel/wave/mma.py
```diff
@@ -591,7 +591,8 @@ def mma(
     # CHECK-LABEL: test_wmma_f32_16x16x32_f16
     # CHECK:          func.func @mma
 
-    # CHECK:        llvm.call_intrinsic "llvm.amdgcn.wmma.f32.16x16x32.f16.v8f32.v16f16"({{.*}}) : (i1, vector<16xf16>, i1, vector<16xf16>, i16, vector<8xf32>, i1, i1) -> vector<8xf32>
+    # CHECK-NOT:        llvm.call_intrinsic "llvm.amdgcn.wmma.f32.16x16x32.f16.v8f32.v16f16"({{.*}}) : (i1, vector<16xf16>, i1, vector<16xf16>, i16, vector<8xf32>, i1, i1) -> vector<8xf32>
```
Hardcode84: I don't think this CHECK-NOT is very useful.

Megan0704-1: Thanks for the feedback! removed.

PR #408, opened 2025-10-31
Tensor load padding support

Hardcode84:
Tensor load op support shared memory padding with some restrictions.

The the padded allocation shape and only drop padding if it's not supported.

### DIFF
--- a/wave_lang/kernel/wave/tensor_load_to_shared.py
+++ b/wave_lang/kernel/wave/tensor_load_to_shared.py
@@ -275,7 +275,7 @@ def clear_padding(write: Write):
 
 def get_allocation_offsets(trace) -> dict[fx.Node, int]:
     allocs, _, alloc_info = get_alloc_info(trace)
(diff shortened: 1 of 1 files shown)


wave_lang/kernel/wave/codegen/read_write.py
```diff
@@ -836,6 +844,16 @@ def rshift(value, bits):
 
     # get data size val packed to i32
     data_size_val = lshift(data_size, 16)
+
+    if padding := dst_memory.padding:
+        bytewidth = element_type.bitwidth() // 8
```
Megan0704-1: I think byte width is already available [here](https://github.com/iree-org/wave/blob/b7569720ea732120e56a19089365bb3571e9f000/wave_lang/kernel/wave/codegen/read_write.py#L767)

Megan0704-1 (commented):
Thank you @Hardcode84!

PR #425, opened 2025-11-04
[water] remove wg_constraint parameter from wave_constraint

tgymnich:
Stacked PRs:
 * #422
 * #426
 * __->__#425

--- --- ---

### [water] remove wg_constraint parameter from wave_constraint

- wg_constraint is an implementation detail and never set in the wave DSL, so don't expose it on the IR level.


tgymnich:
wg_constraint is an implementation detail and never set in the wave DSL, so don't expose it on the IR level.

PR #426, opened 2025-11-04
[water] handle device constraint

tgymnich:
Stacked PRs:
 * #422
 * __->__#426

--- --- ---

### [water] handle device constraint

- make water symbol handling aware of the fixed magic names used for DeviceConstraint
- error out when encountering a device distribution symbol for now

### DIFF
--- a/water/lib/Dialect/Wave/Transforms/LowerReadWriteOps.cpp
+++ b/water/lib/Dialect/Wave/Transforms/LowerReadWriteOps.cpp
@@ -15,7 +15,9 @@
 #include "mlir/Dialect/Vector/IR/VectorOps.h"
 #include "mlir/Transforms/DialectConversion.h"
 
+#include "llvm/ADT/STLExtras.h"
 #include "llvm/Support/Debug.h"
+#include "llvm/Support/LogicalResult.h"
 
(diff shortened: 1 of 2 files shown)


water/lib/Dialect/Wave/Transforms/LowerReadWriteOps.cpp
```diff
@@ -88,6 +88,12 @@ materializeAffine(Location loc, ArrayRef<wave::WaveSymbolAttr> symbols,
       v = blockId(gpu::Dimension::y);
     else if (name == "_WG2")
       v = blockId(gpu::Dimension::z);
+    else if (name == "_DD0")
+      v = blockId(gpu::Dimension::x);
+    else if (name == "_DD1")
+      v = blockId(gpu::Dimension::y);
+    else if (name == "_DD2")
+      v = blockId(gpu::Dimension::z);
```
ftynse: Why is this using `blockId`?

tgymnich: Current wave implementation is a placeholder implementation with `arith.constant(index, 0)`.
I implemented the same placeholder in water.

PR #440, opened 2025-11-11
[water] Add debug locations

tgymnich:

### DIFF
--- a/wave_lang/kernel/_support/location.py
+++ b/wave_lang/kernel/_support/location.py
@@ -8,7 +8,7 @@ from __future__ import annotations
 import inspect
 import sys
 from dataclasses import dataclass
-from typing import List, Optional, Union, Callable
+from typing import List, Optional, TypeVar, Union, Callable
 
 from ...support.location_config import LocationCaptureConfig, LocationCaptureLevel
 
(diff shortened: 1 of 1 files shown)


lit_tests/kernel/wave/mlir_converter_debug_locations.py
```diff
@@ -0,0 +1,257 @@
...
+
+    # CHECK: return loc(#loc)
+
+    # CHECK: loc(#loc)
+    # CHECK: loc(#loc)
+
+    # CHECK: #loc1 = loc("{{.*}}mlir_converter_debug_locations.py":55:12 to :24)
+    # CHECK: #loc2 = loc("{{.*}}mlir_converter_debug_locations.py":56:12 to :24)
+    # CHECK: #loc3 = loc("{{.*}}mlir_converter_debug_locations.py":59:12 to :25)
+    # CHECK: #loc4 = loc("{{.*}}mlir_converter_debug_locations.py":62:4 to :24)
```
ftynse: FYI, I remember that column capture is not supported prior to _some_ version of Python. We may only be running water tests on a single version, but let's structure the test defensively against that. IMO, it's fine not to check for exact columns here.

wave_lang/kernel/wave/mlir_converter/water_emitter.py
```diff
@@ -562,8 +581,9 @@ def _emit_from_captured_trace(
             "Transform dialect pipelines are not implemented yet."
         )
 
-    # TODO: Forward locations properly
-    with ir.Context() as ctx, ir.Location.unknown():
+    with ir.Context() as ctx, (
+        trace.location.to_water() if trace.location else ir.Location.unknown()
```
ftynse: Can we have some sort of warning if `trace.location is None` here but location capturing is enabled?

ftynse (approved):

PR #444, opened 2025-11-13
[Dynamo] Fix fallback to eager mode in Wave inductor backend

### LINKED ISSUE #445: Dynamo Wave backend produces inaccurate output
After changes introduced in #444 , the Wave backend for `torch.compile` introduced in #396 now does call into the wave compiled kernel for `torch.matmul` as written in the accompanying test. However. there is a mismatch against the golden output generated by eager mode execution, with a 100% mismatch and significant absolute differences in the values.

For reference, the testfile is located at https://github.com/iree-org/wave/blob/main/tests/kernel/wave/dynamo_backend_test.py and the source is located at https://github.com/iree-org/wave/blob/main/wave_lang/dynamo/register_backend.py.

### To reproduce

Pull in PR #444 and after installing wave as an editable, run
```sh
 pytest -s tests/kernel/wave/dynamo_backend_test.py -vv
```

Which will produce an output as follows:
```
platform linux -- Python 3.12.12, pytest-8.4.2, pluggy-1.6.0 -- /home/vinaydev/Scratch/PyTorch/wave/.wave-venv/bin/python3.12
cachedir: .pytest_cache
rootdir: /home/vinaydev/Scratch/PyTorch/wave
configfile: setup.cfg
plugins: timeout-2.4.0, xdist-3.8.0
collected 1 item

tests/kernel/wave/dynamo_backend_test.py::test_wave_dynamo_backend FAILED

========================================================================================================================================================= FAILURES =========================================================================================================================================================
_________________________________________________________________________________________________________________________________________________ test_wave_dynamo_backend _________________________________________________________________________________________________________________________________________________

    def test_wave_dynamo_backend():
        assert "wave" in torch.compiler.list_backends()

        a = torch.randn(64, 64)
        b = torch.randn(64, 64)

        def model(x, y):
            a = torch.matmul(x, y.T)
            return a

        # Test eager execution
        with torch.no_grad():
            expected_output = model(a, b)

        # Test compiled model execution with Wave kernel
        compiled_model = torch.compile(model, backend="wave")
        with torch.no_grad():
            actual_output = compiled_model(a, b)

>       assert_close(expected_output, actual_output, check_device=False)
E       AssertionError: Tensor-likes are not close!
E
E       Mismatched elements: 4096 / 4096 (100.0%)
E       Greatest absolute difference: 44.53440856933594 at index (19, 38) (up to 1e-05 allowed)
E       Greatest relative difference: 6212.14013671875 at index (18, 53) (up to 1.3e-06 allowed)

tests/kernel/wave/dynamo_backend_test.py:27: AssertionError
===================================================================================================================================================== warnings summary =====================================================================================================================================================
tests/kernel/wave/dynamo_backend_test.py::test_wave_dynamo_backend
  /home/vinaydev/Scratch/PyTorch/wave/wave_lang/kernel/wave/utils/torch_utils.py:51: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.detach().clone() or sourceTensor.detach().clone().requires_grad_(True), rather than torch.tensor(sourceTensor).
    return to_default_device(torch.tensor(*args, **kwargs))

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================================================================================================================================================= short test summary info ==================================================================================================================================================
FAILED tests/kernel/wave/dynamo_backend_test.py::test_wave_dynamo_backend - AssertionError: Tensor-likes are not close!

Mismatched elements: 4096 / 4096 (100.0%)
Greatest absolute difference: 44.53440856933594 at index (19, 38) (up to 1e-05 allowed)
Greatest relative difference: 6212.14013671875 at index (18, 53) (up to 1.3e-06 allowed)
============================================================================================================================================== 1 failed, 1 warning in 11.93s ===============================================================================================================================================
```

vinayakdsci:
Fixes #445.

The Dynamo backend registry for the Wave backend currently only supports the `matmul` operation to be optimized during graph tracing.

However, it lists `torch.ops.aten.mm.default` as the only optimizable op, and the test for the backend uses `torch.matmul` to generate the FX graph. As `torch.matmul` does not map top `torch.ops.aten.mm.default`  but instead looks something like `<built-in method matmul of type object at ...>`, we never actually call the Wave backend and Dynamo silently falls back to eager mode execution.

This patch fixes the above issue and also fixes another subsequent issue where Wave fails to compile the kernel when `torch.matmul` is added to the match list, due to the `schedule` argument being a boolean in `WaveCompileOptions`, instead of a `SchedulingType`.

Also modifies the backend to fix numerical errors arising due to differences in how Wave expects the inputs and how they were passed in by taking a transpose of the RHS input tensor.

### DIFF
--- a/wave_lang/dynamo/register_backend.py
+++ b/wave_lang/dynamo/register_backend.py
@@ -190,12 +190,9 @@ def wave_gemm_kernel(a: torch.Tensor, b: torch.Tensor) -> torch.Tensor:
     mfma_variant = MMAType.F32_16x16x16_F16
     threads_per_wave = 64
 
+    # TODO(vinayakdsci): Add a branch for `gfx1250` once we add support for it.
     if "gfx120" in get_default_arch():
         mfma_variant = MMAType.RDNA4_WAVE32_F32_16x16x16_F16
-        threads_per_wave = 32
-
-    if "gfx125" in get_default_arch():
-        mfma_variant = MMAType.GFX1250_F32_16x16x32_F16
         threads_per_wave = 32
 
     gemm, hyperparams, dynamic_symbols = get_gemm_kernel(


wave_lang/dynamo/register_backend.py
```diff
@@ -185,8 +189,17 @@ def wave_gemm_kernel(a: torch.Tensor, b: torch.Tensor) -> torch.Tensor:
...
     datatype = torch.float16
     mfma_variant = MMAType.F32_16x16x16_F16
+    threads_per_wave = 64
+
+    if "gfx120" in get_default_arch():
+        mfma_variant = MMAType.RDNA4_WAVE32_F32_16x16x16_F16
+        threads_per_wave = 32
+
+    if "gfx125" in get_default_arch():
+        mfma_variant = MMAType.GFX1250_F32_16x16x32_F16
```
nithinsubbiah: Shouldn't `threads_per_wave` be 32 for RDNA4.5 as well?

wave_lang/dynamo/register_backend.py
```diff
@@ -185,8 +188,18 @@ def wave_gemm_kernel(a: torch.Tensor, b: torch.Tensor) -> torch.Tensor:
     dynamic_dims = False
     datatype = torch.float16
     mfma_variant = MMAType.F32_16x16x16_F16
+    threads_per_wave = 64
+
+    if "gfx120" in get_default_arch():
+        mfma_variant = MMAType.RDNA4_WAVE32_F32_16x16x16_F16
+        threads_per_wave = 32
+
+    if "gfx125" in get_default_arch():
```
nithinsubbiah: Let's remove this arch for now.

PR #453, opened 2025-11-17
Schedule ops: proxy tracking after pipelining

panditsa:
- Update tracked proxies with all mapped nodes (PROLOGUE, KERNEL, EPILOGUE) after pipelining, eliminating the need for manual re-querying. 
- Adds `pipeline_loop.KERNEL/PROLOGUE/EPILOGUE` properties and `filter_nodes()` API to filter proxies by pipeline stage. Uses node-name based mapping to preserve partition distinctions (e.g., `global_load_a` vs `shared_load_a`).
- New API: `insert_at_start` and `insert_at_end` to add barriers at the beginning of the subgraph or before returning from the subgraph.
- Separated schedule and kernel implementation in schedules folder.


harsh-nod (approved):
lgtm!

PR #471, opened 2025-11-21
[water] index sequence analysis as dataflow

ftynse:
Implement a subset of the index sequence/expression analysis within the
MLIR dataflow framework. This creates the forward and backward analyses
sharing the lattice and the interface to be implemented by operations
willing to participate in the analysis.

Only MMAs are currently defined as operations inducing index
expressions. Further hardening is needed to correctly resolve conflicts
between different kinds of index-inducing operations, e.g., writes.

Introduce extensive testing detecting index conflicts and reporting them
as diagnostics. To simplify tests, introduce an additional test pass
that allows for index expressions to be injected on construction from
attributes instead of relying on potentially evolving per-operation
propagation rules.

### DIFF
--- a/water/include/water/Dialect/Wave/IR/IndexExpr.h
+++ b/water/include/water/Dialect/Wave/IR/IndexExpr.h
@@ -20,7 +20,7 @@ namespace wave {
 // Aggregate all symbols from a range of ranges of attributes into a single
 // range of attributes without duplicates.
 template <typename RangeT>
-static void
+static inline void
 aggregateAllSymbols(RangeT &&symbolLists,
                     llvm::SmallVectorImpl<mlir::Attribute> &symbols) {
   static_assert(
--- a/water/lib/Dialect/Wave/Transforms/InferTypes.cpp
+++ b/water/lib/Dialect/Wave/Transforms/InferTypes.cpp
@@ -1089,8 +1089,8 @@ public:
         overrideInitialization(overrideInitialization) {}
 
   llvm::LogicalResult initialize(mlir::Operation *top) override {
-    if (getSolverConfig().isInterprocedural())
-      return top->emitError() << "interprocedural analysis not supported";
+    assert(!getSolverConfig().isInterprocedural() &&
+           "interprocedural analysis not supported");
 
     // Call the base class initialization in order to set up update listeners.
     // Note that this will initialize values at function/region entries to


water/include/water/Dialect/Wave/IR/WaveInterfaces.h
```diff
@@ -381,6 +389,171 @@ template <typename OpTy>
...
+
+private:
+  // The internal storage is either a dictionary attribute with one entry per
+  // symbol indexing the value or one of the top/bottom flags.
+  llvm::PointerIntPair<mlir::Attribute, 2> value;
+
+  // State flags.
+  const static unsigned kUninitializedState = 0;
+  const static unsigned kSpecificTypeState = 1;
+  const static unsigned kUndecidableState = 2;
```
tgymnich: can we make this constexpr?

water/lib/Dialect/Wave/IR/WaveInterfaces.cpp
```diff
@@ -476,4 +478,449 @@ llvm::LogicalResult wave::detail::verifyCompatibleOperandsAndResultsOpTrait(
...
+  for (mlir::NamedAttribute namedAttr : rhsValue) {
+    if (ignoredRhsSymbolNames.contains(namedAttr.getName().getValue())) {
+      continue;
+    }
+
+    auto it = result.find(namedAttr.getName());
+    if (it == result.end()) {
+      result[namedAttr.getName()] = namedAttr.getValue();
+      continue;
+    }
```
tgymnich:
```suggestion
    auto [it, inserted] = result.try_emplace(namedAttr.getName(), namedAttr.getValue());
    if (inserted)
      continue;
```

water/lib/Dialect/Wave/IR/WaveInterfaces.cpp
```diff
@@ -476,4 +478,449 @@ llvm::LogicalResult wave::detail::verifyCompatibleOperandsAndResultsOpTrait(
...
+      changeResult = mlir::ChangeResult::Change;
+      toLattice = newLattice;
+    }
+  }
+  return changeResult;
+}
+
+llvm::LogicalResult wave::detail::checkAndAppendIndexExpr(
+    mlir::Location loc, const IndexExprsLatticeStorage &expr,
+    llvm::Twine description,
```
tgymnich:
```suggestion
    const llvm::Twine &description,
```

water/lib/Dialect/Wave/IR/WaveInterfaces.cpp
ftynse: Change for marginal efficiency gain on copies, but I find `try_emplace` less readable in this context.

martin-luecke (approved):
A very big one, but lgtm

I am debating whether it makes sense to move index and type propagation-related stuff from WaveOps.cpp to its own file, because it grows >100% with this PR. Mixing all of these contexts in this file will make the barrier to entry higher. But of course, the behavior of the ops in the context of the propagations is part of their definition as well, so it makes sense to have them here.
Anyway, either way is fine with me.

PR #472, opened 2025-11-21
[water] fix debug location check lines

tgymnich:
- fix debug location line numbers that changed due to a line break inserted by the formatter


ftynse:
Please elaborate in the PR description why they were broken

tgymnich:
@ftynse Changing the assertion message caused a line break when I ran the formatter on commit :)

ftynse (approved):

PR #473, opened 2025-11-21
[water] WaveExprList replace nb::list with std::vector

tgymnich:
found another occurrence of nb::list in the arguments. This PR makes WaveExprListAttr  consistent with WaveIndexMappingAttr.


tgymnich:
fyi: https://nanobind.readthedocs.io/en/latest/exchanging.html

tldr:
Three options 
1. casters => cast python object to stl (copy)
2. bindings => create c++ obj in python  (no-copy)
3. wrappers => use python obj from c++ (no-copy)

PR #474, opened 2025-11-22
[asm] Add loop support (scf.for) with unified SSA tracking and regist…

harsh-nod:
…er reuse

This PR adds native support for MLIR structured control flow loops (scf.for) to the ASM backend, enabling efficient K-loop tiling for GEMM with MFMA accumulator chaining. It introduces unified SSA-to-VGPR tracking (ssa_to_vgpr dictionary) to replace fragmented tracking variables, simplifying data flow and improving correctness.

### DIFF
--- a/wave_lang/kernel/wave/asm/handlers.py
+++ b/wave_lang/kernel/wave/asm/handlers.py
@@ -17,6 +17,7 @@ from wave_lang.support.ir_imports import (
     arith_d,
     gpu_d,
     memref_d,
+    scf_d,
     stream_d,
     vector_d,
 )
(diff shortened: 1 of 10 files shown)


wave_lang/kernel/wave/asm/handlers.py
```diff
@@ -924,3 +989,77 @@ def _emit_store_instruction(self, operation, kernel_info, memref_ssa, indices):
...
+
+        # Extract loop bounds
+        lower_bound_ssa = str(operation.lowerBound)
+        upper_bound_ssa = str(operation.upperBound)
+        step_ssa = str(operation.step)
+
+        # Get bounds from index_env (should be constants)
+        lower_bound = kernel_info.index_env.get(lower_bound_ssa, 0)
+        upper_bound = kernel_info.index_env.get(upper_bound_ssa, 16)
+        step = kernel_info.index_env.get(step_ssa, 1)
```
ashay: I worry that the default values may cause Wave to miscompile, in addition to making it difficult to find the root cause of the problem.  It seems a better alternative is to let the compilation fail if these SSA values are not present in `index_env`.

ashay (approved):
Left a couple of small questions/suggestions.  Rest LGTM!

PR #478, opened 2025-11-25
Scheduling: Add support for global_to_shared

panditsa:
With this PR we can now filter for tagged nodes of type GatherToLDS in Wave schedule. To enable this change, we pass the tags when creating gather_to_lds nodes in the compiler pass.


lit_tests/kernel/wave/wave_schedule.py
```diff
@@ -70,7 +71,9 @@ def test_gemm_prefetch_reorder_stagger():
     mfma_variant = wave.MMAType.F32_16x16x16_F16
 
     # Get the kernel, schedule, and options from the template
-    gemm, options = get_tagged_gemm(shape, mfma_variant, compile_to_mlir=True)
+    gemm, options = get_tagged_gemm(
+        shape=shape, mfma_variant=mfma_variant, compile_to_mlir=True
```
harsh-nod: Is this just a refactor?

PR #479, opened 2025-11-25
Add water install to main `setup.py`

Hardcode84:
* Update `setup.py`
* Update water module lookup
* Update wave runtime to link nanobind statically so I don't need to bother with `libnanobind.so` in new `CMakeBuild`

### DIFF
--- a/wave_lang/kernel/wave/water_mlir/__init__.py
+++ b/wave_lang/kernel/wave/water_mlir/__init__.py
@@ -1,3 +1,9 @@
+# Copyright 2025 The Wave Authors
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
 from importlib import import_module as _import_module
 import sys
 


setup.py
```diff
@@ -37,30 +62,156 @@ def build_cmake(self, ext):
...
+                    f"LLVM already built with correct SHA {llvm_sha}, skipping rebuild"
+                )
+                return llvm_install_dir
+            else:
+                print(
+                    f"LLVM SHA mismatch: installed={installed_sha}, expected={llvm_sha}"
+                )
+
+        print(f"Removing old LLVM installation...")
+        shutil.rmtree(llvm_install_dir, ignore_errors=True)
```
ftynse: Do we need to ignore errors? It looks like we may be unable to remove something, which will then may or may not get overwritten during installation, leading to a frankenmonster installation that is a nightmare to debug

.gitignore
```diff
@@ -47,6 +47,8 @@ out
 water/.cache
 
 # Water artifacts
+wave_lang/kernel/wave/water_mlir/*
+!wave_lang/kernel/wave/water_mlir/__init__.py # __init__.py is not installed by cmake
```
ftynse: I don't see this file in the repo

setup.py
```diff
@@ -125,6 +276,26 @@ def initialize_options(self):
...
+ext_modules = [
+    CMakeExtension("wave_runtime", "wave_lang/kernel/wave/runtime"),
+]
+
+if BUILD_WATER:
+    ext_modules += [
+        CMakeExtension(
+            "water",
+            "water",
+            install_dir="wave_lang/kernel/wave/water_mlir",
```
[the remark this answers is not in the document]
Hardcode84: I think we need to, so `pip install` machinery can pick up it later

setup.py
Hardcode84: without `ignore_errors` it will fail if directory doesn't exests

setup.py
```diff
@@ -37,30 +62,156 @@ def build_cmake(self, ext):
...
+        )
+        if llvm_install_dir.exists() and vcs_revision_file.exists():
+            # Read the installed LLVM SHA
+            vcs_content = vcs_revision_file.read_text()
+            installed_sha = None
+            for line in vcs_content.split("\n"):
+                if line.strip().startswith("#define LLVM_REVISION"):
+                    # Extract SHA from: #define LLVM_REVISION "478e45fb..."
+                    installed_sha = line.split('"')[1]
+                    break
```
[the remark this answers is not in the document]
Hardcode84: We don't expect this format to change

ftynse (approved):
Awesome!

PR #484, opened 2025-11-26
[water] remove unused function

ftynse:
Less code, less maintenance.


ftynse:
This doesn't affect tests that are specific to mi250.

PR #486, opened 2025-11-26
[water] only allow 2d MMA operations

ftynse:
This is already expected implicitly by shape matchers, but not verified.


martin-luecke (approved):
LGTM

PR #495, opened 2025-11-28
[gfx1250] Enable scheduling for tensor.load.to.lds

tgymnich:
- Change memory access type for GatherToLDS and TensorLoadToLDS to
  WRITE, since we only look at LocalMemory
- Correctly handle bounds and tile_indices in during loop reconstruction
- Memorize if the original producer is a TDM Op in BasicSplitBarrierEmitter when finding nodes for
  signal placement

### DIFF
--- a/wave_lang/kernel/wave/barriers.py
+++ b/wave_lang/kernel/wave/barriers.py
@@ -141,7 +141,7 @@ class BasicSplitBarrierEmitter(BarrierEmitter):
         consumer = region.consumer
         barrier = is_barrier_between(producer, consumer, barId)
 
-        is_tdm = is_tensor_op(producer) or is_tensor_op(consumer) or region.is_tdm
+        is_tdm = is_tensor_op(producer) or region.is_tdm
 
         if barrier is None:
             with producer.graph.inserting_after(producer):


wave_lang/kernel/wave/utils/barriers_utils.py
```diff
@@ -558,14 +561,23 @@ def find_closest(sorted_regions, pos):
     results = []
     for placement in minimal_placements:
         closest_signal = find_closest(ascending_producers, placement.cons_location)
+
+        is_tdm = any(
+            [
```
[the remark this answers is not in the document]
tgymnich: fixed

lit_tests/kernel/wave/gemm.py
```diff
@@ -1381,6 +1381,136 @@ def repeat(acc: tkl.Register[M, N, tkl.f32]) -> tkl.Register[M, N, tkl.f32]:
...
+        c_reg = tkl.Register[M, N, tkl.f32](0.0)
+
+        @tkw.iterate(K, init_args=[c_reg])
+        def repeat(acc: tkl.Register[M, N, tkl.f32]) -> tkl.Register[M, N, tkl.f32]:
+            a_reg = tkw.read(a)
+            b_reg = tkw.read(b)
+            acc = tkw.mma(a_reg, b_reg, acc)
+            return acc
+
+        tkw.write(repeat, c)
```
ftynse: I'd love for us not to repeat this boilerplate all the time, can we use the one from `tempalte`?

ftynse (approved):

wave_lang/kernel/wave/scheduling/scheduler_utils.py
```diff
@@ -139,20 +140,21 @@ def get_local_writes(self, local_loads):
...
                 g
                 for g in memory_users
-                if isinstance(get_custom(g), GatherToLDS) and g.graph == custom.graph
+                if isinstance(get_custom(g), (GatherToLDS, TensorLoadToLDS))
+                and g.graph == custom.graph
             ]
-            lds_gathers.update(cur_gathers)
-        return list(lds_gathers)
+            loads.update(cur_loads)
+        return list(loads)
```
Hardcode84: I know it was in the old code, but `set` iteration order is not stable, we should probably use `dict[key, None]`

lit_tests/kernel/wave/gemm.py
tgymnich: using shared implementation now

PR #496, opened 2025-11-28
WIP analysis python plumbing

ftynse:
needs to be augmented with tests and sliced properly, but feel free to review with that in mind

### DIFF
--- a/tests/kernel/wave/infer_index_exprs.py
+++ b/tests/kernel/wave/infer_index_exprs.py
@@ -1,0 +1,118 @@
+# Copyright 2024 The IREE Authors
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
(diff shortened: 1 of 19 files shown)


ftynse:
The analysis result do match for matmuls, but sympy is unable to prove equality between:

` $WG0*BLOCK_M + Piecewise((Mod($T0, 16), ~$MMA_ACC), (4*floor((Mod($T0, 64))/16), $MMA_ACC))` and 

`Piecewise(($WG0*BLOCK_M + Mod($T0, 16), ~$MMA_ACC), ($WG0*BLOCK_M + 4*floor((Mod($T0, 64))/16), $MMA_ACC))`

even with simplification. Short of finding the way to extract the common component from the piecewise generally, we may need some tweaking in reconstruction to automate the tests.

PR #516, opened 2025-12-03
Force garbage collection between large tests

ftynse:
This should free the GPU memory and avoid the situation where tests fail because they cannot allocate anymore, at some extra runtime penalty.

### DIFF
--- a/tests/kernel/wave/e2e/conftest.py
+++ b/tests/kernel/wave/e2e/conftest.py
@@ -1,0 +1,28 @@
+# Copyright 2024-2025 The Wave Authors
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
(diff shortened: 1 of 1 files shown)


ftynse:
Hopefully this is enough and I don't need to get deeper into the runtime (the error is thrown from IREE trying to allocate something, but I'd hope IREE also deallocates it later on).

tests/kernel/wave/e2e/conftest.py
```diff
@@ -0,0 +1,28 @@
...
+This is a special file that is used by pytest to inject a fixture into all tests
+in this package.
+"""
+
+
+@pytest.fixture(scope="module", autouse=True)
+def cleanup_memory():
+    """Auto-apply fixture that forces garbage collection and releases torch
+    cache memory after each module."""
+    yield
```
tgymnich: what is the yield doing here?

tgymnich (approved):

PR #517, opened 2025-12-03
[water] make IterateOp optionally isolated from above

ftynse:
Introduce a capability for the wave.IterateOp to explicitly forward values defined outside it through block arguments rather than implicitly capturing them. Since this behavior is optional, the op is not declared isolated from above. This form enables more precise dataflow analyses, in particular index expression inference won't leak expressions using the loop iterator outside the loop.

### DIFF
--- a/wave_lang/kernel/wave/mlir_converter/water_emitter.py
+++ b/wave_lang/kernel/wave/mlir_converter/water_emitter.py
@@ -483,7 +483,7 @@ def _emit_ops_from_graph(
                             else ir.Location.current
                         )
 
-                    mlir_op = op_builder(result_types, axis, carried_values)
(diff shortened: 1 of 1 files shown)


water/lib/Dialect/Wave/IR/WaveOps.cpp
```diff
@@ -112,6 +115,89 @@ llvm::LogicalResult wave::AllocateOp::verify() {
...
+        continue;
+      captureOperands.push_back(&operand);
+      captures.insert(operand.get());
+    }
+    return mlir::WalkResult::advance();
+  });
+
+  // Capture values defined above.
+  llvm::SmallVector<mlir::Location> newCaptureLocs =
+      llvm::map_to_vector(captures.getArrayRef(),
```
```suggestion
      llvm::map_to_vector(captures,
```

PR #523, opened 2025-12-03
Water lowering pipeline

Hardcode84:
Finally connect everything so we can lower kernel code using water/upstream-only pipeline and run with `ExecutionEngine`.

### DIFF
--- a/tests/kernel/wave/e2e/test_copy.py
+++ b/tests/kernel/wave/e2e/test_copy.py
@@ -16,15 +16,23 @@ from wave_lang.kernel.wave.utils.general_utils import check_leaks
 from wave_lang.kernel.wave.utils.run_utils import set_default_run_config
 from wave_lang.kernel.wave.utils.torch_utils import device_randn, device_zeros
 
-from ..common.utils import param_bool, require_e2e
+from ..common.utils import param_bool, require_e2e, require_water_and_ee
 from ._test_util import get_test_shapes
+
+_water_enable = [False, pytest.param(True, marks=require_water_and_ee)]
 
 
 @require_e2e
 @pytest.mark.parametrize("shape", get_test_shapes("test_copy"))
 @param_bool("use_buffer_ops", "buf_ops")
+@param_bool("use_water_pipeline", "water", values=_water_enable)
 @check_leaks
-def test_copy(shape, use_buffer_ops, run_bench):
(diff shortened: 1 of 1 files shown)


wave_lang/kernel/wave/compile.py
```diff
@@ -261,6 +264,67 @@ def __call__(self, *args, **kwargs):
...
+    def invoke(self, *args) -> None:
+        """
+        Invokes the wave kernel with the given arguments using the ExecutionEngine.
+        """
+        # Get the current stream
+        stream_ptr = torch.cuda.current_stream().cuda_stream
+
+        # Call the JIT-compiled host wrapper function
+        # Signature: void func(void* stream, PyObject* arg0, PyObject* arg1, ...)
+        self._cfunc(stream_ptr, *(py_object(arg) for arg in args))
```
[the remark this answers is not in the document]
Hardcode84: this is a hot path, so no

water/tools/water-opt/water-opt.cpp
```diff
@@ -45,14 +52,45 @@ int main(int argc, char **argv) {
...
+      mlir::affine::AffineDialect,
+      mlir::amdgpu::AMDGPUDialect,
+      mlir::arith::ArithDialect,
+      mlir::cf::ControlFlowDialect,
+      mlir::func::FuncDialect,
+      mlir::gpu::GPUDialect,
+      mlir::memref::MemRefDialect,
+      mlir::scf::SCFDialect,
+      mlir::vector::VectorDialect,
+      wave::WaveDialect
```
ftynse: Nit: please keep these sorted alphabetically

wave_lang/kernel/wave/water.py
```diff
@@ -173,26 +174,72 @@ def replace_ops_and_collect_subspans(op: Operation) -> WalkResult:
     return local_module.get_asm(binary=False, print_generic_op_form=True)
 
 
-def find_binary(name: str) -> str:
+def find_binary(name: str) -> Optional[str]:
```
ftynse:
```suggestion
def find_binary(name: str) -> str | None:
```

`Optional` is deprecated

ftynse: indeed, I'm just merging that into the same bucket as `List` and `Dict`

PR #529, opened 2025-12-04
NFC: split e2e test into multiple files

ftynse:
Not having a 4.5k LoC test file is a trivial engineering practice. This makes it easier to find relevant tests, isolate failures, disable tests when not needed, etc.

This will further allow us to clean up memory usage per module, which is cheaper than per function.


tests/kernel/wave/e2e/__init__.py
```diff
@@ -0,0 +1,5 @@
+# Copyright 2024 The IREE Authors
```
Hardcode84: Can you update all license headers too

PR #530, opened 2025-12-04
[water] Add lowering for wave.iterate

tyb0807:
This adds lowering pattern to generate scf.for from wave.iterate.

### DIFF
--- a/water/lib/Dialect/Wave/Transforms/LoweringPatterns.cpp
+++ b/water/lib/Dialect/Wave/Transforms/LoweringPatterns.cpp
@@ -411,30 +411,33 @@ public:
     // Find tiling constraint for this dimension to get tile_size.
     std::optional<int64_t> tileSize;
     for (Attribute constraintAttr : constraints) {
-      if (auto tilingConstraint =
-              dyn_cast<wave::TilingConstraintAttr>(constraintAttr)) {
-        wave::WaveSymbolAttr constraintDim = tilingConstraint.getDim();
-        if (constraintDim.getName() == symbolName) {
-          wave::WaveExprListAttr tileSizeAttr = tilingConstraint.getTileSize();
-          AffineMap tileSizeMap = tileSizeAttr.getMap();
-          ArrayRef<Attribute> tileSizeSymbols = tileSizeAttr.getSymbols();
+      auto tilingConstraint =
+          dyn_cast<wave::TilingConstraintAttr>(constraintAttr);
+      if (!tilingConstraint)
+        continue;
 
-          // Evaluate the tile size using hyperparameters.
-          std::optional<SmallVector<int64_t>> evaluatedTileSize =
-              wave::evaluateMapWithHyperparams(tileSizeMap, tileSizeSymbols,
-                                               hyperparams);
-          if (!evaluatedTileSize) {
-            return rewriter.notifyMatchFailure(
-                op, "failed to evaluate tile size from tiling constraint");
-          }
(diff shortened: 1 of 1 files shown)


water/lib/Dialect/Wave/Transforms/LoweringPatterns.cpp
```diff
@@ -416,3 +418,146 @@ void wave::populateWaveMmaLoweringPatterns(WaveTypeConverter &typeConverter,
...
+    }
+
+    // Clone all operations except the terminator.
+    for (Operation &bodyOp : waveBody.without_terminator()) {
+      rewriter.clone(bodyOp, mapping);
+    }
+
+    // Convert wave.yield to scf.yield.
+    auto yieldOp = cast<wave::YieldOp>(waveBody.getTerminator());
+    SmallVector<Value> yieldValues;
```
tgymnich: please reserve some capacity

water/lib/Dialect/Wave/Transforms/LoweringPatterns.cpp
```diff
@@ -416,3 +418,146 @@ void wave::populateWaveMmaLoweringPatterns(WaveTypeConverter &typeConverter,
...
+
+    // Set up insertion point inside the loop body.
+    rewriter.setInsertionPointToStart(&scfBody);
+
+    // Create mapping from old block arguments to new ones.
+    IRMapping mapping;
+
+    // Map iter_args.
+    // Note: wave.iterate doesn't expose the induction variable, so we skip it.
+    for (auto [oldArg, newArg] : llvm::zip(
```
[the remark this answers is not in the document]
tgymnich: Let's assert the assumption that both ranges should have the same length

water/lib/Dialect/Wave/Transforms/LoweringPatterns.cpp
```diff
@@ -416,3 +419,147 @@ void wave::populateWaveMmaLoweringPatterns(WaveTypeConverter &typeConverter,
...
+    }
+    int64_t dimSize = resolvedDims->front();
+
+    // Find tiling constraint for this dimension to get tile_size.
+    std::optional<int64_t> tileSize;
+    for (Attribute constraintAttr : constraints) {
+      if (auto tilingConstraint =
+              dyn_cast<wave::TilingConstraintAttr>(constraintAttr)) {
+        wave::WaveSymbolAttr constraintDim = tilingConstraint.getDim();
+        if (constraintDim.getName() == symbolName) {
```
ftynse: Nit: invert the condition and use `continue` to decrease indentation.

water/lib/Dialect/Wave/Transforms/LoweringPatterns.cpp
```diff
@@ -416,3 +419,147 @@ void wave::populateWaveMmaLoweringPatterns(WaveTypeConverter &typeConverter,
...
+    }
+
+    // TODO(tyb): we reject non-exact division for now, which should require
+    // peeling or padding to be correct.
+    if (*tileSize == 0) {
+      return rewriter.notifyMatchFailure(op, "tile size cannot be zero");
+    }
+    if (dimSize % *tileSize != 0) {
+      return op.emitOpError("non-exact division not supported to prevent "
+                            "potential out-of-bounds access");
```
ftynse: Not for this patch, but can we think of a mechanism to make these better visible to the caller from python than an obscure message that only shows up in C++ debug build?

PR #544, opened 2025-12-08
Add StreamK GEMM

saladpalad:
Key changes:
- Add support for llvm ld/st attributes i.e. volatile and nontemporal
- Add tests for StreamK GEMM Kernel

### DIFF
--- a/wave_lang/kernel/wave/codegen/handlers.py
+++ b/wave_lang/kernel/wave/codegen/handlers.py
@@ -384,7 +384,7 @@ def emit_wmma(
     source_a, source_b = values
     if intr == MMAType.GFX1250_F32_16x16x32_F16:
         v8f32 = VectorType.get((8,), F32Type.get())
-        res = rocdl_d.wmma_f32_16x16x32_f16(v8f32, [source_a, source_b, acc])
+        res = rocdl_d.wmma_f32_16x16x32_f16(v8f32, source_a, source_b, acc)
         return res.result
 
     return amdgpu_d.wmma(m, n, k, source_a, source_b, acc)


wave_lang/kernel/wave/templates/gemm.py
```diff
@@ -339,3 +339,372 @@ def k_loop(acc: tkl.Register[M, N, tkl.f32]) -> tkl.Register[M, N, tkl.f32]:
...
+    import sympy
+    from wave_lang.kernel._support.indexing import sym
+    from wave_lang.kernel._support.dtype import f16, f32, i32
+    from wave_lang.kernel.lang.wave_types import Memory, Register
+
+    if not block_shape:
+        block_shape = (128, 256, 64)
+
+    m, n, k = shape
+    # note: block_m must equal 128, since it matches with number threads along the cta dim i.e. (128, 2, 1)
```
harsh-nod: Why does block_m need to be 128? Why can it not be higher or lower? Why can 1 thread not be responsible for more rows or less?

wave_lang/kernel/wave/templates/gemm.py
```diff
@@ -339,3 +339,372 @@ def k_loop(acc: tkl.Register[M, N, tkl.f32]) -> tkl.Register[M, N, tkl.f32]:
...
+                loop_start = cta_id + tkw.scalar(1, i32)
+                output_tile_iter_end = output_tile_iter_start + iters_per_output_tile
+
+                tkw.set_symbol(NEW_CTA_K_END, cta_k_end)
+                new_cta_k_end = tkw.scalar(NEW_CTA_K_END, i32)
+
+                tid = tkw.scalar(THREAD_0, i32)
+                tkw.set_symbol(THREAD_ID, tid)
+
+                # TODO: unnecessary read/write, right now it reads/writes to partial buffer
```
harsh-nod: lets spend some time together going over the specifics of this. we can invite @willghatch to the call as well.

wave_lang/kernel/wave/codegen/read_write.py
```diff
@@ -747,20 +771,42 @@ def handle_write(emitter: WaveEmitter, node: fx.Node):
...
-        start_indices,
-        start_indices_wg,
-        start_indices_th,
-        elements_per_thread,
-        get_custom(memory),
-        mask,
-        node_index=index,
-    )
+
+    if volatile:
```
willghatch: I'm curious why this logic is here vs inside `_create_vec_read_write`.  Same for the read side.  Eg. it already has stuff computing index and stride stuff.  Part of this question is because I see that `_create_vec_read_write` is very complicated, and I haven't unpacked what all of that is.  But this seems like it's doing similar things just in a different place.  (This is coming from a place of ignorance, where I want to understand the situation better more than I care about any particular decision or defense for it.)

wave_lang/kernel/wave/templates/gemm.py
```diff
@@ -339,3 +339,372 @@ def k_loop(acc: tkl.Register[M, N, tkl.f32]) -> tkl.Register[M, N, tkl.f32]:
...
+
+    if num_ctas is None:
+        num_ctas = 304  # TODO: Should be decided by a heuristic
+
+    iters_per_tile = (k + block_k - 1) // block_k
+    streamk_tiles = total_tiles
+    total_data_parallel_tiles = total_tiles - streamk_tiles
+    total_streamk_iters = streamk_tiles * iters_per_tile
+    streamk_iters_pcu = total_streamk_iters // num_ctas
+    streamk_extra_iters = total_streamk_iters % num_ctas
```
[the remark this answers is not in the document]
willghatch: Let me chime in to say one easy and useful start would be to link some reference, eg. if this design follows any other particular implementation closely, or the most accessible description of the streamk algorithm.

wave_lang/kernel/wave/constraints.py
willghatch: Nit: there are a few places like this with unrelated comment churn that would be good to remove.

wave_lang/kernel/wave/codegen/read_write.py
saladpalad: Perhaps I could refactor the logic into another function `_create_llvm_read_write` that might make it more readable

wave_lang/kernel/wave/templates/gemm.py
```diff
@@ -339,3 +339,372 @@ def k_loop(acc: tkl.Register[M, N, tkl.f32]) -> tkl.Register[M, N, tkl.f32]:
...
+    num_tiles_m = (m + block_m - 1) // block_m
+    num_tiles_n = (n + block_n - 1) // block_n
+    total_tiles = num_tiles_m * num_tiles_n
+
+    if num_ctas is None:
+        num_ctas = 304  # TODO: Should be decided by a heuristic
+
+    iters_per_tile = (k + block_k - 1) // block_k
+    streamk_tiles = total_tiles
+    total_data_parallel_tiles = total_tiles - streamk_tiles
```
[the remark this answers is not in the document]
saladpalad: TODO: remove this variable, but keep in hybrid

saladpalad:
> As discussed, this will need some changes going forward that we can make incrementally.  Let's land it before you leave.

I don't have merge perms so if someone can merge that would be great! :P But everything is good no further changes need to be made from my end.

PR #546, opened 2025-12-08
[water] Add lowerable ops as illegal

tyb0807:
Some wave operations had lowering patterns but weren't marked illegal. If patterns failed, conversion could succeed with operations left unconverted.

This commit marks them as illegal, so pattern failures cause complete
conversion failure (preventing partial conversions).


ftynse:
Why do we need this / why this wasn't needed before?

ftynse (approved):

PR #547, opened 2025-12-09
[water] add python bindings for WaveIndexMappingAttr

ftynse:
We will need to construct and inspect those on the interface with pywave.


ftynse:
Tests passed on the previous commit, the update commit is a mere comment typo fix, no need to re-run.

PR #549, opened 2025-12-09
[water] set index attribute on iterate op

ftynse:
This will be needed to communicate back to pywave.


tyb0807:
Because pywave requires all ops to have index attribute?

PR #550, opened 2025-12-09
[water] register MLIR passes in python bindings

ftynse:
This will enable us to invoke passes programmatically via the pass manager or
transform dialect, without going through water-opt binary.


tyb0807 (approved):
Nice, I was using `water-opt` the whole time....

PR #555, opened 2025-12-10
[water] overhaul index inference for IterateOp

ftynse:
The initial logic for index expression inference around IterateOps, derived
from pywave, was flawed: blindly propagating index expressions across control
flow edges of the IterateOp may result in iterator symbols escaping the
corresponding region, e.g., `_Iter_K` could be outside the loop iterating on
`K` because this loop accesses a register indexable by `K` defined above the
loop. This was easily caught by the verifiers, but has been ignored for a long
time by pywave, demonstrating the utility of strict verifiers.

Update the logic of index expression propagation around IterateOps in three
ways:
1. IterateOps are now forcibly made isolated-from-above before index expression
   propagation, requiring to explicitly remap any value defined outside the loop
   and used inside it to a block argument. After the propagation, the
   implicitly capturing form is restored and the index expression is associated
   with the loop operation instead (this may be revised in the future).
2. This in turn makes it possible to assign different index expressions to
   block arguments and captured values. Propagation of index expression
   backwards from block arguments to captured values intentionally removes the
   expression component using the loop induction variable, removing the
   problematic behavior.
3. The index expression lattice is updated to be a sum of three kinds of
   components: the thread-independent expression, the thread-dependent
   expression and one induction variable-dependent expression per loop
   surrounding the definition of the value for which the index expression is
   computed. This sum can be seen as a tuple lattice where each element is
   itself a lattice that can take explicit top and bottom values as well as a
   concrete expression with identical concrete expressions JOINing into the
   same expression and non-identical concrete expression JOINing into top to
   indicate a conflict.

The latter allows this analysis to correctly handle nested loops.

Note that iterator symbols can still leak to operations after the loop. Fixing
this will require diverging from the control flow-based handling across all
analysis and will be implemented separately. This issue existed before so
it's okay not to fix it here for complexity reduction reasons.

A minor required change updates the WaveIndexMappingAttr printer and parser to
support missing start/step/stride components that may be needed by debug
logging during index expression propagation.

### DIFF
--- a/water/lib/Dialect/Wave/IR/WaveInterfaces.cpp
+++ b/water/lib/Dialect/Wave/IR/WaveInterfaces.cpp
@@ -693,7 +693,7 @@ static bool isIndexExprMapFunctionOf(AffineMap map,
 //   - thread dependent component (thread and GPR indices)
 //   - one component per iter dimension
 //
-// Two maps can be joined if, for all pairwise components:
+// Two start maps can be joined if, for all pairwise components:
 //
 //   - the components are equal;
 //   - the component is absent from one of the maps.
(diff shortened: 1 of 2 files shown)


water/lib/Dialect/Wave/IR/WaveAttrs.cpp
```diff
@@ -43,8 +44,16 @@ using namespace wave;
 /// Helper function to parse an affine wave expression with the wave
 /// symbol names passed in `names`.
 static ParseResult parseExprWithNames(ArrayRef<StringRef> names,
-                                      AffineExpr &outExpr, AsmParser &parser) {
+                                      AffineExpr &outExpr, AsmParser &parser,
+                                      bool allowNull = false) {
   MLIRContext *context = parser.getContext();
+  if (allowNull && succeeded(parser.parseOptionalLess()) &&
+      succeeded(parser.parseKeyword("NULL")) &&
```
tgymnich: NULL should probably be a constant.

tgymnich (approved):
LGTM

PR #556, opened 2025-12-10
AllocToAlloca pass

Hardcode84:
We are currently generating `memref.alloc` for LDS allocations, IREE lowering can handle it properly but upstream lowering does not and tries to generate `malloc` calls for them. Convert to `mamref.alloca`.


ftynse:
Well, how about we use transform instead of the pass manager? :)

PR #565, opened 2025-12-12
Don't silently fail when loop scheduling is not possible

tgymnich:
- don't silently fail when loop scheduling is not possible
- lower GLOBAL_TO_SHARED_DELAY in order to reduce the number of buffer created for multibuffering


harsh-nod (approved):
lgtm!

PR #569, opened 2025-12-14
[wave2water] Add water pipeline in Python

### LINKED ISSUE #593: Make water transformations available in Python
Figure out how to apply water transformations from Python.

tyb0807:
This can be used to apply water transformations/lowering in e2e execution.

Fixes #593

### DIFF
--- a/tests/kernel/wave/test_water.py
+++ b/tests/kernel/wave/test_water.py
@@ -10,7 +10,7 @@ import sys
 from pathlib import Path
 from unittest.mock import patch, MagicMock
 from wave_lang.kernel.wave.water import (
-    apply_water_lowering_passes,
+    apply_water_middle_end_passes,
     find_binary,
     is_water_available,
     is_water_binary_available,
@@ -32,15 +32,15 @@ class TestWaterLowering:
         result = is_water_available()
         assert isinstance(result, bool)
 
-    def test_apply_water_lowering_passes_unavailable(self):
-        """Test apply_water_lowering_passes when water-opt is not available."""
+    def test_apply_water_middle_end_passes_unavailable(self):
+        """Test apply_water_middle_end_passes when water-opt is not available."""
         # Mock find_binary to return None, simulating water-opt not being found
         with patch("wave_lang.kernel.wave.water.find_binary", return_value=None):
             with pytest.raises(RuntimeError, match="water-opt binary not found"):
-                apply_water_lowering_passes("module {}")
+                apply_water_middle_end_passes("module {}")
 
-    def test_apply_water_lowering_passes_success(self):
-        """Test apply_water_lowering_passes with mocked subprocess."""
+    def test_apply_water_middle_end_passes_success(self):
+        """Test apply_water_middle_end_passes with mocked subprocess."""
         mlir_input = (
             "module attributes {wave.normal_form = #wave.normal_form<full_types>} {}"
         )
@@ -52,7 +52,7 @@ class TestWaterLowering:
             with patch("subprocess.check_output") as mock_subprocess:
                 mock_subprocess.return_value = expected_output
 
-                result = apply_water_lowering_passes(mlir_input)
+                result = apply_water_middle_end_passes(mlir_input)
 
                 # Verify basic operation
                 assert result is not None
@@ -67,8 +67,8 @@ class TestWaterLowering:
                 assert kwargs["input"] == mlir_input
                 assert kwargs["text"] is True
 
-    def test_apply_water_lowering_passes_subprocess_error(self):
-        """Test apply_water_lowering_passes when subprocess fails."""
+    def test_apply_water_middle_end_passes_subprocess_error(self):
+        """Test apply_water_middle_end_passes when subprocess fails."""
         with patch("wave_lang.kernel.wave.water.get_water_opt") as mock_get_water_opt:
             mock_get_water_opt.return_value = "/mock/water-opt"
 
@@ -80,7 +80,7 @@ class TestWaterLowering:
                 with pytest.raises(
                     RuntimeError, match="water-opt subprocess failed with return code 1"
                 ):
-                    apply_water_lowering_passes("module {}")
+                    apply_water_middle_end_passes("module {}")
 
 
 @pytest.mark.skipif(
@@ -102,7 +102,7 @@ class TestWaterLoweringIntegration:
         }
         """
 
-        result = apply_water_lowering_passes(wave_mlir)
+        result = apply_water_middle_end_passes(wave_mlir)
         assert result is not None
         assert "module" in result
         assert "func.func @test_kernel" in result
@@ -114,5 +114,5 @@ class TestWaterLoweringIntegration:
         # Wave attributes should be cleaned
         assert "wave.normal_form" not in result
 
-        # Should only have `return` remained
+        # Only `return` should remain
         assert "return" in result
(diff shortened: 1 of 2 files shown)


wave_lang/kernel/wave/water.py
```diff
@@ -446,3 +446,271 @@ def add_opt(pipeline):
...
     with module.context:
         return Module.parse(result)
+
+
+def _find_water_python_path():
+    """Find the Water Python bindings path dynamically."""
+    import sys
+    from pathlib import Path
+
+    debug_info = []  # Collect debug info for troubleshooting
```
ftynse: This comment is useless. Slop?

wave_lang/kernel/wave/water.py
```diff
@@ -446,3 +446,271 @@ def add_opt(pipeline):
...
+
+    # Try to find Water build directory using the same logic as get_water_opt
+    try:
+        water_opt_path = Path(get_water_opt())
+        debug_info.append(f"Found water-opt at: {water_opt_path}")
+
+        # water-opt is typically in build/bin/water-opt, so python packages are in build/python_packages
```
ftynse: Nit: it may be typical for you using it from the build tree, but it is not the general use case, which would be wheels installation

wave_lang/kernel/wave/water.py
```diff
@@ -446,3 +446,271 @@ def add_opt(pipeline):
...
+            # Check if this directory contains the Python package files
+            debug_info.append(
+                f"build_dir is water_mlir itself, checking for package files"
+            )
+            init_file = build_dir / "__init__.py"
+            ir_module = build_dir / "ir.py"
+            dialects_dir = build_dir / "dialects"
+            debug_info.append(f"  __init__.py exists: {init_file.exists()}")
+            debug_info.append(f"  ir.py exists: {ir_module.exists()}")
+            debug_info.append(f"  dialects dir exists: {dialects_dir.exists()}")
```
ftynse: Why this check wasn't needed before?

wave_lang/kernel/wave/water.py
```diff
@@ -446,3 +446,271 @@ def add_opt(pipeline):
...
+            wave_dialect.register_dialect(ctx)
+
+            # Register Wave passes
+            wave_dialect.register_passes()
+
+            # Parse the input MLIR
+            module = water_ir.Module.parse(mlir_text, ctx)
+
+            # Apply the pass pipeline
+            pipeline_str = "builtin.module(water-wave-detect-normal-forms,water-wave-propagate-elements-per-thread,lower-wave-to-mlir,canonicalize,cse)"
```
ftynse: There's some logic to stich these together in a nicer way, in `make_linear_pass_pipeline`, that makes this look nicer.

debug_water_import.py
ftynse: Did you intend to commit this?

wave_lang/kernel/wave/water.py
```diff
@@ -446,3 +450,65 @@ def add_opt(pipeline):
...
+            [
+                binary,
+                make_linear_pass_pipeline(pipeline),
+            ],
+            input=mlir_text,
+            text=True,
+        )
+
+        # Clean Wave-specific attributes that cause issues in later pipelines
+        # TODO(tyb): do this in the very last pass of water pipeline instead
```
ftynse: Open an issue for this so we can track it, I really dislike having random regular expressions mutating the IR. This is difficult to maintain and will be easily forgotten here.

Even better if you just vibe code the pass that removes the normal form tag, it shouldn't be more than a couple dozen lines scattered across files.

wave_lang/kernel/wave/water.py
```diff
@@ -33,6 +34,9 @@
     stream_d,
 )
 
+# Water Python bindings are conditionally available
+# We'll try to import them dynamically when needed
```
ftynse: I don't understand what "conditionally available" means and the general purpose of this comment.

wave_lang/kernel/wave/water.py
```diff
@@ -185,8 +189,8 @@ def find_binary(name: str) -> str | None:
 
 @lru_cache
 def is_water_available() -> bool:
-    """Returns True if the water_mlir package is available."""
-    return (Path(__file__).parent / "water_mlir" / "water_mlir").exists()
+    """Returns True if the water-opt binary is available."""
```
ftynse: It looks like the check for the package was intentional from here: https://github.com/iree-org/wave/commit/c14db2cdd2633e6377b8027e3fefaca81e2fbc14#diff-e7d61d9650d9dcd96695d2cfb2df154bdc27ab31c432264848386c8808b073fbL183. Are you sure you want to check for the binary instead? Binary is expected to be distributed as part of the package

tests/kernel/wave/test_water.py
```diff
@@ -6,11 +6,108 @@
...
+    """Test suite for Water lowering functionality."""
+
+    def test_apply_water_lowering_passes_unavailable(self):
+        """Test apply_water_lowering_passes when water-opt is not available."""
+        with patch(
+            "wave_lang.kernel.wave.water.get_water_opt",
+            side_effect=RuntimeError("Not found"),
+        ):
+            with pytest.raises(RuntimeError, match="Not found"):
+                apply_water_lowering_passes("module {}")
```
ftynse: I don't quite get this test. We patch `get_water_opt` to raise a `RuntimeError` and then check that indeed raised a `RuntimeError`? Feels like we are testing the mock itself...

tests/kernel/wave/test_water.py
```diff
@@ -6,11 +6,108 @@
...
+        assert "func.func" in result
+
+        # Verify Wave operations are lowered
+        assert "wave.register" not in result
+        assert "wave.add" not in result
+
+        # Wave attributes should be cleaned
+        assert "wave.normal_form" not in result
+
+        # Should have been DCE'd
```
ftynse: I don't understand what was DCE'd here if we are checking that `return` _is present_ in the result

ftynse (commented):
Indeed. Note that FWIW I have a "make python shim work with the build version of water w/o wheel installation" on my todo list, as it takes time to re-compile and re-install every time C++ changes. I merely asked whether all the complexity that was previously in this patch was needed. The answer may well be "yes, because <...>, <...> and <...>". Ideally, I shouldn't need to ask the question because commit/PR description does a good job justifying _why_ a change is necessary.

ftynse:
On a separate note, you and @Hardcode84 are in race condition on calling your respective things "water pipeline" (why nobody thought of the word aqueduct so far is beyond me 😆 ). I asked him to call his part "water backend", I suggest you call yours "water mid-end" for disambiguation. Or figure out a terminology between yourselves and document it.

Hardcode84:
I renamed my thing into `use_water_backend`

wave_lang/kernel/wave/water.py
tyb0807: See #595

tests/kernel/wave/test_water.py
```diff
@@ -6,11 +6,113 @@
...
+        assert "func.func @test_kernel" in result
+
+        # Verify Wave operations are lowered
+        assert "wave.register" not in result
+        assert "wave.add" not in result
+
+        # Wave attributes should be cleaned
+        assert "wave.normal_form" not in result
+
+        # Should only have `return` remained
```
ftynse:
```suggestion
        # Only `return` should remain
```

wave_lang/kernel/wave/water.py
```diff
@@ -491,3 +498,66 @@ def add_transform(transform: str, entry_point: str) -> tuple[str, dict[str, Any]
 
     with module.context:
         return Module.parse(result)
+
+
+def apply_water_lowering_passes(mlir_text: str) -> str:
```
ftynse: Let's call this middle_end_lowering_passes to avoid confusion

PR #570, opened 2025-12-14
[wave2water] Fix Water tests failing on machines without ROCm

tyb0807:
These tests attempt to compile GPU kernels to binary using ROCm tools but fail with `ld.lld invocation failed` errors on systems where ROCm is not available.


tyb0807:
Added a TODO, PTAL

PR #574, opened 2025-12-15
Disable minimize_shared_allocs pass by default

adedespirlet:
This change sets the default behavior of the minimize_shared_allocs pass to False.
Currently this pass reduces peak LDS memory usage by combining distinct memref.alloc operations into a single allocation with views. However this prevents the backend from inferring aliasing information. As a result the backend conservatively inserts unnecessary wait count instructions (vmcnt(0)), which breaks software pipelining.
While this optimization is beneficial for kernels with disjoint buffer lifecycles (like extended attention) it causes significant performance regressions in standard pipelined kernels. 
Let's disable it by default and enable it explicitly only for the specific cases where it is required.


ftynse:
This is a prime example why we should do more unit-style tests and less integration tests. We should still have some integration tests, but avoid tight coupling in them: just check that key things happen.

PR #577, opened 2025-12-15
[CI] Install release rocm and pytorch instead of nightly

tgymnich:
- should improve reliability


tgymnich:
I've only so far seen nightly docker images https://hub.docker.com/r/rocm/pytorch-nightly/tags. The problem here is that they use both rocm and PyTorch versions we don't usually test for.

PR #591, opened 2025-12-17
[water] make step and stride constant

ftynse:
In practice, symbolic expressions only appear in the start part of the mapping
tuple whereas its step and stride are always constant. Remove the unnecessary
complexity of using full affine expressions there and use constants instead.


ftynse:
Actually, I see cases where we don't do substitution immediately... Punting for now

Issue #622, opened 2025-12-23
Fix `elements_per_thread` propagation to ignore memory operands


tyb0807:
Memory operands should be decoupled from register dataflow for `elements_per_thread` propagation. This allows:
- Writing N elements per thread to memory
- Reading M elements per thread from same memory location
- This represents valid resharding/redistribution, not an error

PR #623, opened 2025-12-23
[water] Fix elements_per_thread propagation to ignore memory operands

tyb0807:
Changes:
- ReadOp: Only propagate attribute to result (register), ignore memory
- WriteOp: Only validate/propagate with register operand, ignore memory

This fixes false positives where memory resharding was incorrectly
flagged as propagation errors.

Fixes https://github.com/iree-org/wave/issues/622.

### DIFF
--- a/water/lib/Dialect/Wave/IR/WaveOps.cpp
+++ b/water/lib/Dialect/Wave/IR/WaveOps.cpp
@@ -1448,13 +1448,14 @@ LogicalResult ReadOp::verify() {
                                bounds.getMapping());
 }
 
-llvm::FailureOr<mlir::ChangeResult> wave::ReadOp::propagateElementsPerThreadForward(
+llvm::FailureOr<mlir::ChangeResult>
+wave::ReadOp::propagateElementsPerThreadForward(
     llvm::ArrayRef<wave::ElementsPerThreadLatticeValue>,
     llvm::MutableArrayRef<wave::ElementsPerThreadLatticeValue> resultElements,
     llvm::raw_ostream &errs) {
-  // ReadOp only propagates elements_per_thread attribute to result (register)
-  // Memory operand is ignored for propagation - you can read any number of elements
-  // from memory regardless of how many were written
+  // ReadOp only propagates elements_per_thread attribute to result (register).
+  // Memory operand is ignored for propagation - you can read any number of
+  // elements from memory regardless of how many were written.
   std::optional<int64_t> elementsPerThread = getElementsPerThread();
   if (!elementsPerThread)
     return mlir::ChangeResult::NoChange;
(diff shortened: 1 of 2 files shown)


water/lib/Dialect/Wave/IR/WaveOps.cpp
```diff
@@ -1408,6 +1436,50 @@ LogicalResult WriteOp::verify() {
                                bounds.getMapping());
 }
 
+llvm::FailureOr<mlir::ChangeResult>
+wave::WriteOp::propagateElementsPerThreadForward(
+    llvm::ArrayRef<wave::ElementsPerThreadLatticeValue> operandElements,
+    llvm::MutableArrayRef<wave::ElementsPerThreadLatticeValue>,
+    llvm::raw_ostream &errs) {
+  // WriteOp only validates that elements_per_thread attribute matches register
+  // operand Memory operand is ignored for propagation - you can write to memory
```
ftynse: Like here, if you had a full stop, it would have been more readable.

water/test/Dialect/Wave/propagate-elements-per-thread.mlir
```diff
@@ -100,7 +100,7 @@ func.func @missing_elements_per_thread(%mem: !wave.tensor<[@M] of f16, <global>>
 module attributes {wave.normal_form = #wave.normal_form<full_types>} {
 func.func @read_write_conflict(%mem: !wave.tensor<[@M] of f16, <global>>) attributes {wave.hyperparameters = #wave.hyperparameters<{M = 128}>}  {
   %reg = wave.read %mem {elements_per_thread = 4} : (!wave.tensor<[@M] of f16, <global>>) -> !wave.tensor<[@M] of f16, <register>>
-  // expected-error @below {{failed to propagate elements per thread backward: mismatch between elements_per_thread attribute (8) and operand #0 (4)}}
+  // expected-error @below {{failed to propagate elements per thread backward: mismatch between elements_per_thread attribute (8) and register operand #0 (4)}}
```
ftynse: Nit: "register operand #0" reads like there are multiple register operands. I'd consider just keeping as is.

water/test/Dialect/Wave/propagate-elements-per-thread.mlir
```diff
@@ -162,6 +162,66 @@ module {
...
+}
+
+// -----
+
+// CHECK: #wave.normal_form<full_types,memory_only_types>
+module attributes {wave.normal_form = #wave.normal_form<full_types>} {
+// CHECK-LABEL: @write_backward_propagation
+func.func @write_backward_propagation(%mem: !wave.tensor<[@M] of f16, <shared>>) attributes {wave.hyperparameters = #wave.hyperparameters<{M = 128}>} {
+  %cst = arith.constant 0.0 : f16
+  // RegisterOp without explicit elements_per_thread - should get it from backward propagation
```
ftynse: I think we punted on having explicit `elements_per_thread` on registers

Issue #624, opened 2025-12-23
Support vector types in `wave.iterate` and `wave.yield` operations


tyb0807:
This allows `wave.iterate` to work correctly both before and after the `PropagateElementsPerThread` pass, which converts register-resident wave tensors to vector types.

PR #625, opened 2025-12-23
[water] Add support for vector types in wave.iterate and wave.yield operations

tyb0807:
This allows wave.iterate to work correctly both before and after the PropagateElementsPerThread pass, which converts register-resident wave tensors to 1D vector types. The operations now accept both WaveTensorInRegister (before conversion) and 1D vectors (after conversion).

Updated type compatibility and verification logic to handle both tensor and 1D vector type combinations appropriately.

Fixes https://github.com/iree-org/wave/issues/624.

### DIFF
--- a/water/lib/Dialect/Wave/IR/WaveOps.cpp
+++ b/water/lib/Dialect/Wave/IR/WaveOps.cpp
@@ -6,7 +6,6 @@
 
 #include "water/Dialect/Wave/IR/WaveOps.h"
 
-#include "mlir/Dialect/Vector/IR/VectorOps.h"
 #include "mlir/IR/Attributes.h"
 #include "mlir/IR/Builders.h"
 #include "mlir/IR/BuiltinTypes.h"
@@ -195,25 +194,25 @@ void wave::IterateOp::makeNonIsolated(RewriterBase &rewriter) {
 }
 
 bool wave::IterateOp::areTypesCompatible(mlir::Type lhs, mlir::Type rhs) {
-  // Handle both WaveTensorType and VectorType combinations
+  // Handle both WaveTensorType and VectorType combinations.
   auto lhsTensor = llvm::dyn_cast<wave::WaveTensorType>(lhs);
   auto rhsTensor = llvm::dyn_cast<wave::WaveTensorType>(rhs);
   auto lhsVector = llvm::dyn_cast<mlir::VectorType>(lhs);
   auto rhsVector = llvm::dyn_cast<mlir::VectorType>(rhs);
 
-  // Both are wave tensors - use existing logic
+  // Both are wave tensors - check shape and address space compatibility.
   if (lhsTensor && rhsTensor) {
     return detail::verifyTypesCompatible(lhsTensor, rhsTensor,
                                          /*includeAddressSpace=*/true)
         .succeeded();
   }
 
-  // Both are vectors - simple equality check
+  // Both are vectors - simple equality check.
   if (lhsVector && rhsVector) {
     return lhsVector == rhsVector;
   }
 
-  // Mixed types are not compatible
+  // Mixed types are not compatible.
   return false;
 }
 
@@ -280,8 +279,6 @@ LogicalResult wave::IterateOp::verify() {
     // Handle verification for both wave tensors and vectors.
     auto iterArgTensor = llvm::dyn_cast<wave::WaveTensorType>(iterArg);
     auto resultTensor = llvm::dyn_cast<wave::WaveTensorType>(result);
-    auto iterArgVector = llvm::dyn_cast<mlir::VectorType>(iterArg);
-    auto resultVector = llvm::dyn_cast<mlir::VectorType>(result);
 
     auto istr = std::to_string(i);
 
@@ -300,11 +297,11 @@ LogicalResult wave::IterateOp::verify() {
         return mlir::failure();
     }
     // Both are vectors - check exact type equality.
-    else if (iterArgVector && resultVector) {
-      if (iterArgVector != resultVector) {
-        return emitOpError() << "iter_args #" << i << " type (" << iterArgVector
-                             << ") must match result #" << i << " type ("
-                             << resultVector << ")";
+    else if (isa<VectorType>(iterArg) && isa<VectorType>(result)) {
+      if (iterArg != result) {
+        return emitOpError()
+               << "iter_args #" << i << " type (" << iterArg
+               << ") must match result #" << i << " type (" << result << ")";
       }
     }
     // Mixed types are not allowed.
@@ -337,12 +334,17 @@ llvm::LogicalResult wave::IterateOp::verifyRegions() {
                        blockIterArgTypes)) {
     auto istr = std::to_string(i);
 
+    auto iterArgTensor = llvm::dyn_cast<wave::WaveTensorType>(iterArg);
+    auto resultTensor = llvm::dyn_cast<wave::WaveTensorType>(result);
+    auto blockIterArgTensor =
+        llvm::dyn_cast<wave::WaveTensorType>(blockIterArg);
+    auto terminatorOperandTensor =
+        llvm::dyn_cast<wave::WaveTensorType>(terminatorOperand);
+
     // Verify result type vs terminator operand type.
-    if (isa<wave::WaveTensorType>(result) &&
-        isa<wave::WaveTensorType>(terminatorOperand)) {
+    if (resultTensor && terminatorOperandTensor) {
       if (llvm::failed(detail::verifyTypesCompatible(
-              llvm::cast<wave::WaveTensorType>(result),
-              llvm::cast<wave::WaveTensorType>(terminatorOperand),
+              resultTensor, terminatorOperandTensor,
               /*includeAddressSpace=*/true, getLoc(), "result #" + istr,
               "terminator operand #" + istr))) {
         return llvm::failure();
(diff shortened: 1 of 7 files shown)


water/lib/Dialect/Wave/IR/WaveOps.cpp
```diff
@@ -1129,14 +1113,104 @@ LogicalResult MmaOp::verify() {
...
-  return checkMmaTypeCompatibility(getLoc(), *getKind(),
+  return checkMmaTypeCompatibility(getLoc(), getKind(),
                                    lhsType.getElementType(),
                                    accumulatorType.getElementType());
 }
 
+/// Compute the expected elements per thread for this MMA operation.
+/// Extracts threadsPerWave from ancestor operations with hardware constraints.
+/// Returns 0 if no constraints are found.
+unsigned wave::MmaOp::computeElementsPerThread() {
```
tgymnich: Can we return a `llvm::FailureOr` here instead?

water/lib/Dialect/Wave/IR/WaveOps.cpp
```diff
@@ -1129,14 +1113,104 @@ LogicalResult MmaOp::verify() {
...
+  wave::ElementsPerThreadLatticeValue expectedAccumulator(
+      expectedElementsPerThread);
+
+  unsigned accumulatorOperandNumber =
+      getAccumulatorMutable().getOperandNumber();
+
+  // Validate that LHS and RHS operands have concrete elements_per_thread
+  // values. We don't propagate to them, but we check they've been properly
+  // initialized.
+  for (unsigned i = 0; i < 2 && i < operandElements.size();
```
```suggestion
  for (unsigned i = 0; i < min(2, operandElements.size());
```

ftynse: Also, precompute this value so it is not re-evaluated on each iteration :)

ftynse (commented):
@tgymnich , when reviewing stacked PRs (the target branch is not `main`), click on the last commit and only review that one to avoid making comments on things that should be addressed in other PRs.

ftynse:
I can adapt to the style folks use. It did accidentally click on "squash and merge" in a thus stacked PR before, polluted the other branch and had to do a bunch of force-pushing and PR re-opening to fix that.

water/lib/Dialect/Wave/IR/WaveOps.cpp
```diff
@@ -199,10 +199,26 @@ void wave::IterateOp::makeNonIsolated(mlir::RewriterBase &rewriter) {
 }
 
 bool wave::IterateOp::areTypesCompatible(mlir::Type lhs, mlir::Type rhs) {
-  return detail::verifyTypesCompatible(llvm::cast<wave::WaveTensorType>(lhs),
-                                       llvm::cast<wave::WaveTensorType>(rhs),
-                                       /*includeAddressSpace=*/true)
-      .succeeded();
+  // Handle both WaveTensorType and VectorType combinations
```
ftynse: I usually don't comment on this, but please just take the habit (or configure your code generation assistant) to use full stop at the end of a sentence in comments.

water/lib/Dialect/Wave/IR/WaveOps.cpp
```diff
@@ -251,19 +267,42 @@ mlir::LogicalResult wave::IterateOp::verify() {
...
-    auto allDims =
-        llvm::to_vector(llvm::iota_range<int>(0, iterArgTensor.getRank(),
-                                              /*Inclusive=*/false));
     auto istr = std::to_string(i);
-    if (mlir::failed(detail::verifyTypesMatchingDimensions(
-            getLoc(), "iter_args #" + istr, iterArgTensor, allDims,
-            "result #" + istr, resultTensor, allDims)))
-      return mlir::failure();
+
+    // Both are wave tensors - use existing shape verification logic
```
ftynse: "existing" doesn't make sense in the standalone comment when reading the code. It only makes sense when reading the _diff_.

water/test/Dialect/Wave/propagate-elements-per-thread.mlir
```diff
@@ -259,3 +259,33 @@ func.func @mma_uninitialized_rhs(%mem1: !wave.tensor<[@M, @K] of f16, <global>>,
...
+
+// -----
+
+// Test iterate working with vectors after PropagateElementsPerThread conversion
+module attributes {wave.normal_form = #wave.normal_form<full_types>} {
+
+// CHECK-LABEL: @iterate_with_vectors_after_ept
+func.func @iterate_with_vectors_after_ept(%mem: !wave.tensor<[@M] of f32, <global>>)
+  attributes {wave.hyperparameters = #wave.hyperparameters<{M = 128, I = 4}>,
+              wave.constraints = [#wave.hardware_constraint<threads_per_wave = 64, waves_per_block = [1, 1, 1], mma_type = #wave.mma_kind<f32_32x32x8_f16>, vector_shapes = {M = 1, N = 1, K = 8}, max_bits_per_load = 128>]} {
```
ftynse: Do we need all of these constraints?

water/test/Dialect/Wave/propagate-elements-per-thread.mlir
```diff
@@ -259,3 +259,33 @@ func.func @mma_uninitialized_rhs(%mem1: !wave.tensor<[@M, @K] of f16, <global>>,
...
+
+  // Read into register tensor - this will become a vector after PropagateElementsPerThread
+  %init = wave.read %mem {elements_per_thread = 8} : (!wave.tensor<[@M] of f32, <global>>) -> !wave.tensor<[@M] of f32, <register>>
+
+  // Iterate should work with vectors after transformation
+  // CHECK: wave.iterate @I iter_args({{.*}})
+  %result = wave.iterate @I iter_args(%init) {
+  ^bb0(%arg: !wave.tensor<[@M] of f32, <register>>):
+    // Simple operation within the loop - should also work with vectors
```
ftynse: At which point maybe this should `// CHECK` that it actually does?

water/test/Dialect/Wave/propagate-elements-per-thread.mlir
```diff
@@ -259,3 +259,33 @@ func.func @mma_uninitialized_rhs(%mem1: !wave.tensor<[@M, @K] of f16, <global>>,
...
+    %doubled = wave.add %arg, %arg : (!wave.tensor<[@M] of f32, <register>>, !wave.tensor<[@M] of f32, <register>>) -> !wave.tensor<[@M] of f32, <register>>
+    // CHECK: wave.yield {{.*}} : vector<8xf32>
+    wave.yield %doubled : !wave.tensor<[@M] of f32, <register>>
+  } : (!wave.tensor<[@M] of f32, <register>>) -> (!wave.tensor<[@M] of f32, <register>>)
+
+  // Write should also work with the vector result
```
ftynse: How do we know it does?

water/lib/Dialect/Wave/IR/WaveOps.cpp
```diff
@@ -6,6 +6,7 @@
 
 #include "water/Dialect/Wave/IR/WaveOps.h"
 
+#include "mlir/Dialect/Vector/IR/VectorOps.h"
```
ftynse: Do we really need this include?

water/lib/Dialect/Wave/IR/WaveOps.cpp
```diff
@@ -310,14 +336,67 @@ llvm::LogicalResult wave::IterateOp::verifyRegions() {
...
-            /*includeAddressSpace=*/true, getLoc(), "result #" + istr,
-            "terminator operand #" + istr))) {
-      return llvm::failure();
+
+    // Verify result type vs terminator operand type.
+    if (isa<wave::WaveTensorType>(result) &&
+        isa<wave::WaveTensorType>(terminatorOperand)) {
+      if (llvm::failed(detail::verifyTypesCompatible(
+              llvm::cast<wave::WaveTensorType>(result),
+              llvm::cast<wave::WaveTensorType>(terminatorOperand),
```
ftynse: `isa` followed by `cast` is an anti-pattern in LLVM. Just do `dyn_cast` before the if/else block similarly to what you do above.

water/lib/Dialect/Wave/Transforms/LowerReadWriteOps.cpp
```diff
@@ -125,6 +127,96 @@ materializeAffine(Location loc, ArrayRef<Attribute> symbols, AffineMap map,
...
+      // We need the tile size to compute iteration_value = induction_var *
+      // tile_size Get this from the function's tiling constraints.
+      func::FuncOp parentFunc = current->getParentOfType<func::FuncOp>();
+      if (!parentFunc) {
+        return rewriter.notifyMatchFailure(loc, "no parent function found");
+      }
+
+      auto hyperparamAttr =
+          parentFunc->getAttrOfType<wave::WaveHyperparameterAttr>(
+              wave::WaveDialect::kHyperparameterAttrName);
```
ftynse: There is absolutely no guarantee there is a parent `func.func` specifically or that hyperparameters are attached to a `func.func`. Just walk the parents until you see the attribute. Even better if the attribute is passed down this function from the pass somehow.

ftynse: I double-checked, and we already do this in this pass! Hyperparams are available from the type converter and are even used below in this file.

water/lib/Dialect/Wave/Transforms/LowerReadWriteOps.cpp
```diff
@@ -125,6 +127,96 @@ materializeAffine(Location loc, ArrayRef<Attribute> symbols, AffineMap map,
...
+            tileSize = (*evaluatedTileSize)[0];
+            break;
+          }
+        }
+      }
+
+      if (!tileSize) {
+        return rewriter.notifyMatchFailure(loc,
+                                           "could not determine tile size");
+      }
```
ftynse: I think I'd differentiate two cases:
 - there is no tiling constraint for the loop, then it is simply not tiled, its "tile size" is equal to the trip count (unless something different happens with pywave);
 - we failed to evaluate the map: this is an error; ideally it is a user-visible error and not just failure-to-lower.

water/lib/Dialect/Wave/Transforms/LowerReadWriteOps.cpp
```diff
@@ -13,7 +13,9 @@
 
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
```
ftynse: This is no longer necessary

PR #630, opened 2025-12-23
[water] Add error message for failed normal form pass postcondition

tgymnich:


ftynse (commented):
Would it make sense to put the message into the `setNormalFormPassPostcondition` itself? Separately, later, we can also intercept error messages coming from verification and have them attached as notes (not for this PR).

PR #646, opened 2025-12-28
Separate wave control flow lowering into dedicated pass

tyb0807:
Creates a new LowerWaveControlFlowPass that specifically handles wave.iterate -> scf.for conversion, avoiding the issue where mixed lowering was creating unlegalizable IR.

The problem is that if LowerWaveToMLIRPass tries to do both:
1. Control flow lowering (wave.iterate -> scf.for)
2. Operation lowering (wave.read, wave.mma, etc. -> standard ops)

This will cause ` wave.iterate` bodies to be cloned with unconverted wave operations inside `scf.for` loops, creating IR that couldn't be legalized.

The solution separates the pipeline into two phases:
1. LowerWaveControlFlowPass: wave.iterate -> scf.for (leaves other wave ops unchanged)
2. LowerWaveToMLIRPass: Remaining wave ops -> standard MLIR ops

This enables a clean two-phase lowering pipeline where control flow is handled first, then remaining wave operations are lowered in the standard `scf.for` context.

### DIFF
--- a/water/test/Dialect/Wave/ops.mlir
+++ b/water/test/Dialect/Wave/ops.mlir
@@ -228,10 +228,10 @@ func.func @mma_elements_per_thread_interface_explicit() attributes {
   %rhs_init = arith.constant 2.0 : f16
   %acc_init = arith.constant 0.0 : f32
 
-  // Create register values with explicit elements_per_thread.
-  %lhs = wave.register %lhs_init {elements_per_thread = 8} : !wave.tensor<[@M, @K] of f16, <register>>
-  %rhs = wave.register %rhs_init {elements_per_thread = 8} : !wave.tensor<[@N, @K] of f16, <register>>
-  %acc = wave.register %acc_init {elements_per_thread = 16} : !wave.tensor<[@M, @N] of f32, <register>>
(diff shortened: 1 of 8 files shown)


water/test/Dialect/Wave/ops.mlir
```diff
@@ -254,9 +254,9 @@ func.func @mma_elements_per_thread_32_threads_explicit() attributes {
   %rhs_init = arith.constant 2.0 : f16
   %acc_init = arith.constant 0.0 : f32
 
-  // Create register values with explicit elements_per_thread.
-  %lhs = wave.register %lhs_init {elements_per_thread = 8} : !wave.tensor<[@M, @K] of f16, <register>>
-  %rhs = wave.register %rhs_init {elements_per_thread = 8} : !wave.tensor<[@N, @K] of f16, <register>>
+  // Create register values - elements_per_thread determined by MMA backward propagation.
+  %lhs = wave.register %lhs_init : !wave.tensor<[@M, @K] of f16, <register>>
+  %rhs = wave.register %rhs_init : !wave.tensor<[@N, @K] of f16, <register>>
```
ftynse: Ditto.

PR #667, opened 2026-01-02
[water] Use distributed shape when accessing allocated LDS

### LINKED ISSUE #659: Use `distributed_shape` for Read/WriteOp accessing memory from AllocateOp

tyb0807:
When accessing shared memory allocated by wave.allocate, the read/write ops are supposed to operate on the distributed shape and not the logical shape. This PR implements the logic to handle this when lowering read/write ops to vector.load/store, transforming logical indices to distributed indices.

Fixes #659.

### DIFF
--- a/water/include/water/Dialect/Wave/IR/WaveOps.td
+++ b/water/include/water/Dialect/Wave/IR/WaveOps.td
@@ -270,6 +270,36 @@ def AllocateOp : WaveOp<"allocate"> {
     " (`in` $parent^ `:` type($parent))? attr-dict `:` type($result)";
 
   let hasVerifier = 1;
+
+  let extraClassDeclaration = [{
+    /// Returns true if this is a parent allocation (flat buffer with no index expressions).
+    /// Parent allocations have the following characteristics:
+    /// - No index attribute (no per-thread access patterns)
+    /// - Empty symbol list in distributed_shape
+    /// - Distributed shape rank = 1 (flattened buffer)
+    /// - Distributed shape is constant (no symbolic expressions)
+    bool isParentAllocation() {
+      mlir::ArrayAttr indexAttr = getIndexAttr();
+      if (indexAttr && !indexAttr.empty())
+        return false;  // Regular allocations have index expressions.
+
+      wave::WaveExprListAttr distributedShape = getDistributedShape();
+
+      // Empty symbol list in distributed_shape.
+      if (!distributedShape.getSymbols().empty())
+        return false;
+
+      // Distributed shape rank must be 1 (flattened).
+      if (distributedShape.getMap().getNumResults() != 1)
+        return false;
+
+      // Distributed shape must be constant.
+      if (!distributedShape.getMap().isConstant())
+        return false;
+
+      return true;
+    }
+  }];
 }
 
 def ExtractSliceOp : WaveOp<"extract_slice", [WaveInferTypeOpInterface, IdentityTypeInferenceOpTrait, CompatibleOperandsAndResultsOpTrait]> {
--- a/water/lib/Dialect/Wave/IR/WaveOps.cpp
+++ b/water/lib/Dialect/Wave/IR/WaveOps.cpp
@@ -89,6 +89,44 @@ static void printSingleSymbol(OpAsmPrinter &printer, Operation *,
 #define GET_OP_CLASSES
 #include "water/Dialect/Wave/IR/WaveOps.cpp.inc"
 
+/// Validates that index expressions match tensor dimension symbols.
+/// Returns failure if the count or symbols don't match.
+static LogicalResult
+verifyIndexDimensionMatch(Operation *op, ArrayAttr indexAttr,
+                          ArrayRef<wave::WaveSymbolAttr> tensorShape) {
+  if (!indexAttr || indexAttr.empty())
+    return success(); // No index expressions is valid
+
+  if (!llvm::hasSingleElement(indexAttr.getValue())) {
+    return op->emitError() << "'index' attribute must contain exactly one "
+                              "dictionary for this op, got "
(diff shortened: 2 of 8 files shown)


water/lib/Dialect/Wave/IR/WaveOps.cpp
```diff
@@ -89,6 +89,44 @@ static void printSingleSymbol(OpAsmPrinter &printer, Operation *,
...
+  auto indexDict = dyn_cast<DictionaryAttr>(indexAttr[0]);
+  if (!indexDict)
+    return success(); // Empty dictionary is valid
+
+  // Check count matches
+  if (indexDict.size() != tensorShape.size()) {
+    return op->emitError() << "number of index expressions ("
+                           << indexDict.size()
+                           << ") must match logical shape rank ("
+                           << tensorShape.size() << ")";
```
ftynse: This is a low-value diagnostic that suppresses the higher-value diagnostic below. If there is a missing symbol, it will just say "size mismatch" without saying what is missing, leaving the user figure out something that code should have done. You can instead keep the diagnostic below and add a diagnostic for symbols that are present in the index expression but not used. Arguably, this should be an error and not a warning since nothing bad will happen due to the unnecessary symbol being present

water/include/water/Dialect/Wave/IR/WaveOps.td
```diff
@@ -270,6 +270,36 @@ def AllocateOp : WaveOp<"allocate"> {
...
 
   let hasVerifier = 1;
+
+  let extraClassDeclaration = [{
+    /// Returns true if this is a parent allocation (flat buffer with no index expressions).
+    /// Parent allocations have the following characteristics:
+    /// - No index attribute (no per-thread access patterns)
+    /// - Empty symbol list in distributed_shape
+    /// - Distributed shape rank = 1 (flattened buffer)
+    /// - Distributed shape is constant (no symbolic expressions)
```
ftynse: You need to better explain this distinction in the op documentation. Eventually, we should create a proper "view" operation to fix this awful design mistake of calling an allocation something that isn't allocating.

water/lib/Dialect/Wave/IR/WaveOps.cpp
```diff
@@ -107,6 +145,29 @@ llvm::LogicalResult wave::AllocateOp::verify() {
...
   }
 
+  // For regular allocations (not child or parent allocations), enforce rank
+  // consistency.
+  if (!getParent() && !isParentAllocation()) {
+    ArrayAttr indexAttr = getIndexAttr();
+    wave::WaveExprListAttr distributedShape = getDistributedShape();
+    wave::WaveTensorType resultType = getResult().getType();
+
+    uint64_t logicalRank = resultType.getShape().size();
```
ftynse: I highly suspect that `.size()` returns `size_t` and not `uint64_t`. If you for whatever reason insist on using typedefed integers instead of plain `unsigned` like most of LLVM-based code, use the right ones.

water/lib/Dialect/Wave/Transforms/LoweringPatterns.cpp
```diff
@@ -56,27 +56,61 @@ class AllocateOpLoweringPattern : public OpConversionPattern<wave::AllocateOp> {
...
+      }
+    }
+
+    if (isUsedAsParent) {
+      // For allocations that will be used as parents for views,
+      // we need to allocate a flat buffer rather than a multi-dimensional
+      // memref.
+      ArrayRef<int64_t> shape = memrefType.getShape();
+      int64_t totalElements = 1;
+      for (int64_t dim : shape) {
```
ftynse: Don't we require "parent" allocation to be 1D anyway?

water/include/water/Dialect/Wave/IR/WaveOps.td
ftynse: I also think the distributed shape being constant is an unnecessary early design limitation (or incompetence) that we should eventually lift.

tyb0807:
> Have you thought of alternatives that don't require jumping through hoops in lowering?

ftynse:
> Yes, and I came to conclusion that doing this transformation during lowering might be the least worst.

Perhaps. This is something that a commit/PR message can explain, so the reviewers and future contributors can understand the reasoning, not guess it.

> What other alternatives you have in mind? Please shed some light on this matter!

I did not really think about it, just reviewing what I see. Given the conjunction of conditions in the code that are required for the lowering to work, it looks excessively fragile, which makes me worried about its scalability and longer-term health. Hence the question: it may well be the best available approach, or there may be others, I just don't know and don't have enough information to judge, so I'm requesting it. If I had to figure it out myself, I might as well write the code at that point.

The specific part that appears problematic to me, though it was present in the code already, is having to "look through" the unrealized conversion cast.  This feels like leaking the abstraction of the dialect conversion, which may be the wrong one here as we have the same wave tensor type converted to different upstream types based on usage. 

> Ok, the more I think about this, the more I feel the cleanest solution would be: [...]

Indeed, this looks like the least "internally complex" approach. It can be similar to, or even integrate with, the one we have for register-resident tensors. Note that it is possible to call `setType` on a value, the only time when you need to re-create an operation is when its _number_ of results changes. Note also that calling `setType` from within dialect conversion will likely cause deep problems with the infra, so it will have to be a separate pass.

> This doesn't sound any less work tbh.

It may well be the same amount of work, but is consistent with the approach taken for register-resident tensors that are converted to vectors, so less overall complexity in the system (one approach rather than two approaches). I don't necessarily insist on one approach or another, I haven't thought too much about the problem, but the more complex the logic is, the stronger should be the arguments behind it.

tyb0807:
Superseded by #677, #684 and https://github.com/iree-org/wave/pull/686.

PR #677, opened 2026-01-06
[water] Create a pass to resolve distributed shapes for AllocateOp

tyb0807:
This pass transforms wave.allocate operations with distributed_shape attribute from returning WaveTensorType to returning MemRefType with the resolved distributed shape. This makes the physical memory layout explicit in the type system before lowering.

Additionally, update wave.read/wave.write op to accept memory operands of MemRefType.

### DIFF
--- a/water/include/water/Dialect/Wave/Transforms/Passes.td
+++ b/water/include/water/Dialect/Wave/Transforms/Passes.td
@@ -165,8 +165,7 @@ def WaterWaveResolveDistributedAllocationsPass
   }];
 
   let dependentDialects = [
-    "::mlir::gpu::GPUDialect",
-    "::mlir::memref::MemRefDialect"
+    "::mlir::gpu::GPUDialect"
   ];
 }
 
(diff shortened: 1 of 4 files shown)


water/include/water/Dialect/Wave/Transforms/Passes.td
```diff
@@ -147,4 +147,26 @@ def WaterWavePropagateDefaultsFromConstraintsPass
...
+    attribute from returning WaveTensorType to returning MemRefType with the
+    resolved distributed shape. This makes the physical memory layout explicit
+    in the type system before lowering.
+
+    The pass also updates all users of the allocation (wave.read and wave.write
+    operations) to use the new memref-typed value.
+
+    This pass must run before LowerWaveToMLIR to ensure that read/write
+    operations can directly use the distributed memref without needing to look
+    through unrealized_conversion_cast operations.
```
ftynse: Not for this commit: this can be turned into a normal form.

water/include/water/Dialect/Wave/Transforms/Passes.td
```diff
@@ -147,4 +147,26 @@ def WaterWavePropagateDefaultsFromConstraintsPass
...
+  }];
+
+  let dependentDialects = [
+    "::mlir::memref::MemRefDialect"
```
ftynse: Is this needed? Memref type doesn't belong to the memref dialect...

ftynse (commented):
Did you forget to `git add` the test for the new pass?

PR #686, opened 2026-01-07
[water] Support read/write lowering with MemRefType memory operands

tyb0807:
After ResolveDistributedAllocations converts WaveTensorType to MemRefType,
read/write ops need to determine dimension ordering for correct lowering.

With IndexExprsSpecified as a precondition for LowerWaveToMLIR, read/write
ops are guaranteed to have index expressions. Since DictAttr is internally
an ArrayRef<NamedAttribute>, the index dictionary keys are ordered and can
be used directly for dimension ordering.

Fixes https://github.com/iree-org/wave/issues/659.


ftynse:
> This is needed because the index dictionary keys are
dimension names, and we need to know their ordering for correct lowering.

`DictAttr` is an `ArrayRef<NamedAttribute>` internally, so it is always ordered (otherwise we would have had massive flakiness or all tests without custom syntax for attributes). So do we actually need this? Maybe we rather need a verifier that the index attribute is present on memory operations that use memref types?

PR #687, opened 2026-01-07
Added support for extract op dynamic position

xintin:
* added support for dynamic position in extract op
* added a lit test 
* ran ruff on the touched files.


harsh-nod (approved):
lgtm!

PR #689, opened 2026-01-08
[water] Require IndexExprsSpecified normal form for LowerWaveToMLIR

tyb0807:
Add IndexExprsSpecified to the preconditions for LowerWaveToMLIR pass. This ensures all memory operations have index expressions before lowering, removing the need for runtime checks in the lowering patterns.

### DIFF
--- a/water/lib/Dialect/Wave/IR/WaveAttrs.cpp
+++ b/water/lib/Dialect/Wave/IR/WaveAttrs.cpp
@@ -897,23 +897,28 @@ wave::detail::verifyNormalFormAttr(Operation *root, wave::WaveNormalForm form,
                                  wave::WaveNormalForm::IndexExprsSpecified)) {
       if (op->hasTrait<wave::HasWaveIndexMapping>() &&
           !op->getAttr(wave::WaveDialect::kIndexWaveExprListAttrName)) {
-        // Only require index expressions for ops with WaveTensorType
-        // operands/results, or Read/Write ops (which access memory).
-        auto isWaveTensor = [](Type t) {
-          return llvm::isa<wave::WaveTensorType>(t);
-        };
+        // Only require index expressions for read/write ops, or ops with
+        // WaveTensorType operands/results. Vector-only ops (after
+        // elements-per-thread propagation) don't need index expressions.
         bool hasWaveTensor =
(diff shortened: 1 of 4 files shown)


water/lib/Dialect/Wave/IR/WaveAttrs.cpp
```diff
@@ -897,6 +897,19 @@ wave::detail::verifyNormalFormAttr(Operation *root, wave::WaveNormalForm form,
                                  wave::WaveNormalForm::IndexExprsSpecified)) {
       if (op->hasTrait<wave::HasWaveIndexMapping>() &&
           !op->getAttr(wave::WaveDialect::kIndexWaveExprListAttrName)) {
+        // Only require index expressions for ops with WaveTensorType
+        // operands/results, or Read/Write ops (which access memory).
+        auto isWaveTensor = [](Type t) {
+          return llvm::isa<wave::WaveTensorType>(t);
+        };
+        bool hasWaveTensor =
+            llvm::any_of(op->getOperandTypes(), isWaveTensor) ||
```
ftynse:
```suggestion
            llvm::any_of(op->getOperandTypes(), llvm::IsaPred<wave::WaveTensorType>) ||
```

water/lib/Dialect/Wave/IR/WaveAttrs.cpp
ftynse: The message needs to be updated so it is not confusing

wave_lang/kernel/wave/water.py
```diff
@@ -494,6 +494,8 @@ def apply_water_middle_end_passes(mlir_text: str) -> str:
     binary = get_water_opt()
 
     # Define the pass pipeline for Wave lowering
+    # Note: water-wave-infer-index-exprs is not included because for vector-only
+    # IR (after elements-per-thread propagation), index expressions aren't needed.
```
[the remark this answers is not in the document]
tyb0807: Yeah we don't need to run the pass here because the index expressions are already generated by PyWave. Once we are able to run water on the pristine fx.graph (without any transformations from Python side), we'll need to include `water-wave-infer-index-exprs` and the C++ version of the Python passes prior to index sequence analysis currently.

PR #690, opened 2026-01-08
[water] make WaveIndexMappingAttr more robust against null elements

ftynse:
Since 93488b339ebcf37d0699fa999d6bedab6984f4ce, we allow null elements in the indexing map attribute, make the logic in printing and conversion from mlir to python more robust to there being a null expression in the mapping.

Drop redundant verification that was not accounting for null elements, but never actually reporting any errors since the attribute verifier had been intercepting the same cases.

### DIFF
--- a/tests/mlir_wave_iface/mlir_to_wave_test.py
+++ b/tests/mlir_wave_iface/mlir_to_wave_test.py
@@ -291,6 +291,22 @@ class TestConvertIndexMappingAttrToSympy:
         assert result.size == 16
         assert result.stride == 1
 
+    def test_index_mapping_with_null_start(self):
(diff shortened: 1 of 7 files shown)


martin-luecke (approved):
LGTM!

I have been looking for the addition of the parsing of `<NULL>`, but it seems this has been added previously. I think it would be more accurate to say that this PR adds support for null elements in the Python to MLIR conversion. The description could be clearer in that regard.
I think it would be good to document this change on the [`WaveIndexMappingAttr`](https://github.com/iree-org/wave/blob/dc52d09dee904c0deb4fac6824b1782bdfe59591/water/include/water/Dialect/Wave/IR/WaveAttrs.td#L386). I.e., that it supports NULL elements only for better compatibility with pywave, and this is not something we expect to be able to lower successfully.

PR #692, opened 2026-01-08
[water] sideways and non-conflictual index for wave.write

ftynse:
Add two features to index inference that make it consistent with PyWave behavior for memory writes even though they appear at odds with classical dataflow propagation.

1. Propagate index expressions lattice values between two operands of write operation during backward analysis (instead of propagating from results that write does not have anyway). This requires manually adding supplementary dependencies between lattice states in the framework since it is not set up for such sideways propagation. Practically, this ensures we index expressions are propagated between different _uses_ of the same memory operand, which in turn allows us to infer index expressions for, e.g., reads from global memory that are stored to shared memory before being reloaded. Arguably, this is not desirable and is concretely at odds with the elements-per-thread propagation, but this is currently expected by Pywave that corrects index sequences in a later "minimize global loads" pass that would otherwise be considered an optimization. Note that inferred index expressions are correct, at least for simple cases, just severely suboptimal.

2. Not update an operand lattice from another operand in a write operation if such an update would lead to a conflict as indicated by newly reaching lattice top state. This is equally questionable from the UX perspective since subtle changes in the input that would lead to conflicts without a write would not lead to them with a write. This is necessary, however, to avoid getting conflicts due to 1.

### DIFF
--- a/water/lib/Dialect/Wave/IR/WaveOps.cpp
+++ b/water/lib/Dialect/Wave/IR/WaveOps.cpp
@@ -6,12 +6,14 @@
 
 #include "water/Dialect/Wave/IR/WaveOps.h"
 
+#include "mlir/IR/AffineExpr.h"
 #include "mlir/IR/Attributes.h"
 #include "mlir/IR/Builders.h"
 #include "mlir/IR/BuiltinTypes.h"
 #include "mlir/IR/OpImplementation.h"
 #include "mlir/IR/PatternMatch.h"
 #include "mlir/Interfaces/ControlFlowInterfaces.h"
+#include "mlir/Interfaces/FunctionInterfaces.h"
 #include "mlir/Transforms/RegionUtils.h"
 #include "water/Dialect/Wave/IR/IndexExpr.h"
 #include "water/Dialect/Wave/IR/WaveAttrs.h"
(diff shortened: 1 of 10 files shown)


water/lib/Dialect/Wave/IR/WaveOps.cpp
```diff
@@ -991,34 +993,52 @@ populateMmaIndexingExpr(wave::WaveMmaKind kind, bool isAccumulator,
...
+/// constraints or MMA shapes. The first argument indicates for which operation
+/// the constraints are being used, which is in particular necessary to only
+/// apply tiling constraints inside the relevant loops.
 static void mixInThreadIndependentConstraints(
-    llvm::ArrayRef<wave::WaveSymbolAttr> indexingSymbols,
+    Operation *where, llvm::ArrayRef<wave::WaveSymbolAttr> indexingSymbols,
     const llvm::DenseMap<wave::WaveSymbolAttr, llvm::SmallVector<Attribute>>
         &symbolConstraints,
-    llvm::SmallVector<NamedAttribute> &symbolMappings) {
+    llvm::SmallVector<mlir::NamedAttribute> &symbolMappings) {
```
```suggestion
    llvm::SmallVector<NamedAttribute> &symbolMappings) {
```

water/lib/Dialect/Wave/IR/WaveOps.cpp
```diff
@@ -1673,14 +1693,32 @@ llvm::FailureOr<ChangeResult> wave::WriteOp::propagateIndexExprsForward(
   return ChangeResult::NoChange;
 }
 
-// WriteOp has no results, nothing to propagate to. Not propagating sideways
-// between operands either, the same index expression is used for the memory
-// operand, we can then read from the same memory with different index.
+// Propagating "sidewise" between operands, but only if this would not result in
```
```suggestion
// Propagating "sideways" between operands, but only if this would not result in
```
typo?

ftynse:
The unhappy CI is unhappy...

PR #701, opened 2026-01-09
gfx1250: preliminary triple buffer schedule

panditsa:
- Pass tags if available on TensorLoadToLDS nodes
- Merge tags if we fuse TensorLoadToLDS
- Sample triple buffer schedule


Hardcode84:
Can you add test, similar to `testGemmCodegen` to check it at least compiles?

Hardcode84 (approved):

PR #714, opened 2026-01-13
[water] Add ordered_syms attribute to preserve dimension ordering

tyb0807:
MLIR's DictionaryAttr sorts entries alphabetically by key, which loses the semantic dimension ordering. This caused incorrect code generation with transposed memory indices.

This change adds an ordered_syms attribute to ReadOp and WriteOp that captures the dimension ordering from WaveTensorType before it is converted to MemRefType. The ResolveDistributedAllocations pass now sets this attribute, and a new OrderedSymsSpecified normal form bit ensures this happens before LowerWaveToMLIR.

For ops with WaveTensorType memory operands, the ordering is extracted from the type's shape. For MemRefType memory operands (after allocation resolution), the ordered_syms attribute is used.

### DIFF
--- a/water/lib/Dialect/Wave/Transforms/ResolveDistributedAllocations.cpp
+++ b/water/lib/Dialect/Wave/Transforms/ResolveDistributedAllocations.cpp
@@ -33,7 +33,6 @@ struct ResolveDistributedAllocations
   /// memory operand's WaveTensorType shape.
   void setOrderedSymsOnReadWriteOps(Operation *root) {
     root->walk([&](ReadOp readOp) {
-      // Skip if already set
       if (readOp.getOrderedSymsAttr())
         return;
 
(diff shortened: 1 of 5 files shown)


water/lib/Dialect/Wave/Transforms/ResolveDistributedAllocations.cpp
```diff
@@ -29,6 +29,46 @@ struct ResolveDistributedAllocations
...
+      auto tensorType = dyn_cast<WaveTensorType>(readOp.getMemory().getType());
+      if (!tensorType)
+        return;
+
+      // Convert shape symbols to ArrayAttr
+      SmallVector<Attribute> symbolAttrs;
+      for (WaveSymbolAttr sym : tensorType.getShape())
+        symbolAttrs.push_back(sym);
+
+      readOp.setOrderedSymsAttr(
```
ftynse: Nit: can we call `setOrderedSyms` instead and not construct the attribute?

martin-luecke (approved):
With your addressing of previous reviews, this looks good to me. 
Please fix the conflict, and we are good to merge.

PR #730, opened 2026-01-14
[water] Replace DictionaryAttr with WaveIndexExprsAttr for ordered index expressions

tyb0807:
Index expressions map tensor dimensions to thread/workgroup coordinates for memory access lowering. The dimension order in these expressions must match the tensor type's shape order (e.g., `[@M, @K] -> {M: ..., K: ...}`). However, DictionaryAttr alphabetically sorts its entries, so `{M, K}` becomes `{K, M}`. This causes incorrect lowering when the tensor type is no longer available (e.g., after conversion to memref).

This PR introduces WaveIndexExprsAttr, an ordered attribute that preserves insertion order. It consists of:
  - WaveIndexEntryAttr: A single (dimension, mapping) pair where dimension is a WaveSymbolAttr and mapping is a WaveIndexMappingAttr
  - WaveIndexExprsAttr: An ordered array of WaveIndexEntryAttr entries

  Syntax: `#wave.index_exprs<[@M : <mapping>, @K : <mapping>]>`

The join() operation in lattice analysis uses a LHS-first policy (LHS entries first, then RHS-only entries) as a conventional choice to ensure deterministic output.

Changes:
  - Add WaveIndexEntryAttr and WaveIndexExprsAttr attribute definitions
  - Update IndexExprsLatticeStorage to use WaveIndexExprsAttr and MapVector for order preservation
  - Add C API and Python bindings
  - Update Python emitter/converter

### DIFF
--- a/water/include/water/Dialect/Wave/IR/WaveAttrs.td
+++ b/water/include/water/Dialect/Wave/IR/WaveAttrs.td
@@ -589,4 +589,13 @@ def WaveIndexExprsAttr : AttrDef<WaveDialect, "WaveIndexExprs"> {
   }];
 }
 
+//-----------------------------------------------------------------------------
+// Typed array attributes
+//-----------------------------------------------------------------------------
+
+def WaveIndexExprsArrayAttr : TypedArrayAttrBase<WaveIndexExprsAttr,
+  "array of WaveIndexExprsAttr"> {
+  let constBuilderCall = "$_builder.getArrayAttr($0)";
+}
+
 #endif // WATER_DIALECT_WAVE_WAVEATTRS
--- a/water/include/water/Dialect/Wave/IR/WaveInterfaces.h
+++ b/water/include/water/Dialect/Wave/IR/WaveInterfaces.h
@@ -42,11 +42,6 @@ public:
   }
 };
 
-mlir::ParseResult parseWaveIndexDict(mlir::OpAsmParser &parser,
-                                     mlir::ArrayAttr &out);
-void printWaveIndexDict(mlir::OpAsmPrinter &printer, mlir::Operation *op,
-                        mlir::ArrayAttr arr);
-
 //-----------------------------------------------------------------------------
 // WaveInferTypeOpInterface and implementation traits
 //-----------------------------------------------------------------------------
@@ -380,7 +375,10 @@ public:
 
 // Lattice for propagating index expressions across wave dialect operations.
 // In addition to the bottom and top states, it can represent a concrete state
-// manifested as a dictionary attribute mapping symbol names to index mappings.
+// manifested as a WaveIndexExprsAttr mapping dimension symbols to index
+// mappings. The entries preserve order, corresponding to the tensor type's
+// dimension order.
+//
 // The JOIN function is defined similarly to other lattices with special
 // handling for combining thread-dependent and thread-independent index
 // expressions.
@@ -388,7 +386,7 @@ class IndexExprsLatticeStorage {
 public:
   IndexExprsLatticeStorage();
   IndexExprsLatticeStorage(const IndexExprsLatticeStorage &value) = default;
-  IndexExprsLatticeStorage(mlir::DictionaryAttr concreteValue);
+  IndexExprsLatticeStorage(wave::WaveIndexExprsAttr concreteValue);
 
   IndexExprsLatticeStorage &
   operator=(const IndexExprsLatticeStorage &other) = default;
@@ -404,7 +402,7 @@ public:
 
   // Returns the concrete value stored in the lattice instance, be it fully
   // specified or not, or null if the lattice instance is a top or a bottom.
-  mlir::DictionaryAttr getConcreteValue() const;
+  wave::WaveIndexExprsAttr getConcreteValue() const;
 
   // Return the top lattice instance.
   static IndexExprsLatticeStorage top();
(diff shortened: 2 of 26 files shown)


water/include/water/Dialect/Wave/IR/WaveAttrs.td
```diff
@@ -519,4 +519,83 @@ def WaveReadWriteBoundsAttr : AttrDef<WaveDialect, "WaveReadWriteBounds"> {
...
+  }];
+}
+
+//-----------------------------------------------------------------------------
+// Typed array attributes
+//-----------------------------------------------------------------------------
+
+def WaveIndexExprsArrayAttr : TypedArrayAttrBase<WaveIndexExprsAttr,
+  "array of WaveIndexExprsAttr"> {
+  let constBuilderCall = "$_builder.getArrayAttr($0)";
```
tgymnich: Can we leave this out and just use the default builder?

water/include/water/Dialect/Wave/IR/WaveAttrs.td
```diff
@@ -519,4 +519,83 @@ def WaveReadWriteBoundsAttr : AttrDef<WaveDialect, "WaveReadWriteBounds"> {
...
+    /// Complexity: O(n) where n is the number of dimensions (typically 2-4).
+    std::optional<::wave::WaveIndexMappingAttr>
+    lookup(::wave::WaveSymbolAttr dimension) const;
+
+    /// Look up by dimension name string.
+    std::optional<::wave::WaveIndexMappingAttr>
+    lookup(::llvm::StringRef dimensionName) const;
+
+    /// Get the ordered list of dimension symbols.
+    ::llvm::SmallVector<::wave::WaveSymbolAttr> getDimensions() const;
```
ftynse: Nit: I think you can make the return type `auto` if you put implementation here, and just `return llvm::map_range(...)` (the actual type is atrocious), which will avoid the need to construct and potentially copy a vector.

water/include/water/Dialect/Wave/IR/WaveInterfaces.h
```diff
@@ -404,15 +402,38 @@ class IndexExprsLatticeStorage {
...
   // Return the bottom lattice instance.
   static IndexExprsLatticeStorage bottom();
 
-  // Join two lattice instances and return the result.
+  /// Join two lattice instances and return the result.
+  ///
+  /// Ordering semantics:
+  /// - LHS entries come first (in LHS order), then RHS-only entries (in RHS
+  ///   order).
+  /// - Entries with the same dimension have their mappings merged.
```
ftynse: Merged how exactly? This should refer to the actual implementation function that has this documentation.

water/include/water/Dialect/Wave/IR/WaveInterfaces.h
```diff
@@ -404,15 +402,38 @@ class IndexExprsLatticeStorage {
...
+  /// Valid usage scenarios:
+  /// 1. **Same dimensions, same order**: Both LHS and RHS have matching
+  ///    dimension order (e.g., both have {M, K}). The result preserves this
+  ///    order. This is the common case for elementwise ops like wave.add.
+  ///
+  /// 2. **MMA ops**: LHS has {M, K}, RHS has {N, K}, accumulator/result has
+  ///    {M, N}. The `ignoredRhsSymbols` parameter filters dimensions that
+  ///    don't apply (e.g., ignore M when propagating from LHS to result).
+  ///    Joins are done incrementally: start with bottom, join LHS (filtered),
+  ///    then join RHS, then join accumulator.
```
ftynse: I don't see why this is relevant and why we would want to describe in English what is obvious from the code. If the logic evolves this documentation will likely rot.

water/include/water/Dialect/Wave/IR/WaveInterfaces.h
```diff
@@ -404,15 +402,38 @@ class IndexExprsLatticeStorage {
...
+  ///
+  /// 3. **Iterate ops**: Block arguments are joined with iter_args, and
+  ///    terminator operands with results. Both should have matching tensor
+  ///    types and thus matching dimension order.
```
ftynse: This describes a quite fundamental verification constraint that any control-flow operation follows, why do we need it here exactly?

water/include/water/Dialect/Wave/IR/WaveAttrs.td
```diff
@@ -519,4 +519,83 @@ def WaveReadWriteBoundsAttr : AttrDef<WaveDialect, "WaveReadWriteBounds"> {
...
+  let mnemonic = "index_entry";
+  let description = [{
+    A single entry mapping a tensor dimension symbol to its index mapping.
+
+    This is a component of WaveIndexExprsAttr, representing one dimension's
+    index expression. The dimension is a WaveSymbolAttr (e.g., @M, @K, @N)
+    and the mapping is a WaveIndexMappingAttr specifying start/step/stride.
+
+    Syntax: @M : [symbols] -> (start, step, stride)
+    Example: @M : [#wave.index_symbol<WG0>, #wave.symbol<"BLOCK_M">] -> (WG0 * BLOCK_M, 1, BLOCK_M)
```
ftynse: THis PR literally has tests that show different syntax. Don't put examples in here, they are bitrotten before you even sent your PR!

PR #734, opened 2026-01-14
[water] priority-based index expression propagation

ftynse:
This mimics the index exrepssion propagation behavior pywave has with
its multiple passes of propagation based on different "source" ops in a
single dataflow process, with convergence guarantees. Earlier "source"
operations assign higher-priority index expressions to values that
completely override lower-priority expressions on join. Only
equal-priority expressions perform actual per-dimension join.

Several design decisions are replicated for consistency purposes, but
should be revised in the future to provide a unified approach to
inference without surprises such as the "mode switch" that results in
completely different expressions when an mma is removed.

This is particularly true for the case of missing `vector_shape` part of
the hardware constraint that is only necessary when a "write" operation
isn't writing a value transitively obtained from an MMA. Supporting this
requires deferring diagnostic emission to account for the case where the
process would converge thanks to a higher-priority expression coming
from elsewhere.

### DIFF
--- a/wave_lang/kernel/wave/analysis/index_sequence_analysis.py
+++ b/wave_lang/kernel/wave/analysis/index_sequence_analysis.py
@@ -379,7 +379,20 @@ def _check_water_indices(trace: CapturedTrace, inferred: dict[str, IndexSequence
                 for dim, seq in seqs.items()
             }
 
-        node_index = ensure_symbols_positive(node.index)
+        # Filter out symbols that should not belong to the index sequence,
+        # such as those corresponding to reduction dimensions in operations
+        # defining the operands of this operation. Keep everything for
+        # oeprations that are reductions.
+        # TODO: fixing this properly requires redefining `indexing_dims` to
+        # be local rather than look at operands.
+        if not isinstance(custom, (MMABase, ReduceOp)):
+            node_index = {
+                k: v for k, v in node.index.items() if k in custom.type.symbolic_shape
+            }
+        else:
+            node_index = node.index
+
+        node_index = ensure_symbols_positive(node_index)
         inferred_index = ensure_symbols_positive(inferred_index)
 
         # Check that that indices match, raise an error if they don't. Start by


water/lib/Dialect/Wave/IR/WaveOps.cpp
```diff
@@ -1690,6 +1695,202 @@ llvm::FailureOr<ChangeResult> wave::WriteOp::propagateIndexExprsForward(
...
+    }
+
+    // TODO: copied from pywave... This may be the wrong place for this
+    // logic, we likely want to infer this number from the constraints
+    // somehow early.
+    auto it = initObject.symbolConstraints.find(tensorType.getShape()[i]);
+    // TODO: pywave just ignores this not sure if we want to, including the
+    // case below where there may be zero constraints. Interestingly, it
+    // asserts if trailing dimensions are not found when computing the
+    // stride...
```
ftynse: Is it safe to simply ignore the symbols for which there are no constraints when setting index sequences from write?

water/lib/Dialect/Wave/IR/WaveOps.cpp
```diff
@@ -1690,6 +1695,202 @@ llvm::FailureOr<ChangeResult> wave::WriteOp::propagateIndexExprsForward(
...
+    if (wgConstraint.getWorkgroupDim().getValue() == WaveWorkgroupDim::X) {
+      startExpr = startExpr % hardwareConstraint.getThreadsPerWave();
+    } else {
+      // TODO: in pywave, we always do `startExpr % threadsPerWave` where
+      // threadsPerWave == 1 for workgroup dims other than X, making it
+      // always zero. It mentions an assumption about the (64, 1, 1) thread
+      // shape, but it is unclear whether that assumption always holds.
+      // It looks like the intention for this was to express lane ID rather
+      // than thread ID, but it is unclear how it accounts for multiple
+      // wavefronts running in parallel.
```
ftynse: The comment in the original source (https://github.com/iree-org/wave/blob/601ab68fdba54eff066bee01ed060b537df69209/wave_lang/kernel/wave/constraints.py#L498-L501) says something about preventing double counting of thread id, but I can't infer where and why it would be counted twice. The support for it was added in a commit for atomics, https://github.com/iree-org/wave/commit/4eeee9ac0a8c2d5cf30ee0096bc9262d1f826393, which is doesn't provide an explanation either

water/lib/Dialect/Wave/IR/WaveInterfaces.cpp
```diff
@@ -909,17 +952,21 @@ wave::IndexExprsLatticeStorage wave::IndexExprsLatticeStorage::join(
...
-      DictionaryAttr::get(ctx, llvm::map_to_vector(result, [](auto &&pair) {
-                            return NamedAttribute(pair.first, pair.second);
-                          })));
+      DictionaryAttr::get(ctx, llvm::map_to_vector(result,
+                                                   [](auto &&pair) {
+                                                     return NamedAttribute(
+                                                         pair.first,
+                                                         pair.second);
+                                                   })),
+      std::max(lhs.getPriority(), rhs.getPriority()));
```
ftynse: do we need priority per-symbol?..

water/lib/Dialect/Wave/IR/WaveOps.cpp
ftynse: For other dimensions, I suppose the intent is that they are going to be expanded at which point they should just start with 0 and then the expansion will update them. No need to obfuscate that via modulo operations.

water/lib/Dialect/Wave/IR/WaveOps.cpp
```diff
@@ -1690,6 +1695,202 @@ llvm::FailureOr<ChangeResult> wave::WriteOp::propagateIndexExprsForward(
...
+    for (int64_t j = i + 1; j < e; ++j) {
+      Attribute vectorShape = hardwareConstraint.getVectorShapes().get(
+          tensorType.getShape()[j].getName());
+      if (!vectorShape) {
+        emitError() << "couldn't find vector shape for dimension "
+                    << tensorType.getShape()[j];
+        return failure();
+      }
+      stride *= cast<IntegerAttr>(vectorShape).getValue().getSExtValue();
+    }
```
ftynse: I have a strong suspicion that usage of strides is inconsistent: here we use linear row-major stride (changed in https://github.com/iree-org/wave/commit/1ddf92d9b5ddb445477afbb8acbf96d5c0ace7ad), but for MMAs these remain per-dimension strides. That being said, they don't seem to affect code generation at all, which makes me wonder why do we even use them (except for mma-style strides that may be involved in strided write splitting)....

water/include/water/Dialect/Wave/IR/WaveInterfaces.h
```diff
@@ -635,6 +648,10 @@ class IndexExprsLatticeStorage {
   // symbol indexing the value or one of the top/bottom flags.
   llvm::PointerIntPair<mlir::Attribute, 2> value;
 
+  // Priority of this value. Specific values with higher priority override
+  // values with lower priority in joins.
+  int32_t priority;
```
martin-luecke: The problem is that `priority` is currently not compared in the `==` operator. This means we might silently drop a higher priority and not propagate it.

water/lib/Dialect/Wave/IR/WaveInterfaces.cpp
martin-luecke: The Python approach is implicitly per-dimension: combine_indices
https://github.com/iree-org/wave/blob/49678f942a31a8aeeebf2568fe1811e21ab14fb3/wave_lang/kernel/wave/analysis/index_sequence_analysis.py#L735-L750
 only overwrites dimensions the source actually knows about (i.e. what's contained in `thread_dependent_index`). 

The case i currently see is with a batched MMA operation. The C++ priority mechanism is per-lattice, which forces the MMA to "claim" B even though `mixInThreadIndependentConstraints` only gives it a default initializer of (0,1,1). 
I don't think this is an issue for the kernels we currently look at, but if we want to distribute B accross waves, the corresponding Write would compute the correct index expression while the MMA would just have a default and overwrite it.

The solution to this specific problem could be for the MMA to only propagate the expressions for its M,N,K and not batch dimensions. e.g. here
https://github.com/iree-org/wave/blob/49678f942a31a8aeeebf2568fe1811e21ab14fb3/water/lib/Dialect/Wave/IR/WaveOps.cpp#L1188-L1196
we could drop `batchSymbols` and this not provide an expression for B.
The merging is already designed to handle this case:
https://github.com/iree-org/wave/blob/49678f942a31a8aeeebf2568fe1811e21ab14fb3/water/lib/Dialect/Wave/IR/WaveInterfaces.cpp#L1252-L1255

During propagation B would then get the correct expression from the Write.

While this is not a general solution with full priority per symbol as Python does, I think it should cover the currently reachable cases.

PR #740, opened 2026-01-15
Enable kernel argument preloading in ASM backend

harsh-nod:
Implement kernel argument preloading for gfx950/MI350X following LLVM's pattern:

1. Emit s_load into preload locations (s[2:3], s[4:5], etc.) BEFORE branch
2. s_waitcnt lgkmcnt(0)
3. s_branch to 256-byte aligned entry point
4. .p2align 8 for alignment
5. Copy from preload locations to SRD ranges (s_mov_b64)
6. Continue with kernel code

This ensures compatibility:
- If hardware preloading works: s_load is idempotent, values already there
- If hardware preloading fails: s_load provides the values

Changes:
- kernel_compilation_context.py: Restructured _emit_srd_prologue() to emit s_load into preload locations before branch, then copy to SRD ranges after. Added KERNARG_PTR_SGPRS constant and _emit_wgid_prologue() helper (DRY).
- kernel_generator.py: Added handler for _preload_branch pseudo-instruction
- kernel_ir.py: Updated is_branch to handle _preload_branch
- kernel_module_compiler.py: Gate preloading by target (gfx95*) and COv5+. Added MAX_PRELOAD_SGPRS limit (16 SGPRs = 8 pointer args).
- metadata_emitter.py: Added preloading AMDHSA metadata fields, guarded by COv5+
- kernel_expr_emitter.py: Added _handle_wgid() helper (DRY)
- instruction_defs/common.yaml: Added _preload_branch pseudo instruction
- docs/wave/asm_backend.rst: Added Kernel Argument Preloading documentation
- docs/wave/register_allocation.rst: Updated ABI docs with preloading SGPR layout

Expected benefit: ~12-25% speedup for small kernels by eliminating ~100 cycle s_load latency at kernel startup.

All 56 ASM backend tests passing.

### DIFF
--- a/wave_lang/kernel/wave/asm/instruction_defs/common.yaml
+++ b/wave_lang/kernel/wave/asm/instruction_defs/common.yaml
@@ -1098,6 +1098,14 @@ instructions:
     uses: []
     latency: 0
 
+  _srd_copy_base:
+    mnemonic: ""
+    category: pseudo
+    format: pseudo
+    defs: []
+    uses: []
+    latency: 0
+
   _srd_fill_size:
     mnemonic: ""
     category: pseudo
@@ -1164,3 +1172,14 @@ instructions:
     defs: []
     uses: []
     latency: 0
+
+  # Kernel argument preloading entry point pattern (s_branch to aligned label)
+  _preload_branch:
+    mnemonic: "s_branch"
+    category: pseudo
+    format: pseudo
+    defs: []
+    uses: []
+    latency: 0
+    special:
+      branch: true
--- a/wave_lang/kernel/wave/asm/kernel_compilation_context.py
+++ b/wave_lang/kernel/wave/asm/kernel_compilation_context.py
@@ -56,6 +56,10 @@ from .kernel_loops import _LoopSupport
 from .kernel_mfma import _MFMASupport
 from .kernel_passes import _CompilationPasses
 
+# ABI Constants
+# Number of SGPRs used for kernarg segment pointer (s[0:1])
+KERNARG_PTR_SGPRS = 2
+
 
 @dataclass
 class KernelCompilationContext(_LoopSupport, _MFMASupport, _CompilationPasses):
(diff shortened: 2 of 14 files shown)


wave_lang/kernel/wave/asm/instruction_defs/common.yaml
```diff
@@ -1164,3 +1172,15 @@ instructions:
     defs: []
     uses: []
     latency: 0
+
+  # Kernel argument preloading entry point pattern (s_branch to aligned label)
+  # Branch target label is stored in comment for CFG compatibility
```
xintin: For my knowledge, can we store target label as a field here instead of the comment? It might be more robust.

wave_lang/kernel/wave/asm/kernel_compilation_context.py
```diff
@@ -150,20 +154,44 @@ class KernelCompilationContext(_LoopSupport, _MFMASupport, _CompilationPasses):
...
     # Statistics
     _cse_hits: int = field(default=0, init=False)
 
+    # Precolored SGPRs: virtual SGPR -> physical SGPR index
+    # Used to force kernarg SGPRs to their hardware-preloaded locations
+    _precolored_sregs: Dict[KSReg, int] = field(default_factory=dict, init=False)
+
     def __post_init__(self):
+        # Calculate preload SGPR count (2 SGPRs per 64-bit pointer arg)
+        preload_sgpr_count = self.num_kernargs * 2 if self.use_kernarg_preloading else 0
```
xintin: is it possible here that we have 32 bit scalars kernargs?

harsh-nod: Yes it is possible. I have added a check so that if it is not a memref or stream.binding (which is 64-bit) we throw an error.

xintin (approved):
LGTM! One quick question: do we always preload all the args or only those which use SRDs?
Basically, can we do subset args preload or it is always better to preload all args?

harsh-nod:
> LGTM! One quick question: do we always preload all the args or only those which use SRDs? Basically, can we do subset args preload or it is always better to preload all args?

It is better to preload all args since we can avoid a lot of cycles (~100 on the smaller ones i tested) by doing so.

PR #743, opened 2026-01-16
[asm] Expression-level CSE: cache intermediate results in Add/Mul han…

harsh-nod:
…dlers

When computing compound expressions like 'a + b + c', the expression emitter now caches intermediate results (e.g., 'a + b') in addition to the final result. This eliminates redundant instructions when the same intermediate value is needed later.

Example: If we compute 'a + b + c' and later need 'a + b':
- Before: emits v_add_u32 twice for 'a + b'
- After: reuses cached result from first computation

The same approach is applied to Mul expressions with multiple factors.

Added lit test (test_cse_intermediate_caching) that verifies:
- Lane ID computation appears only once (not 4x for 4 tensor reads)
- All buffer loads and MMA instructions execute correctly

### DIFF
--- a/wave_lang/kernel/wave/asm/kernel_expr_emitter.py
+++ b/wave_lang/kernel/wave/asm/kernel_expr_emitter.py
@@ -148,7 +148,7 @@ class KernelIRExprEmitter(_FloorExpressionOps):
         result = self.ctx.vreg()
         self.ctx.program.emit(
             KInstr(
-                "v_mov_b32",
+                Instruction.V_MOV_B32,
                 (result,),
                 (sreg,),
                 comment=f"{name} from s{sreg.index}",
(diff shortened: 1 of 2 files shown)


wave_lang/kernel/wave/asm/kernel_expr_emitter.py
```diff
@@ -873,31 +896,56 @@ def _get_or_emit_impl(self, expr, cache_in_global: bool = False) -> KVReg:
...
                 else:
-                    self.ctx.program.emit(
-                        KInstr(
-                            "v_add_u32", (new_result,), (result, arg_reg), comment="add"
+                    # Emit new instruction
+                    new_result = self.ctx.vreg()
+
+                    # Check if we can use OR instead of ADD
+                    # OR is valid when bit ranges don't overlap
+                    if not self._bits_overlap(result_range, arg_range):
```
xintin: Should result_range be updated with new_result? If it hits cached_intermediate, only result is getting updated.

xintin (approved):

wave_lang/kernel/wave/asm/kernel_expr_emitter.py
harsh-nod: good catch, I will fix it. Thanks!

harsh-nod: result_range was not being handled correctly, so I fixed that. Also refactored the whole file to make it easier to read and remove string dependence.

PR #751, opened 2026-01-19
[water] add wave.extract op

tgymnich:
Add `wave.extract` op that extracts a single element vector from a `WaveTensorInRegister`.
Lowers to `vector.extract` + `vector.broadcast`.


water/test/Dialect/Wave/ops.mlir
```diff
@@ -21,6 +21,21 @@ func.func @extract_slice(%memory: !wave.tensor<[@A, @B] of f16>) -> !wave.tensor
...
+func.func @extract_static(%source: !wave.tensor<[@A] of f32>) -> !wave.tensor<[@A] of f32> {
+  // CHECK: wave.extract
+  %0 = wave.extract %source[#wave.expr_list<[] -> (2)>] : (!wave.tensor<[@A] of f32>) -> !wave.tensor<[@A] of f32>
+  return %0 : !wave.tensor<[@A] of f32>
+}
+
+// CHECK-LABEL: @extract_dynamic
+func.func @extract_dynamic(%source: !wave.tensor<[@A] of f32>) -> !wave.tensor<[@A] of f32> {
+  // CHECK: wave.extract
+  // CHECK: #wave.index_symbol<T0>
```
martin-luecke: nit: do `CHECK-SAME` to ensure it matches on the same line as the op.
Not technically required here as it can't accidentally match in the next test due to `CHECK-LABEL` but still good practice.

PR #758, opened 2026-01-20
Fix handling of GetResult injection for complex flow / Placeholder

willghatch:
To access results of `Iterate` or `Conditional` nodes, you need to use `GetResult`.  However, machinery around `GetResult` would fail when used on `Placeholder` nodes, which are used to reference nodes from other graphs, eg. when inside another `Iterate` or `Conditional` node.  IE this PR fixes issues that arise when using multiple loops and conditionals in kernels like StreamK.

### DIFF
--- a/wave_lang/kernel/wave/utils/graph_utils.py
+++ b/wave_lang/kernel/wave/utils/graph_utils.py
@@ -144,6 +144,22 @@ def erase_graph(graph: fx.Graph):
         graph.erase_node(node)
 
 
+def _placeholder_captures(placeholder_node: fx.Node, target: fx.Node) -> bool:
+    """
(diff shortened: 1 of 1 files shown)


wave_lang/kernel/wave/utils/graph_utils.py
```diff
@@ -337,7 +356,14 @@ def get_inputs(
             inputs.append(input)
 
     inputs = [propagate_placeholders(i) for i in inputs if i is not None]
-    return inputs, region
+    # Flatten any tuples/lists in inputs and filter out None values
+    flattened_inputs = []
+    for inp in inputs:
+        if isinstance(inp, (tuple, list)):
```
harsh-nod: Could you check if isinstance of Sequence?

PR #780, opened 2026-01-26
Unified swizzling for mixed granularity LDS access

adedespirlet:
This PR extends the LDS swizzling mechanism to support scenarios with mismatched instruction granularities (e.g  2 element gather_to_lds writes vs 8 element reads).
The current implementation only applies swizzling when the producer and consumer share the same granularity. If they differ, swizzling is skipped entirely, leaving potential bank conflicts unaddresed.

**Implementation:**
1.  **Swizzle unit**: a unified swizzle unit is implemented based on the maximum granularity across the gather and read op. For example, if gather_to_lds writes 8 elements and vector.load reads 4 elements, the swizzle unit is 8.
2. **Max_phase:** max_phase is now calculated relative to the swizzle unit rather than individual operation granularities. This prevents OOB accesses when a large swizzle unit is multiplied by a phase count derived from smaller granularity operations.
3. 	**Internal offset handling**: an internal_offset is computed to preserve the thread's position within the swizzle unit when you access a chunk of the swizzle unit 

**What has been added:**
- Added lit_test for the new swizzling implementation (test_gather_to_shared_with_mixed_granularity_swizzling )
- Added new option **enable_swizzle** (default=False) to WaveCompileOptions for explicit control over swizzling in tests 
- Updated **test_gather_to_shared_wave_tile_aligned_coalescing** to use enable_swizzle=False  to avoid interference with coalescing affine map verification
- Updated **test_gather_to_shared_scaled_dims** with new swizzle pattern
- Updated **testScaledBatchedGemmMXFP4Codegen** to reflect new VGPR counts and waitcnt patterns with swizzling enabled

**Limitations**

Swizzling only works for the GEMM kernels and is currently disabled for all the attention kernels. It will require separate analysis


wave_lang/kernel/wave/gather_to_shared.py
```diff
@@ -627,19 +612,19 @@ def gather_to_shared_swizzling(
...
         distributed_shape = mem_custom.distributed_shape
         col_size = subs_idxc(distributed_shape[-1])
-        num_col_chunks = col_size // elements_per_thread
-        max_phase = min(8, num_col_chunks)
-        logger.info(
-            f"distributed_shape={distributed_shape}, col_size={col_size}, "
-            f"num_col_chunks={num_col_chunks}, max_phase={max_phase}"
-        )
+
+        # Make sure swizzling logic respects elements_per_thread in gather and read. Take the maximum value to prevent swizzling of smaller chunks which could break order of elements
```
nithinsubbiah: nit: even comment lengths

PR #791, opened 2026-01-30
[water] fixes to wave.extract

ftynse:
This operation normally appears after expansion, so after analyses and should not participate in them. We also don't exactly know how it would participate in them as it refers to a partially-lowered state reasoning about vectors.

It shouldn't have identical tensor types before and after given its semantics, and such typing would contradict the Python equivalent.

Also enforce vector-level invariants when the operation is in vector form.


tgymnich (approved):
LGTM

PR #796, opened 2026-01-31
[water] Simplify read/write lowering by relying on MemoryOnlyTypes normal form

### LINKED ISSUE #676: Use normal form to enforce precondition of lowering read/write ops
https://github.com/iree-org/wave/pull/668#discussion_r2658205643

Lowering of read/write ops currently assumes `PropagateElementsPerThread` pass to have run, converting `WaveTensorType` results to `VectorType`. Enforce this using normal form.

tyb0807:
LowerWaveToMLIR requires the MemoryOnlyTypes normal form, which guarantees no WaveTensorType with register address space remains. Since wave.read/write operate on register values, the only valid type after PropagateElementsPerThread has run is VectorType.

This allows us to replace the runtime error handling with assertions, since the normal form precondition is verified at pass entry.

Fixes #676.

### DIFF
--- a/water/lib/Dialect/Wave/Transforms/LowerReadWriteOps.cpp
+++ b/water/lib/Dialect/Wave/Transforms/LowerReadWriteOps.cpp
@@ -378,13 +378,13 @@ public:
   LogicalResult
   matchAndRewrite(wave::ReadOp op, wave::ReadOp::Adaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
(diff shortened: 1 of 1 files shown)


tyb0807:
Yeah I think you are right. Then maybe we can just simplify the check in read/write lowering with this reasoning? Or you think explicit check is still better in this case? I think maybe just keep the check for now, in case PropagateElementsPerThread pass hadn't run beforehand.

ftynse (approved):
Don't forget to update the PR description.

PR #797, opened 2026-02-02
do not wrap root module body by default

### LINKED ISSUE #792: detect normal forms pass should amend the existing normal form module
When we run `--water-wave-detect-normal-forms` on the following IR:

```
module {
  normalform.module {} {
  }
}
```

it will introduce a third nested `normalform.module`. This is not desired. We should have only one. Perhaps a canonicalization to merge the two or some other detection.

tgymnich:
- Wrapping of the root module should not happen, when the module only contains `normalform.modules`.

fixes #792


ftynse (approved):
Thanks!

ftynse:
I'll merge if you don't mind, I need this to make progress

PR #800, opened 2026-02-02
[water] Add wave.broadcast operation definition

### LINKED ISSUE #721: [water] Implement wave.broadcast
part of attention conversion from torch.fx to late Wave dialect

> Let's put some effort and care into designing `broadcast` and `reshape` as these affect shapes and therefore all of the dataflow analyses that rely on them. One particular point that may be worth considering is explicit/implicit broadcasting. Maybe we want to have a pass that makes all broadcasts explicit at some point to simplify analyses and lowerings, for example.

 _Originally posted by @ftynse in [#680](https://github.com/iree-org/wave/issues/680#issuecomment-3750011977)_

ftynse:
Add BroadcastOp to the Wave dialect for broadcasting a tensor to a larger
shape by replicating values along specified dimensions. The operation takes
a source tensor and a broadcast_dims attribute specifying which dimensions
are being added.

Some design decisions:
- broadcast_dims attribute explicitly specifies which dimensions are added
  (source_shape + broadcast_dims = result_shape).
- BroadcastElementsPerThreadOpTrait: EPT propagation depends on broadcast
  dimension. When broadcasting along thread X, EPT is not propagated
  (NoChange) since source has no thread X and result EPT should come from
  downstream users (similar to how PyWave copies index from context).
  For other dims, identity propagation is used.
- IdentityIndexExprsOpTrait: index expressions for shared dims propagate
  bidirectionally, broadcast dims are filled by backward propagation.
- Custom type inference: backward propagation can infer source shape from
  result shape minus broadcast_dims.

Revival of #778 , Fixes #721

### DIFF
--- a/water/include/water/Dialect/Wave/IR/WaveOps.td
+++ b/water/include/water/Dialect/Wave/IR/WaveOps.td
@@ -569,18 +569,16 @@ def BroadcastOp : WaveOp<"broadcast", [
   let summary = "Broadcast a tensor to a larger shape by replicating values";
   let description = [{
     Broadcasts the source tensor to the result shape by replicating values
-    along the broadcast dimensions. The source shape symbols must be a subset
-    of the result shape symbols.
+    along the broadcast dimensions, which must be the trailing dimensions
+    of the result shape.
 
-    The broadcast dimensions are inferred from the difference between the
-    result shape and source shape (result_shape - source_shape).
+    When either operand or result are underspecified, the broadcast dimension
+    must be explicitly provided. Otherwise, it is inferred from the difference
+    between their respective symbolic shapes.
 
-    Example:
-    ```mlir
-    // Broadcast from [@M] to [@M, @N] by adding dimension @N
-    %result = wave.broadcast %src :
-        !wave.tensor<[@M] of f32, <register>> -> !wave.tensor<[@M, @N] of f32, <register>>
-    ```
+    Developer remark: unless one of the dimensions introduced by broadcasting
+    is indexed by a symbol that is mapped to thread X, this operation does not
+    change elements-per-thread.
   }];
 
   let arguments = !con((ins


water/lib/Dialect/Wave/IR/WaveOps.cpp
```diff
@@ -1956,3 +1934,122 @@ LogicalResult wave::ReciprocalOp::verify() {
...
+
+  if (!sourceType || !resultType)
+    return success();
+
+  if (!sourceType.getFullySpecified() || !resultType.getFullySpecified())
+    return success();
+
+  if (getBroadcastDims()) {
+    return emitOpError("does not expect explicit dims when source and result "
+                       "types are fully specified");
```
martin-luecke: I get that we do not want to have the information on which dimensions are broadcast duplicated. Is it possible, that we invalidate this op during type inference, when it starts out with the attribute and not fully specified? The type inference could push the types into fully specified and then the op would fail the verifier because of the duplicated information in the attribute.

water/test/Dialect/Wave/ops.mlir
```diff
@@ -559,3 +559,38 @@ func.func @underspecified_reduction(%input: !wave.tensor<any of f32>, %init: !wa
...
+  %0 = wave.broadcast %arg0 : (!wave.tensor<[@M] of f16, <register>>) -> !wave.tensor<[@M, @N, @K] of f16, <register>>
+  return %0 : !wave.tensor<[@M, @N, @K] of f16, <register>>
+}
+
+// -----
+
+// CHECK-LABEL: @broadcast_explicit_dims
+func.func @broadcast_explicit_dims(%arg0: !wave.tensor<any of f32>) {
+  wave.broadcast %arg0 dims [@K] : (!wave.tensor<any of f32>) -> !wave.tensor<any of f32>
+  return
```
martin-luecke: Check line for broadcast is missing

water/lib/Dialect/Wave/IR/WaveOps.cpp
```diff
@@ -1956,3 +1934,122 @@ LogicalResult wave::ReciprocalOp::verify() {
...
+  for (WaveSymbolAttr sym : sourceType.getShape()) {
+    if (!llvm::is_contained(resultType.getShape(), sym))
+      return emitOpError("source dimension '")
+             << sym.getName() << "' not found in result shape";
+  }
+
+  return success();
+}
+
+llvm::FailureOr<ChangeResult> wave::BroadcastOp::propagateForward(
```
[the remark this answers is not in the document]
ftynse: Now we can

water/include/water/Dialect/Wave/IR/WaveOps.td
```diff
@@ -562,4 +562,48 @@ def CastOp : WaveOp<"cast", [
...
+    DeclareOpInterfaceMethods<WaveElementsPerThreadOpInterface>,
+    WaveInferIndexExprsOpInterface, IdentityIndexExprsOpTrait]> {
+  let summary = "Broadcast a tensor to a larger shape by replicating values";
+  let description = [{
+    Broadcasts the source tensor to the result shape by replicating values
+    along the broadcast dimensions. The source shape symbols must be a subset
+    of the result shape symbols.
+
+    The broadcast dimensions are inferred from the difference between the
+    result shape and source shape (result_shape - source_shape).
```
martin-luecke: Only when broadcast_dims is not supplied. Let's document its semantics here as well

PR #811, opened 2026-02-03
[water] better diagnostics

### LINKED ISSUE #679: Better connect water-opt diagnostics to python during analyses
When running analyses through water, we may capture C++ diagnostics that currently just appear as the "reason" string of an exception. Ideally, we'd have a stack trace in Python pointing to the right locations. Last time I tried, it wasn't possible to programmatically construct a stack trace object or to capture it in a way since it has pointers to interpreter internals. We can, however, print a message that looks similar and has file/line/col info that we have available.

TBD: repro

tgymnich:
- prints MLIR Diagnostics in stack-trace style with the error message and location information.

fixes #679

### DIFF
--- a/lit_tests/kernel/wave/mlir_converter_diagnostics.py
+++ b/lit_tests/kernel/wave/mlir_converter_diagnostics.py
@@ -178,7 +178,7 @@ def test_location_capture_file_line_col():
     # CHECK-LABEL: test_location_capture_file_line_col
     # CHECK: ERROR: test error
     # CHECK: Traceback (Wave DSL source):
-    # CHECK:   File "{{.*}}mlir_converter_diagnostics.py", line 45
+    # CHECK:   File "{{.*}}mlir_converter_diagnostics.py", line 50
     # CHECK:     @wave.wave(constraints)
     # CHECK: location frame count: 1
     # CHECK: first frame type: file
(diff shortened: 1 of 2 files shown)


wave_lang/kernel/wave/mlir_converter/water_emitter.py
```diff
@@ -755,14 +755,59 @@ def _emit_wave_constraints(constraint: Constraint) -> ir.Attribute:
     raise NotImplementedError(f"Unsupported constraint type: {type(constraint)}")
 
 
+def _serialize_location(loc: ir.Location) -> list[dict]:
+    """Convert ir.Location to serializable list of location frames (stack trace).
+
+    Parses the string representation of the location to extract file/line/column
+    information, since direct attribute access may not be available in all bindings.
```
ftynse: What's wrong with direct access? We are getting an `ir.Location` object here, why can't we inspect it?

wave_lang/kernel/wave/mlir_converter/water_emitter.py
```diff
@@ -930,7 +988,12 @@ def _emit_from_captured_trace(
     )
 
     if enable_debug_info and not trace.location:
-        diagnostics.append("Missing debug location for wave trace")
+        diagnostics.append(
+            {
+                "type": "Error",
```
ftynse: I'm not sure how this type is handled

wave_lang/kernel/wave/mlir_converter/mlir_converter.py
```diff
@@ -27,23 +28,151 @@
...
+}
+
+
+def _get_severity_color(severity: str) -> str:
+    """Get ANSI color code based on diagnostic severity."""
+    severity_lower = severity.lower()
+    if "error" in severity_lower:
+        return _COLORS["red"]
+    if "warning" in severity_lower:
+        return _COLORS["yellow"]
```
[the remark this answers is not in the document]
tgymnich: fixed

wave_lang/kernel/wave/mlir_converter/water_emitter.py
tgymnich: All those nanobind object are not pickable.

lit_tests/kernel/wave/mlir_converter_diagnostics.py
```diff
@@ -84,7 +84,6 @@ def compile_and_emit_diagnostics(
     Args:
         location_level: The LocationCaptureLevel to use for capturing locations.
         test_diagnostic_emission: Whether to emit a test diagnostic for verification.
-        use_local_scope: Whether to use local scope naming in MLIR output.
```
ftynse: The argument is still there in the function, just not used.

PR #816, opened 2026-02-03
[WIP] gfx1250 scales load fusion

Hardcode84:


water/include/water/Transforms/Passes.td
```diff
@@ -169,4 +169,21 @@ def WaterMemrefDecompositionPass : Pass<"water-memref-decomposition"> {
...
+    This pass fuses scale loads for `amdgpu.scaled_wmma` operations when the
+    number of scales is less than the wave size. This allows multiple scale
+    values to be loaded together, reducing memory traffic.
+  }];
+  let dependentDialects = [
+    "::mlir::arith::ArithDialect",
+    "::mlir::gpu::GPUDialect",
+  ];
+  let options = [
+    Option<"waveSize", "wave-size", "unsigned", /*default=*/"32",
```
tgymnich: it does not make sense to have a default value here. The wave size should always be set intentionally and correctly, there is no real default that always works.

water/lib/Transforms/FuseScaleLoads.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+// Collect elementwise ops between value and its defining load.
+// Returns the load op and populates elemOps in reverse order (load -> value).
+static LLVM::LoadOp
+collectElementwiseChain(Value value, SmallVectorImpl<Operation *> &elemOps) {
+  while (isElementwise(value.getDefiningOp())) {
+    elemOps.push_back(value.getDefiningOp());
+    value = value.getDefiningOp()->getOperand(0);
+  }
+  std::reverse(elemOps.begin(), elemOps.end());
```
tgymnich:
```suggestion
  llvm::reverse(elemOps);
```

PR #818, opened 2026-02-04
bump llvm to f3df4b9292ce6fbd90de53f0124e55db0a9ee714

ftynse:


Hardcode84:
> This seems to significantly increase the waitcnt number. I have not investigated why specifically, there are some MLIR changes around dataflow but we are not using that currently, are we?

This is most likely from the llvm backend, should be ok, I may look into it later

PR #824, opened 2026-02-04
[Compiler] Filter sync regions with barriers post global optimization

nithinsubbiah:
Currently, filtering of sync regions with barriers between the producer and consumer happens before the global optimization (eg. `minimize_placement_strategy`, `find_disjoint_interval_strategy`). That inhibits the optimization from seeing all hazards thus producing additional barriers in some cases. This PR ensures that the filtering occurs post-optimization. 

Lit test for this already exists in `barrier_strategies.py::test_manual_barriers_prevent_sync_regions`.

Example of why this is better:

Consider three hazard regions A, B, and C with these topological intervals:

Region A: producer@10 → consumer@30
Region B: producer@20 → consumer@60   (has a manual barrier between 20 and 60)
Region C: producer@40 → consumer@60

Old behavior: B is dropped early. The optimizer sees A and C, places barriers at 30 and 60. Total = 3 barriers.
New behavior: The optimizer sees A, B, and C:
1. Region A [10, 30]: place at 30. last_pos = 30.
2. Region B [20, 60]: last_pos=30 is in (20, 60] → covered, skipped.
3. Region C [40, 60]: last_pos=30 is not in (40, 60] → place at 60. last_pos = 60.

Now filter_regions_with_barriers runs on the output (A, C). The optimizer could also have merged B and C since they share the same consumer endpoint. If the manual barrier at position ~55 covers the [20, 60] range, it also covers [40, 60]. In the general case, the optimizer may find that the manual barrier's position already "stabs" region C's interval too — and after filtering, Region C gets removed. Total = 2 barriers instead of 3.

### DIFF
--- a/tests/unittests/test_barrier_strategies.py
+++ b/tests/unittests/test_barrier_strategies.py
@@ -34,27 +34,6 @@ class DummyNode:
         self._topo_location = topo
         self.graph = graph
         self.prev = None
-        self.next = None  # No next node - iteration will stop immediately
-
-    def __ge__(self, other: "DummyNode") -> bool:
-        return self._topo_location >= other._topo_location
-
-    def __le__(self, other: "DummyNode") -> bool:
-        return self._topo_location <= other._topo_location
-
-    def __gt__(self, other: "DummyNode") -> bool:
-        return self._topo_location > other._topo_location
-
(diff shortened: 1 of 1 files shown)


harsh-nod (approved):
lgtm!

PR #832, opened 2026-02-05
[water] make index expr inference handle wave constraints

ftynse:
This was not implemented previously since we didn't have a need.


water/lib/Dialect/Wave/IR/IndexExpr.cpp
```diff
@@ -112,6 +130,27 @@ WaveIndexMappingAttr applyConstraint(WorkgroupConstraintAttr constraint,
                                           context, baseMapping);
 }
 
+WaveIndexMappingAttr applyConstraint(WaveConstraintAttr constraint,
+                                     WaveWorkgroupDim dim,
+                                     uint64_t threadsPerWave,
+                                     WaveIndexMappingAttr baseMapping) {
+  llvm::SmallVector<mlir::Attribute> symbols =
+      llvm::to_vector(constraint.getTileSize().getSymbols());
```
tgymnich:
```suggestion
  llvm::SmallVector<Attribute> symbols(constraint.getTileSize().getSymbols());
```

PR #843, opened 2026-02-06
Add MLIR to FX conversion for Wave dialect roundtrip

martin-luecke:
## Summary

- Add `fx_emitter.py`, the MLIR-to-FX converter that reconstructs a `CapturedTrace` from Wave MLIR text. Runs in a subprocess (like the existing FX-to-MLIR `water_emitter.py`) to isolate the Water MLIR Python bindings. Handles an initial set of Wave operations including `Read`, `Write`, `MMA`, `Iterate`, `Allocate`.
- Add `mlir_to_fx()` entry point in `mlir_converter.py`, making the module a bidirectional converter.
- Add roundtrip tests (`lit_tests/kernel/wave/mlir_to_fx.py`) covering a minimal read/write kernel, a simple matmul, and a pipelined GEMM with software prefetching after full `wave_compile`.
- Expose `iterate_make_isolated` / `iterate_make_non_isolated` through the C API and Python bindings so the FX-to-MLIR emitter can ensure Iterate regions have explicit block arguments for captured values.

## Supporting changes

- Extend graph_utils.py trace comparison to support directional semantics: the roundtripped trace may carry more information (e.g. eagerly populated indices or vector_shapes) than the source, but must not lose any. Factor placeholder reconciliation into dedicated helpers. The FX-to-MLIR emitter calls makeIsolated on Iterate regions, which turns implicitly captured values into explicit block arguments. The MLIR-to-FX converter reconstructs these as "lifted" placeholders. Since the source trace may have transformed these away - depending on the stage in the wave pipeline where conversion was initiated - the two sides can differ in placeholder count and ordering, so the comparison must reconcile them by matching on the outer node they reference rather than by position.
- Extract induction-symbol cleanup into shared utilities in `symbol_utils.py` (`collect_allowed_induction_symbols`, `strip_out_of_scope_induction_symbols`), used by both `water_emitter.py` and `graph_utils.py`.
- Fix `water_emitter.py` to emit 4 index entries for MMA nodes (lhs, rhs, acc, result), matching the C++ `MmaOp::setIndexFromLattices` convention.
- Simplify `Iterate.index` property in `wave_ops.py` to safely handle partially-populated graphs and always return a list.

### DIFF
--- a/wave_lang/kernel/wave/mlir_converter/fx_emitter.py
+++ b/wave_lang/kernel/wave/mlir_converter/fx_emitter.py
@@ -571,22 +571,28 @@ def _handle_allocate_op(op: AllocateOp, parse_ctx: _OpParseContext) -> None:
         raise ValueError("Allocate cannot target Register address space")
     address_space = _address_space_to_symbol(result_type.address_space, parse_ctx)
 
-    distributed_shape = tuple(expr_list_attr_to_exprs(op.distributed_shape))
+    # MLIR stores the *unpadded* distributed shape, padding is a separate
+    # attribute. Reconstruct the padded distributed_shape that the Python
+    # Allocate operation expects, i.e. padding added to the last dimension.
+    unpadded_dist = list(expr_list_attr_to_exprs(op.distributed_shape))
+    padding = int(op.padding) if op.padding is not None else 0
+    tail_padding = int(op.tail_padding) if op.tail_padding is not None else 0
+    if padding and unpadded_dist:
+        unpadded_dist[-1] = unpadded_dist[-1] + padding
+    distributed_shape = tuple(unpadded_dist)
+
     parent_node = (
         parse_ctx.resolve_operand(op.parent) if op.parent is not None else None
     )
     offset = int(op.offset) if op.offset is not None else None
 
-    # distributed_shape and offset are already handled above as constructor
-    # arguments - ignore them so _convert_supported_attrs picks up only the
-    # remaining attrs (e.g. index).
+    # distributed_shape, offset, padding, and tail_padding are handled above
+    # as explicit constructor arguments - ignore them so
+    # _convert_supported_attrs picks up only the remaining attrs.
     converted_attrs = _convert_supported_attrs(
-        op, ignore_attrs={"distributed_shape", "offset"}
+        op,
+        ignore_attrs={"distributed_shape", "offset", "padding", "tail_padding"},
     )
-
-    # TODO: Serialize padding in MLIR AllocateOp.
-    padding = 0
-    tail_padding = 0
 
     allocate_op = Allocate.create(
         parse_ctx.graph,
--- a/wave_lang/kernel/ops/wave_ops.py
+++ b/wave_lang/kernel/ops/wave_ops.py
@@ -1583,10 +1583,10 @@ class Allocate(CustomOp):
     distributed_shape: tuple[IndexExpr]
     dtype: DataType
     address_space: AddressSpace = SHARED_ADDRESS_SPACE
-    padding: int = field(default=0, compare=False)
+    padding: int = 0
     parent: Optional[fx.Node] = None
     offset: Optional[IndexExpr] = None
-    tail_padding: int = field(default=0, compare=False)
+    tail_padding: int = 0
 
     @property
     def indexing_dims(self) -> list[IndexSymbol]:


water/python/WaterExtensionNanobind.cpp
```diff
@@ -876,6 +876,18 @@ NB_MODULE(_waterDialects, m) {
...
+      "iterate_make_isolated",
+      [](MlirOperation op) { mlirWaveIterateOpMakeIsolated(op); },
+      nb::arg("op"),
+      "Makes a wave.iterate operation's region isolated from above.");
+  d.def(
+      "iterate_make_non_isolated",
+      [](MlirOperation op) { mlirWaveIterateOpMakeNonIsolated(op); },
+      nb::arg("op"),
+      "Makes a wave.iterate operation's region non-isolated from above.");
+
```
ftynse: I find it rather gross to expose these at the dialect level, but I know that we are not binding every operation separately. If you don't see an immediate way of attaching these to operation classes, let's create an issue for this and revise later.

wave_lang/kernel/wave/mlir_converter/fx_emitter.py
```diff
@@ -0,0 +1,1115 @@
...
+    op: ir.OpView, ignore_attrs: Optional[set[str]] = None
+) -> dict[
+    str,
+    dict[IndexSymbol, IndexSequence]
+    | dict[IndexSymbol, IndexExpr]
+    | int
+    | MMAType
+    | ScaledMMAType
+    | str,
+]:
```
ftynse: Nit: hoist this into a `TypeAlias` declaration so it can be used below?

wave_lang/kernel/wave/mlir_converter/fx_emitter.py
```diff
@@ -0,0 +1,1115 @@
...
+                    con.set_wave_id_from_hardware_and_workgroup_constraint(hw, wg)
+    return constraints, []
+
+
+def _derive_vector_shapes_from_constraints(
+    constraints: Sequence[Constraint], subs: dict[IndexSymbol, int] | None
+) -> dict[IndexSymbol, int] | None:
+    """Derive vector_shapes from constraint tile sizes for non-MMA kernels.
+
+    This is the fallback used when ``_derive_vector_shapes_from_graph`` found no
```
ftynse: Ultra-nit: I don't know that does the syntax with double backticks mean

wave_lang/kernel/wave/mlir_converter/fx_emitter.py
```diff
@@ -0,0 +1,1115 @@
...
+    Mirrors the core logic of ``get_mma_dimensional_mapping`` (mma_utils.py):
+    for the first MMA node found, determines M, N, K from the operand tensor
+    shapes and maps them to the intrinsic sizes via ``mma_matrix_shapes``.
+
+    Returns ``None`` when the graph contains no MMA nodes (the caller should
+    fall back to ``_derive_vector_shapes_from_constraints``).
+    """
+    for node in trace.walk(lambda n: n):
+        custom = get_custom(node)
+        if not isinstance(custom, MMA):
```
ftynse: Should this also include `ScaledMMA`?

wave_lang/kernel/wave/mlir_converter/fx_emitter.py
```diff
@@ -0,0 +1,1115 @@
...
+
+    def resolve_operand(self, value: ir.Value) -> fx.Node | int | float:
+        """Resolve MLIR value to its corresponding FX node or scalar."""
+        if value not in self.value_map:
+            raise ValueError(f"Missing operand mapping for value {value}")
+        return self.value_map[value]
+
+    def add_mapping(self, mlir_value: ir.Value, fx_node: fx.Node | int | float) -> None:
+        """Add MLIR value → FX node mapping."""
+        self.value_map[mlir_value] = fx_node
```
ftynse: Nit: let's assert for sanity that `mlir_value` isn't already in the map

wave_lang/kernel/wave/mlir_converter/fx_emitter.py
```diff
@@ -0,0 +1,1115 @@
...
+        distributed_shape,
+        dtype,
+        address_space,
+        padding,
+        parent_node,
+        offset,
+        tail_padding,
+    )
+    _apply_mlir_attrs_to_fx_node(allocate_op.fx_node, converted_attrs)
+    parse_ctx.add_mapping(op.result, allocate_op.fx_node)
```
ftynse: I wonder if it makes sense to make a sort of mini-conversion infra here where each `_handle_*` function takes a list of remapped operands, a list of converted result types and returns the list of ignored attributes and the list of results? The caller than takes care of mapping (both lookup and add_mapping), converting result types, and applying attributes. These look boilerplate-y on all operations, but may be the overall complexity is not worth it...

wave_lang/kernel/wave/mlir_converter/fx_emitter.py
```diff
@@ -0,0 +1,1115 @@
...
+
+def _handle_read_op(op: ReadOp, parse_ctx: _OpParseContext) -> None:
+    """Handle wave.read operation."""
+    memory_node = parse_ctx.resolve_operand(op.memory)
+    converted_attrs = _convert_supported_attrs(
+        op,
+        ignore_attrs={
+            AttrNames.OFFSET.mlir_name,
+            AttrNames.SIZE.mlir_name,
+            AttrNames.STRIDE.mlir_name,
```
ftynse: I don't see these attributes on the `ReadOp` definition, but I do see `ordered_syms`.

wave_lang/kernel/wave/mlir_converter/fx_emitter.py
```diff
@@ -0,0 +1,1115 @@
...
+        converted_attrs.get(AttrNames.KIND.mlir_name, None)
+        or parse_ctx.default_mma_type
+    )
+
+    mma_op = MMA.create(
+        parse_ctx.graph,
+        lhs_node,
+        rhs_node,
+        acc_node,
+        *([mma_type] if mma_type is not None else []),
```
ftynse: Why this complexity? Can't we just pass `mma_type` that may be None into the constructor? The type defaults to None anyway: `    mma_type: Optional["MMAType"] | "GenericDot" = None`.

wave_lang/kernel/wave/mlir_converter/mlir_converter.py
```diff
@@ -121,3 +121,65 @@ def emit_wave_dialect(
         [d.decode("utf-8") for d in diagnostics],
         inferred_attributes,
     )
+
+
+def mlir_to_fx(
+    mlir_text: str,
+) -> tuple[CapturedTrace, list[Constraint], WaveCompileOptions, list[str]]:
+    """Convert Wave MLIR text back into a Wave FX trace via subprocess."""
```
ftynse: Nit: please elaborate more on the returned values.

wave_lang/kernel/wave/mlir_converter/mlir_converter.py
```diff
@@ -121,3 +121,65 @@ def emit_wave_dialect(
...
+            diagnostics = (
+                unpickled.get("diagnostics") if isinstance(unpickled, dict) else None
+            )
+        except Exception:
+            diagnostics = None
+        diag_text = f"\nDiagnostics: {diagnostics}" if diagnostics else ""
+        raise RuntimeError(
+            f"fx_emitter failed (code {proc.returncode}):\n"
+            f"{err.decode('utf-8', errors='replace')}{diag_text}"
+        )
```
ftynse: Would it make sense to connect this to https://github.com/iree-org/wave/pull/811?

wave_lang/kernel/wave/utils/graph_utils.py
```diff
@@ -66,29 +72,28 @@ class CheckResult(NamedTuple):
...
 
+# Semantic attributes stored on fx.Node (via setattr) rather than declared as
+# dataclass fields.  These complement the dataclass field introspection in
+# _check_nodes_equivalent: dataclass fields are compared automatically via
+# fields(), while these must be checked explicitly because they live on the
+# fx.Node, not on the CustomOp dataclass.
+#
+# Excluded from comparison (non-semantic / scheduling artifacts):
+#   location, expanded_dims, scheduling_parameters
+_ADDITIONAL_NODE_ATTRS = ("vector_shapes", "reduction_dim", "iter_idx")
```
ftynse: Could you please create a document somewhere and write down these design quirks more prominently?

wave_lang/kernel/wave/utils/graph_utils.py
```diff
@@ -369,48 +384,303 @@ def _check_index_equivalent(
...
+        # The wave to MLIR conversion strips these before emitting (see
+        # strip_out_of_scope_induction_symbols in symbol_utils.py).
+        # Apply the same cleanup to both sides before comparing.
+        if isinstance(lhs_custom, Allocate):
+            if isinstance(lhs_index, dict):
+                allowed = collect_allowed_induction_symbols(lhs_custom.fx_node)
+                lhs_index = strip_out_of_scope_induction_symbols(lhs_index, allowed)
+            if isinstance(rhs_index, dict):
+                allowed = collect_allowed_induction_symbols(rhs_custom.fx_node)
+                rhs_index = strip_out_of_scope_induction_symbols(rhs_index, allowed)
```
ftynse: We shouldn't need this on the RHS since it is supposedly comes from MLIR import where it is impossible

wave_lang/kernel/wave/utils/graph_utils.py
```diff
@@ -369,48 +384,303 @@ def _check_index_equivalent(
...
+) -> tuple[list[fx.Node], list[fx.Node]]:
+    """Reconcile lifted placeholders between two subgraphs.
+
+    "Lifted" placeholders represent captured values from outer scopes.
+    The fx_emitter calls `makeIsolated` on every Iterate node,
+    which ensures captured values become explicit block arguments. They are
+    reconstructed as placeholder nodes in the FX graph.  However,
+    the source FX graph may have had them transformed away (e.g. by hoisting),
+    so the two sides can differ in whether a lifted placeholder exists.
+    This function updates `node_map` so that references resolve correctly
```
ftynse: I don't quite understand this. Why would the source graph had them transformed away to start with? Is there some implicit optimization going on? Furthermore, if they are hoisted from the source graph to its parent, wouldn't that simply mean that graph now has a placeholder too?

wave_lang/kernel/wave/mlir_converter/fx_emitter.py
martin-luecke: Yes, changed to check for `MMABase`

wave_lang/kernel/wave/mlir_converter/fx_emitter.py
```diff
@@ -0,0 +1,1115 @@
...
+        default_mma_type=parse_ctx.default_mma_type,
+    )
+
+    # The YieldOp handler always creates an output node, so we can directly access it
+    output_node = next(node for node in subgraph.nodes if node.op == "output")
+    output_custom = get_custom(output_node)
+    output_vals = output_custom.return_vals[0]
+
+    # Normalize output values to a flat list. The yield handler wraps values in a
+    # tuple for the FX output node, which may result in nested sequences when a
```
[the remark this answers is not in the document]
martin-luecke: simplified 👍

wave_lang/kernel/wave/mlir_converter/water_emitter.py
```diff
@@ -398,19 +298,12 @@ def _attach_attributes(
...
-            if isinstance(parent_custom, Iterate):
-                induction_symbol = get_induction_symbol(parent_custom.axis)
-                allowed_induction_symbols.add(induction_symbol)
+        allowed_induction_symbols = collect_allowed_induction_symbols(node.fx_node)
 
         if isinstance(node, MMA):
-            # Build one index mapping dict per operand for MMA nodes
+            # Build one index mapping dict per operand plus the result for
+            # MMA nodes (lhs, rhs, acc, result), matching the C++ InferTypes
+            # pass which stores operandExprs + resultExprs in that order.
```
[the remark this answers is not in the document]
martin-luecke: The existing index_mapping conversion for MMA here:
https://github.com/iree-org/wave/blob/19e8031a6257c63632cfa2cce2a40a0f6f6122cf/wave_lang/kernel/wave/mlir_converter/mlir_to_wave.py#L313-L315
asserts on four entries to reconstruct the combined mapping

docs/wave/ir_design_notes.rst
```diff
@@ -0,0 +1,89 @@
...
+2. **Dynamic attributes** – set on the `fx.Node` object at runtime via
+   `setattr` (e.g. `node.index`, `node.vector_shapes`,
+   `node.reduction_dim`).  Because they are not dataclass fields, the
+   comparison logic in `graph_utils._check_nodes_equivalent` must
+   enumerate them explicitly (see `_ADDITIONAL_NODE_ATTRS`).
+
+A third category of dynamic attributes is *non-semantic* and deliberately
+excluded from comparison: `location`, `expanded_dims`, and
+`scheduling_parameters`.  These are artefacts of scheduling or debugging
+and do not affect functional equivalence.
```
ftynse: Arguably, the relevant passes should clean those up after they are no longer needed. The irony is that `expanded_dims` is actually load-bearing until #877.

IMO, locations be preserved during roundtrip, but we can defer to a separate PR for this.

ftynse (commented):
Looks good!

I would suggest adding dedicated unit tests for the lifted placeholder reconciliation. We will want to know that the reconciliation itself is correct when we debug graph mismatches (graph isomorphism, NP-complete, yada yada)

ftynse (approved):

PR #848, opened 2026-02-07
[asm] Add region-based control flow and remove labels mode

harsh-nod:
Replace the label-based SCF translation path with structured region-based control flow using waveasm.while and waveasm.if ops.

Key changes:
- WhileOp/IfOp/ConditionOp/YieldOp implement RegionBranchOpInterface for structured loop and conditional control flow
- RegionBuilder translates scf.for -> waveasm.while with proper loop-carried values (induction var + iter_args as block arguments)
- LinearScanPass ties while-op block args, init args, iter args, and results to ensure consistent physical register assignment
- Liveness uses region structure directly for live range extension (no CFG backward dataflow needed)
- AssemblyEmitter lowers WhileOp to label + body + s_cbranch_scc1
- SGPR reservation driven by explicit PrecoloredSRegOp for ABI registers instead of WhileOp-gated blanket reservation
- Waitcnt insertion resets at WhileOp/IfOp/ConditionOp boundaries

Removed:
- Label-based SCF translation path (LoopContext, ControlFlowMode)
- StructuredControlFlow placeholder pass
- CFG/BasicBlock backward dataflow infrastructure
- BranchOp/CondBranchOp structured ops (waveasm.br/waveasm.cond_br)

All 63 e2e tests pass.

### DIFF
--- a/wave_lang/kernel/wave/asm/wave_asm/include/waveasm/Dialect/WaveASMControlFlowOps.td
+++ b/wave_lang/kernel/wave/asm/wave_asm/include/waveasm/Dialect/WaveASMControlFlowOps.td
@@ -1,0 +1,155 @@
+// Copyright 2026 The Wave Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#ifndef WAVEASM_DIALECT_WAVEASMCONTROLFLOWOPS
+#define WAVEASM_DIALECT_WAVEASMCONTROLFLOWOPS
+
+include "mlir/IR/OpBase.td"
+include "mlir/Interfaces/ControlFlowInterfaces.td"
+include "mlir/Interfaces/LoopLikeInterface.td"
+include "mlir/Interfaces/SideEffectInterfaces.td"
+include "waveasm/Dialect/WaveASMDialect.td"
+include "waveasm/Dialect/WaveASMTypes.td"
+
+//===----------------------------------------------------------------------===//
+// Structured Control Flow Operations
+//===----------------------------------------------------------------------===//
+
+def WaveASM_LoopOp : WAVEASMOp<"loop", [
+    DeclareOpInterfaceMethods<RegionBranchOpInterface, ["getSuccessorInputs", "getEntrySuccessorOperands"]>,
+    DeclareOpInterfaceMethods<LoopLikeOpInterface, ["getInitsMutable", "getRegionIterArgs", "getYieldedValuesMutable"]>,
+    RecursiveMemoryEffects,
+    SingleBlockImplicitTerminator<"ConditionOp">
+  ]> {
+  let summary = "Structured loop with trailing condition (do-while semantics)";
+  let description = [{
+    Executes the body region at least once, then re-enters if the trailing
+    waveasm.condition evaluates to true.  This is a do-while style loop:
+    the condition is checked after the body, not before.
+    Uses block arguments for loop-carried values (SSA φ-nodes).
+
+    The body region must terminate with a waveasm.condition operation
+    that computes the loop condition and yields updated values.
+
+    See test/Dialect/region-control-flow.mlir for round-trip examples
+    (simple loops, multi-iter_arg loops, MFMA accumulation, nesting).
+  }];
+
+  let arguments = (ins Variadic<WaveASM_AnyReg>:$initArgs);
+  let results = (outs Variadic<WaveASM_AnyReg>:$results);
+  let regions = (region SizedRegion<1>:$body);
+
+  let skipDefaultBuilders = 1;
+  let builders = [
+    OpBuilder<(ins "::mlir::ValueRange":$initArgs)>
+  ];
+
+  let extraClassDeclaration = [{
+    // Get the loop body block
+    ::mlir::Block &getBodyBlock();
+
+    // Get the number of loop-carried values
+    unsigned getNumRegionIterArgs();
+
+    // RegionBranchOpInterface will be implemented in .cpp
+  }];
+
+  let hasCustomAssemblyFormat = 1;
+  let hasVerifier = 1;
+}
+
+def WaveASM_IfOp : WAVEASMOp<"if", [
+    DeclareOpInterfaceMethods<RegionBranchOpInterface, ["getSuccessorInputs"]>,
+    RecursiveMemoryEffects,
+    SingleBlockImplicitTerminator<"YieldOp">
+  ]> {
+  let summary = "Conditional execution (if-then-else)";
+  let description = [{
+    Conditionally executes then or else region based on condition.
+    Both regions must terminate with waveasm.yield.
+
+    See test/Dialect/region-control-flow.mlir for round-trip examples
+    (simple if, if-then-else, multi-result if, if nested in loop).
+  }];
+
+  let arguments = (ins WaveASM_AnySGPR:$condition);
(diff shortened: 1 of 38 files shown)


wave_lang/kernel/wave/asm/wave_asm/include/waveasm/Dialect/WaveASMControlFlowOps.td
```diff
@@ -0,0 +1,180 @@
...
+include "mlir/Interfaces/LoopLikeInterface.td"
+include "mlir/Interfaces/SideEffectInterfaces.td"
+include "waveasm/Dialect/WaveASMDialect.td"
+include "waveasm/Dialect/WaveASMTypes.td"
+
+//===----------------------------------------------------------------------===//
+// Structured Control Flow Operations
+//===----------------------------------------------------------------------===//
+
+def WaveASM_WhileOp : WAVEASMOp<"while", [
```
willghatch: I think this op is a little weird, and its semantics are a little unclear.  Does the ConditionOp always have to go at the end?  Does in unconditionally execute the first iteration of the loop?  If it does unconditionally execute the first iteration, maybe name it `doWhile` instead of `while`.  Generally, this loop has some aspects of different kinds of loop constructs, maybe make a new name based on the word “loop”, since it is more generic and brings less preconceived baggage of what it is supposed to mean, or make it more like what the name suggests.

willghatch (commented):
I gave this a quick look-over.  As mentioned in a different comment, I find `while` op confusing, I think it should at least have a different name.  Looking at the tests, I was left a little uncertain of what they are trying to test -- eg. some have some waveasm with `while` or other region blocks, and they are just testing that the output has the same overall structure (eg. the output also has `waveasm.while`).  Is the intention that those are just tests that you can parse `waveasm` dialect?

harsh-nod:
> I gave this a quick look-over. As mentioned in a different comment, I find `while` op confusing, I think it should at least have a different name. Looking at the tests, I was left a little uncertain of what they are trying to test -- eg. some have some waveasm with `while` or other region blocks, and they are just testing that the output has the same overall structure (eg. the output also has `waveasm.while`). Is the intention that those are just tests that you can parse `waveasm` dialect?

Yes they were mostly for parsing the dialect but I have added more rigorous checking that should help catch regressions as well.

wave_lang/kernel/wave/asm/wave_asm/include/waveasm/Dialect/WaveASMControlFlowOps.td
```diff
@@ -0,0 +1,182 @@
...
+include "mlir/Interfaces/SideEffectInterfaces.td"
+include "waveasm/Dialect/WaveASMDialect.td"
+include "waveasm/Dialect/WaveASMTypes.td"
+
+//===----------------------------------------------------------------------===//
+// Structured Control Flow Operations
+//===----------------------------------------------------------------------===//
+
+def WaveASM_LoopOp : WAVEASMOp<"loop", [
+    DeclareOpInterfaceMethods<RegionBranchOpInterface, ["getSuccessorRegions", "getSuccessorInputs", "getEntrySuccessorOperands"]>,
```
ftynse: Only methods that have a default implemntation should be listed here, `getSuccessorRegions` doesn't have one so it should be omitted. May be worth double-checking other interfaces below.

wave_lang/kernel/wave/asm/wave_asm/include/waveasm/Dialect/WaveASMControlFlowOps.td
```diff
@@ -0,0 +1,182 @@
...
+    waveasm.condition %cond iter_args(%next_i, %next_acc)
+      : !waveasm.sreg, !waveasm.vreg<4>
+    ```
+  }];
+
+  let arguments = (ins WaveASM_AnySGPR:$condition,
+                       Variadic<AnyType>:$iterArgs);
+
+  let assemblyFormat = [{
+    $condition `:` type($condition) `iter_args` `(` $iterArgs `)`
```
ftynse: What happens when there are no iterArgs?

wave_lang/kernel/wave/asm/wave_asm/include/waveasm/Transforms/Liveness.h
```diff
@@ -168,24 +96,25 @@ struct LivenessInfo {
...
-
-/// Compute local use/def sets for a basic block
-void computeBlockLocalInfo(BasicBlock &block,
-                           llvm::ArrayRef<mlir::Operation *> instructions);
-
-/// Run backward dataflow to compute live_in/live_out sets
-/// Iterates until fixed point is reached
-void computeCFGLiveness(CFG &cfg,
-                        llvm::ArrayRef<mlir::Operation *> instructions);
+LivenessInfo computeLiveness(ProgramOp program);
```
ftynse: Not necessary for this PR, but we should get liveness for free provided correct interface specifications: https://github.com/llvm/llvm-project/blob/b48833bf6619c49ed15390db4c36b217bcbf8a02/mlir/include/mlir/Analysis/DataFlow/LivenessAnalysis.h#L78

ftynse (commented):
I only looked at the op design so far

wave_lang/kernel/wave/asm/wave_asm/include/waveasm/Dialect/WaveASMControlFlowOps.td
```diff
@@ -0,0 +1,182 @@
...
+      %tile = waveasm.ds_read_b64 %iter
+      %new_acc = waveasm.v_mfma_f32_16x16x16_f16 %tile, %tile, %accumulator
+
+      // Update loop counter and check condition
+      %next_i = waveasm.s_add_u32 %iter, %one
+      %cond = waveasm.s_cmp_lt_u32 %next_i, %limit
+
+      waveasm.condition %cond iter_args(%next_i, %new_acc)
+        : !waveasm.sreg, !waveasm.vreg<4>
+    }
```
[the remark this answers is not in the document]
harsh-nod: I have updated them to point to files instead of give examples.

wave_lang/kernel/wave/asm/wave_asm/include/waveasm/Dialect/WaveASMControlFlowOps.td
```diff
@@ -0,0 +1,182 @@
...
+    ```
+  }];
+
+  let arguments = (ins Variadic<AnyType>:$initArgs);
+  let results = (outs Variadic<AnyType>:$results);
+  let regions = (region SizedRegion<1>:$body);
+
+  let skipDefaultBuilders = 1;
+  let builders = [
+    OpBuilder<(ins "::mlir::TypeRange":$resultTypes, "::mlir::ValueRange":$initArgs)>
```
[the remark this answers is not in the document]
harsh-nod: thanks, fixed

wave_lang/kernel/wave/asm/wave_asm/lib/Dialect/WaveASMControlFlowOps.cpp
```diff
@@ -0,0 +1,460 @@
...
+
+  // Verify result types match iter arg types
+  if (getResults().size() != getNumRegionIterArgs()) {
+    return emitOpError("result count must match iter args count");
+  }
+
+  for (auto [resType, argType] :
+       llvm::zip(getResultTypes(), getBodyBlock().getArgumentTypes())) {
+    if (!typesCompatible(resType, argType)) {
+      return emitOpError("result types must match iter arg types");
```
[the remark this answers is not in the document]
harsh-nod: fixed

PR #861, opened 2026-02-09
[water] add select operation and its lowering

### LINKED ISSUE #858: [water] Implement select operation

ftynse:
It is straightdorward as an extension of binary ops, but we are unlikely
to have other ternary ops so not overgeneralizing.

Closes #858.


tgymnich (approved):
LGTM, missing bindings?

PR #862, opened 2026-02-09
MXFP4 GEMM with manual schedule

panditsa:
- 8-wave double buffer MXFP gemm kernel
- 4-wave double buffer MXFP gemm kernel
- Normalized scaled dimensions (like K/2) to base dimensions (like K) for lookup
- Added missing propagate tags

### DIFF
--- a/wave_lang/kernel/wave/templates/tagged_mxfp4_gemm.py
+++ b/wave_lang/kernel/wave/templates/tagged_mxfp4_gemm.py
@@ -109,6 +109,7 @@ def get_tagged_mxfp4_gemm(
         canonicalize=True,
         schedule=SchedulingType.MANUAL,
         use_global_to_shared=True,
+        minimize_shared_allocs=False,
     )
 
     return gemm, options


examples/python/7.1_schedule.py
```diff
@@ -0,0 +1,848 @@
...
+from wave_lang.kernel.wave.constraints import ScaledMMAType
+
+from utils import parse_args, list_tests, run_test
+
+
+# Note this is specified by the HW and cannot be changed.
+SCALE_GROUP_SIZE = 32
+
+
+def generate_gemm_afp4wfp4_inputs(
```
harsh-nod: Can all these functions be moved to a utils folder? So that they can be reused in the tests?

examples/python/7.1_schedule.py
```diff
@@ -0,0 +1,848 @@
+"""
+MXFP GEMM Scheduling: Scaled MMA with MXFP4 Data Type for MI350 (GFX950)
+
+This example demonstrates GFX950 (MI350/CDNA4) scaled MMA GEMM using MXFP4 (f4e2m1fn)
+inputs with f8e8m0fnu scales, with advanced buffering strategies and wave staggering.
+
+Two buffering strategies are demonstrated:
```
panditsa: Fix docstring to reference correct tests

wave_lang/kernel/wave/minimize_global_loads.py
```diff
@@ -98,11 +98,14 @@ def materialize_shape(
             if vector_shapes is None:
                 materialized_shape.append(subs_idxc(dim))
             else:
-                materialized_shape.append(
-                    subs_idxc(
-                        sympy.ceiling(dim / vector_shapes[dim]) * vector_shapes[dim]
+                # Look up vector_shapes using the original dim_expr first, then fall back to base dim
```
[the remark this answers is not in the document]
panditsa: Didn't end up needing these changes, will double check on gfx1250.

wave_lang/kernel/wave/templates/tagged_mxfp4_gemm.py
```diff
@@ -0,0 +1,114 @@
...
+        K: shape[2],
+    }
+    hyperparams.update(get_default_scheduling_params())
+
+    options = WaveCompileOptions(
+        subs=hyperparams,
+        canonicalize=True,
+        schedule=SchedulingType.MANUAL,
+        use_global_to_shared=True,
+    )
```
adedespirlet: I'd suggest adding minimize_shared_allocs=False
This will prevent creating views instead of plain memref allocs for each buffer.  And since we use all buffers at the same time we dont need this optimization

adedespirlet: and it degrades in general the performance as the backend cannot infer aliasing with views

harsh-nod: I would say it depends on the backend :) Not sure if asm backend has the same pitfalls but would be good to compare both options against both backends.

PR #868, opened 2026-02-11
[water] add self index

tgymnich:
- add `self_index` operation to water


water/lib/Dialect/Wave/IR/WaveOps.cpp
```diff
@@ -2013,6 +2013,37 @@ LogicalResult wave::SelectOp::verify() {
...
 }
 
+//-----------------------------------------------------------------------------
+// SelfIndexOp
+//-----------------------------------------------------------------------------
+
+LogicalResult wave::SelfIndexOp::verify() {
+  Type elementType = getElementType(getResult().getType());
+
+  if (!isa<IntegerType>(elementType))
```
ftynse: Nit: should we accept index as well?

PR #869, opened 2026-02-11
[water] add apply_expr

tgymnich:
Introduce an `apply_expr` operation that computes the result of a Wave expression containing symbols and operands. This required introduction of another kind of symbol usable in Wave expressions - WaveOperandAttr. It refers to a positional entity such as the operand of the `apply_expr` operation in an expression. For brevity, introduce ODS attributes constraints to restrict WaveExprListAttr to only accept certain symbols as inputs and use that instead of programmatic verification throughout the codebase.

The lowering of `apply_expr` uses affine expression visitors to directly construct arith and vector dialect operations. It cannot rely on the upstream visitor as that expects `index`-typed operands rather than vectors even though arith natively supports those now.


water/include/water/Dialect/Wave/IR/WaveOps.td
```diff
@@ -637,6 +637,54 @@ def BroadcastOp : WaveOp<"broadcast", [
...
+        : (!wave.tensor<[@M] of i32>, !wave.tensor<[@M] of i32>)
+        -> !wave.tensor<[@M] of i32>
+    ```
+  }] # baseDescription;
+
+  // Accept both WaveTensorType (before PropagateElementsPerThread) and
+  // 1D vectors (after). We cannot use Variadic<WaveTensorInRegister> because
+  // Variadic requires a Type and not a TypeConstraint.
+  let arguments = !con((ins
+    Arg<Variadic<WaveIterableType>, "Input registers">:$arguments,
```
ftynse: This does use `WaveIterableType` though, which I suspect is a workaround

water/python/WaterExtensionNanobind.cpp
```diff
@@ -136,6 +136,39 @@ struct PyWaveIterSymbolAttr
...
+        [](const std::string &symbolName,
+           mlir::python::MLIR_BINDINGS_PYTHON_DOMAIN::DefaultingPyMlirContext
+               context) {
+          MlirStringRef symbolNameStrRef =
+              mlirStringRefCreate(symbolName.data(), symbolName.size());
+          return PyWaveSSAValueAttr(
+              context->getRef(),
+              mlirWaveSSAValueAttrGet(context->get(), symbolNameStrRef));
+        },
+        nb::arg("symbolName"), nb::arg("context") = nb::none(),
```
ftynse:
```suggestion
        nb::arg("symbol_name"), nb::arg("context") = nb::none(),
```

ftynse:
Taking this over

water/include/water/Dialect/Wave/IR/WaveAttrs.td
```diff
@@ -542,9 +558,27 @@ def WaveExprListAttr : AttrDef<WaveDialect, "WaveExprList"> {
...
+
+class WaveExprListAttrOf<list<AttrDef> inputTypes> : Attr<
+  And<[WaveExprListAttr.predicate,
+       WaveExprListAllowedInputsPred<inputTypes>]>,
+  "expression list with "
+  # !interleave(!foreach(entry, inputTypes, entry.cppClassName), ", ")
+  # " inputs"> {
+  let storageType = WaveExprListAttr.cppType;
+  let returnType = WaveExprListAttr.returnType;
+}
```
martin-luecke: nice!

lit_tests/kernel/wave/mlir_converter.py
```diff
@@ -497,6 +497,73 @@ def sum(
...
+    @wave.wave(apply_expr_constraints)
+    def apply_expr_kernel(
+        a: Memory[M, N, ADDRESS_SPACE_A, tkl.i32],
+        c: Memory[M, N, ADDRESS_SPACE_C, tkl.i32],
+    ):
+        a_reg = wave.read(a)
+        c_reg = wave.read(c)
+        result1 = wave.apply_expr(a_reg, lambda x: x + 1)
+        result2 = wave.apply_expr([a_reg, c_reg], lambda x, y: 2 * x + y)
+        result = result1 + result2
```
martin-luecke: Let's also have a`wave.apply_expr`  without operands for sanity checking that that also works

```suggestion
        result1 = wave.apply_expr(a_reg, lambda x: x + 1)
        result2 = wave.apply_expr([a_reg, c_reg], lambda x, y: 2 * x + y)
        result3 = wave.apply_expr([], lambda : 42)
        result = result1 + result2 + result3
```

martin-luecke (approved):
LGTM with small nits on naming and tests

water/include/water/Dialect/Wave/IR/WaveOps.td
```diff
@@ -637,6 +637,54 @@ def BroadcastOp : WaveOp<"broadcast", [
...
+    Arg<WaveExprListAttr, "Expression to apply">:$expr
```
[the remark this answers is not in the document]
ftynse: now #898

PR #890, opened 2026-02-16
Expose compilation passes and add progressive MLIR roundtrip test

martin-luecke:
This extracts `build_graph_passes` from the internal compilation pipeline in `compile.py`, making the ordered list of graph passes available forexternal use.  

Use this to add a new LIT test (`mlir_roundtrip_pipeline.py`) that validates the FX <-> Water MLIR roundtrip after every compilation pass on a matmul kernel.

Each pass is classified as OK, XFAIL (known gap), XPASS (stale xfail), or FAIL (regression).  The test asserts zero FAIL and zero XPASS, so newly fixed passes force us to update the expected set.

## Changes

- Factor out the pass construction from  `_trace_launchable_and_get_kernel_signature`.
- New progressive roundtrip test for matmul with per-pass xfail tracking and stale-entry validation.
- Lift the helpers `error_diagnostics` and `compare_hardware_constraints_for_mlir_roundtrip` to utils files so they are reusable

### DIFF
--- a/tests/kernel/wave_utils_test.py
+++ b/tests/kernel/wave_utils_test.py
@@ -361,9 +361,9 @@ def test_result_type_comparisons(shape1, shape2, dtype1, dtype2, expected_error)
     result = _check_result_types_equivalent(mem1, mem2)
 
     if expected_error:
-        assert not result.success and expected_error in result.reason
+        assert not result and expected_error in result.error
     else:
-        assert result.success
+        assert result
 
 
 @pytest.mark.parametrize(
@@ -380,9 +380,7 @@ def test_payload_comparisons(lhs, rhs, expected_error):
     """Test payload comparison for dtypes, sequences, and dicts."""
     result = _check_payloads_equivalent(lhs, rhs, None, {})
     assert (
-        result.success
-        if not expected_error
-        else not result.success and expected_error in result.reason
+        result if not expected_error else not result and expected_error in result.error
     )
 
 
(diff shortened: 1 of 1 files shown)


lit_tests/kernel/wave/mlir_roundtrip_pipeline.py
```diff
@@ -0,0 +1,248 @@
...
+
+def _try_roundtrip(
+    trace: CapturedTrace,
+    constraints: list[Constraint],
+    options: WaveCompileOptions,
+) -> tuple[bool, str]:
+    """Attempt an MLIR roundtrip on the current trace state.
+
+    Returns `(True, "")` on success, or `(False, reason)` if the
+    roundtrip cannot be performed or produces a non-equivalent trace.
```
ftynse: Nit: I find this tuple structure weird though I have seen it elsewhere, especially that it can represent a malformed state `(True, "but still has a message")`. I'd rather introduce a simple `FailureOr` class which is a wrapper around `str | None` interpreting `None` as success with appropriate contextual casting...

lit_tests/kernel/wave/mlir_roundtrip_pipeline.py
```diff
@@ -0,0 +1,248 @@
...
+        xfail_count = 0
+        xpass_count = 0
+        fail_count = 0
+        xpass_names: list[str] = []
+        fail_names: list[str] = []
+
+        for i, p in enumerate(graph_passes, 1):
+            name = p.__name__
+            expected_fail = name in expected_failures
+            p()
```
ftynse: Not for this commit, but the fact that the pipeline constructor captures a reference to a trace smells. It is difficult to convince oneself that it doesn't change something under the hood, make a copy or somehow accidentally make it easier for tests to be okay.

lit_tests/kernel/wave/mlir_roundtrip_pipeline.py
```diff
@@ -0,0 +1,248 @@
...
+        stale = expected_failures - actual_names
+        assert not stale, (
+            f"expected_failures contains pass names that no longer "
+            f"exist in the pipeline -- remove them: {', '.join(sorted(stale))}"
+        )
+
+        total = len(graph_passes)
+        ok_count = 0
+        xfail_count = 0
+        xpass_count = 0
```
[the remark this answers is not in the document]
martin-luecke: I reused the standard [lit result codes](https://llvm.org/docs/CommandGuide/lit.html#test-status-results). XPASS there means "uneXpected pass"

PR #891, opened 2026-02-17
Add support for dynamic strides

suryajasper:
## Problem
Wave treats all tensor arguments as if they are contiguous. For a non-contiguous input tensor such as `A = torch.randn((M, K * 4))[:, :K]`, while the shape is (M, K), the strides are (K * 4, 1). Since wave runtime only passes a pointer to the linearized tensor data in memory without any information on the strides, the input buffers will be addressed incorrectly, resulting in numerical inaccuracies. 

## Solution
This PR provides universal support for non-contiguous input & output tensor layouts through wave runtime. Wave automatically passes one stride per dimension per buffer through the pipeline and uses them in the kernel IR so loads/stores use the correct layout.

- Provide extra index arguments for strides
- Emitter builds memref.reinterpret_cast with those stride values and a layout like strided<[?, 1]> (dynamic leading strides, unit stride on the last dim to satisfy vector.load).
- During runtime invocation, Wave parses the tensor metadata to pass in the appropriate strides for each dimension; We extend the kernarg buffer in the C++ runtime with a strides section after pointers, scalars, and dynamic dims.

## Validation
Added new test: `dynamic_strides_test.py`

### DIFF
--- a/wave_lang/kernel/wave/asm/waveasm_e2e.py
+++ b/wave_lang/kernel/wave/asm/waveasm_e2e.py
@@ -38,6 +38,7 @@ from typing import Optional, List, Tuple
 import torch
 
 from wave_lang.kernel.wave.asm.utils import extract_func_from_stream_mlir
+from wave_lang.kernel.wave.utils.run_utils import get_dynamic_stride_args
 from wave_lang.kernel.wave.utils.classes import Failure, Result, Success
 
 
@@ -664,8 +665,11 @@ def run_with_wave_runtime(
     kern_args = [tensor.data_ptr() for tensor in all_tensors]
     kernel_args = wave_runtime.Int64Vector(kern_args)
 
+    # Prepare dynamic stride arguments
+    stride_args = get_dynamic_stride_args(all_tensors)
+
     # Launch
-    wave_runtime.launch(kernel_launch_info, kernel_args, [], [])
+    wave_runtime.launch(kernel_launch_info, kernel_args, [], [], stride_args)
 
     # Sync
     torch.cuda.synchronize()
--- a/wave_lang/kernel/wave/compile_options.py
+++ b/wave_lang/kernel/wave/compile_options.py
@@ -114,7 +114,7 @@ class WaveCompileOptions:
     # Enable dynamic strides through Wave runtime and LLVM backend
     @property
     def dynamic_strides(self) -> bool:
-        return self.wave_runtime and self.backend != "asm"
+        return self.wave_runtime and self.backend == "llvm"
 
     # === Print options ===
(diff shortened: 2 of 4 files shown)


tests/kernel/asm_backend_test.py
```diff
@@ -54,6 +54,7 @@ def _mma_type_params():
 
 @require_e2e
 @require_cdna_3_or_4
+@pytest.mark.xfail(reason="Dynamic strides are not supported in the ASM backend yet")
```
raikonenfnu: Can we set dynamic_stride=False instead of xfailing?

wave_lang/kernel/compiler/wave_codegen/emitter.py
```diff
@@ -152,10 +152,21 @@ def abi_type(binding: BindingDesc):
...
         arg_types = [abi_type(b) for b in bindings]
 
+        stride_arg_count = 0
+        if self.options.wave_runtime:
+            stride_arg_count = sum(
+                len(b.kernel_buffer_type.symbolic_shape)
+                for b in self.root_sig.sig.kernel_buffer_bindings
+            )
+            if stride_arg_count > 0:
+                arg_types += [IndexType.get()] * stride_arg_count
```
raikonenfnu: does dynamic stride not work with non wave_runtime?

wave_lang/kernel/compiler/wave_codegen/emitter.py
```diff
@@ -192,36 +204,84 @@ def get_static_dim(s: Optional[IndexExpr]) -> int:
...
+                        entry_block.arguments[
+                            stride_arg_offset : stride_arg_offset + rank
+                        ]
+                    )
+                    stride_arg_offset += rank
+                    dyn_stride_sentinel = MemRefType.get_dynamic_stride_or_offset()
+                    if rank == 1:
+                        stride_vals = [arith_d.constant(IndexType.get(), 1)]
+                        static_strides = [1]
+                        layout = StridedLayoutAttr.get(offset=dyn_val, strides=[1])
```
raikonenfnu: Does the latter case not work when rank==1?

wave_lang/kernel/wave/compile.py
```diff
@@ -309,7 +309,14 @@ def get_dynamic_dimension_actual(sym):
             )
         else:
             tensors = [t.data for t in chain(kernel_inputs, kernel_outputs)]
-            self.launchable(*tensors, *scalar_args)
+            if self.options.wave_runtime and self.options.backend != "asm":
```
harsh-nod: This check seems to be present everywhere in the code. Could you pull it out as a separate function that can be utilized throughout the PR?

PR #897, opened 2026-02-17
enable scaled_mma opsel

willghatch:
IE rather than extracting an i8 (and bitcasting to f8e8m0fnu) from a 4x vector (IE one VGPR) to use as mfma argument, use the extraction offset as the opsel argument for the mfma.  This should reduce register pressure.

With the shuffle, the vector load, and the opsel, this kernel has about a 23.4% speedup over the vanilla kernel (in the same example file) based on some quick benchmarking.

Depends on https://github.com/iree-org/wave/pull/892


willghatch:
But also I was lazy and just looked at the code and benchmark of the example file without running tests earlier when I did this, and it looks like this breaks a bunch of tests that I'll need to fix.

PR #899, opened 2026-02-18
[water] Allow sparse bounds dictionaries in read/write wave MLIR ops

martin-luecke:
Relax the WaveReadWriteBoundsAttr verifier to accept sparse bounds dictionaries. The Python pipeline only stores bounds for dimensions that actually need out-of-bounds masking with downstream passes relying on this sparsity. Previously, the MLIR verifier required an entry for every tensor dimension, causing verification failures during emission.

- Remove the "all dimensions required" check from the verifier.
- Update `buildMask` in lowering to skip dimensions without a bounds entry instead of asserting.
- Add lowering and round-trip tests for sparse bounds.
- Remove `generate_bounds_exprs`, `merge_contiguous_reads`, and `location_check_pass` from the roundtrip `expected_failures` set.

### DIFF
--- a/water/include/water/Dialect/Wave/IR/WaveAttrs.td
+++ b/water/include/water/Dialect/Wave/IR/WaveAttrs.td
@@ -603,11 +603,6 @@ def WaveReadWriteBoundsAttr : AttrDef<WaveDialect, "WaveReadWriteBounds"> {
     (strings) to a WaveExprListAttr specifying the bounds of the read/write
     operations this is attached to. The dictionary may be sparse: only
     dimensions that require masking need an entry.
-
-    Example:
-    ```
-    #wave.read_write_bounds<{M = #wave.expr_list<[#wave.symbol<"M">] -> (M)>}>
-    ```
   }];
 
   let parameters = (ins "::mlir::DictionaryAttr":$mapping);


water/include/water/Dialect/Wave/IR/WaveAttrs.td
```diff
@@ -575,37 +542,20 @@ def WaveExprListAttr : AttrDef<WaveDialect, "WaveExprList"> {
...
+    This attribute contains a dictionary mapping from symbolic dimensions
+    (strings) to a WaveExprListAttr specifying the bounds of the read/write
+    operations this is attached to. The dictionary may be sparse: only
+    dimensions that require masking need an entry.
 
     Example:
     ```
-    #wave.read_write_bounds<{M = #wave.expr_list<[BLOCK_M] -> BLOCK_M * 2>}>
+    #wave.read_write_bounds<{M = #wave.expr_list<[#wave.symbol<"M">] -> (M)>}>
     ```
```
ftynse: This is why I usually advise against putting examples in ODS: they bitrot faster than you blink. I had seen PRs where examples were already incorrect :)  We can point to the test file for syntax example, but the assembly format below (which is normally included in the generated .md) should be descriptive enough

ftynse (approved):
Please update the PR name and description, unfortunately github doesn't when the branch is force-pushed and it will take the PR description for the squashed commit.

PR #908, opened 2026-02-18
NFC: remove commented out code and rewrap

ftynse:


ftynse:
Reformatting comments cannot affect tests in a sane universe

PR #909, opened 2026-02-18
[asm] Fix critical correctness bugs and add missing op handlers

harsh-nod:
Consolidated correctness fixes from the reduce_reg_pressure branch, excluding AGPR-related changes (not yet upstream).

Tier 1 -- Critical correctness fixes:

1. Fix tied-register double-allocation at loop boundaries: when a tied value (e.g., MFMA accumulator loop result) was processed after its partner expired from the active list, the physical register was never re-reserved, allowing silent data corruption. Also extend block arg live ranges to bridge the gap to loop result def points, preventing pool fragmentation. (from 8cf5e21c)

2. Fix waitcnt double-free crash: combineAdjacentWaitcnts could erase the same operation twice when consecutive combine pairs shared an op. Also replace program.walk() with pre-collected op vector iteration to avoid iterator invalidation on large kernels. (from 2152583)

Tier 2 -- High-impact correctness + performance fixes:

3. Reduce s_waitcnt count from 51 to 4 in kernel loop body: fix incremental wait tracking (don't reset on new loads), correct barrier outstanding detection, remove ConditionOp from resetWaits (loop back-edge), add LGKM latency-covering heuristic (skip waits with 40+ intervening ops), and combine adjacent single-counter waits. (from de5ba25)

4. Add missing MLIR op handlers that were silently dropped: amdgpu.memory_counter_wait -> s_waitcnt vmcnt(N), rocdl.s_barrier -> s_barrier, rocdl.s_setprio -> s_setprio N. Fixes barrier count (2 -> 4), adds wave priority (0 -> 8 s_setprio instructions). (from 93cad74)

5. Fix LDS offset tracking for direct memref.alloc: MXFP4 kernels use typed shaped allocations (memref<256x128xi8, 3>) instead of memref.view. Add cumulative ldsAllocOffset counter so each alloc gets a distinct byte offset. (from de2f312)

6. Fix SALU/VOP2 literal handling: SALU instructions support 32-bit literals natively (don't materialize into VGPR scratch). VOP2 instructions require literal in src0 -- swap commutative ops, fall back to materialization for non-commutative. (from 53b5b9d)

7. Fix VALU literal operand handling: all v_* instructions (except v_mov_b32) correctly materialize out-of-range literals. Fix v_cmp_* to use VOP3 encoding (_e64 suffix) with explicit VCC destination. (from 2196f9c1)

Also adds --max-vgprs/--max-sgprs CLI options, cap waitcnt values to hardware limits (lgkmcnt<=15, vmcnt<=63), and add lit tests for LDS offset tracking, scaled MFMA pipelining, and double-buffer assembly.

### DIFF
--- a/wave_lang/kernel/wave/asm/wave_asm/include/waveasm/Dialect/WaveASMOps.td
+++ b/wave_lang/kernel/wave/asm/wave_asm/include/waveasm/Dialect/WaveASMOps.td
@@ -998,8 +998,21 @@ def WaveASM_S_BARRIER : WAVEASMOp<"s_barrier", [WaveASM_ControlFlowOp]> {
   let assemblyFormat = "attr-dict";
 }
 
-def WaveASM_S_NOP : WaveASM_CountedWaitOp<"s_nop"> {
+// Ops that take a single integer operand but are not memory waits.
+// Structurally identical to CountedWaitOp (same arguments and format)
+// but semantically distinct: these are scalar control ops, not waitcnts.
+class WaveASM_CountedScalarOp<string mnemonic> :
+  WAVEASMOp<mnemonic, [WaveASM_ControlFlowOp]> {
+  let arguments = (ins I32Attr:$count);
+  let assemblyFormat = "$count attr-dict";
+}
+
+def WaveASM_S_NOP : WaveASM_CountedScalarOp<"s_nop"> {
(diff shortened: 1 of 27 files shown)


panditsa (approved):
Overall LGTM! Just one concern with the handlers_memory.py change

PR #912, opened 2026-02-18
Expand 7.2 example to have double buffer scheduling and scale preshuffle

willghatch:
Previously 7.2 just did scale preshuffle, and didn't build on top of the optimiziations of 7.1. This adds a scheduled preshuffle kernel to 7.2 like in 7.1.  However, the exact schedule is not the same, because 7.2 does not use LDS memory for the scales, since optimization to merge loads into vector loads is not yet working for LDS.  So this includes a new schedule. It has a `--benchmark` flag that can be used to benchmark and see results of vanilla mxfp4, scheduled, preshuffled (non-scheduled), and preshuffled scheduled.

### DIFF
--- a/examples/python/7.2_mxfp4_gemm_preshuffle_scale.py
+++ b/examples/python/7.2_mxfp4_gemm_preshuffle_scale.py
@@ -1,48 +1,61 @@
 """
-MXFP4 GEMM Example: Unshuffled vs Shuffled Scales
+MXFP4 GEMM with Pre-shuffled Scales - Scheduled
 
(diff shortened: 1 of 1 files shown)


xintin:
Just a thought, I think we shall keep 7.2, just scale preshuffle. Benchmarking script could be a separate one.
This will align with the current file structure

willghatch:
I'm fine with moving out the benchmark, it's just been convenient when working on it.  But do you mean that you want the preshuffle scale + scheduling in a different file, too?  Or leave it as it is here except to remove the benchmarking?

PR #914, opened 2026-02-19
MXFP4 Workgroup Reordering Based on SP3 Kernel

suryajasper:
# MXFP4 Workgroup Reordering
Remaps the workgroup grid by creating groups of 32 tiles along the N dimension. This way, the first 32 workgroups in launch order are assigned along the N dimension on the same M row (0), then the next 32 to the next M row (1), and so on until wrapping around back to the first row. 

Through this grouping, consecutive workgroups are able to reuse the same A row slices rather than spreading over a larger, sparse footprint. This improves L2 cache reuse since workgroups running close together in launch order work on the same row of output tiles. 

The index remapping is calculated through two cases:
- **Main case**: Creates full groups of 32 consecutive tiles along N dimension
- **Tail case**: If the number of tiles along the N dimension is not a multiple of our group size (32), after forming as many full groups as possible, we are left with a small rectangle in the launch grid, which we fill with the tailing workgroups in row-major order.

### DIFF
--- a/wave_lang/kernel/wave/templates/tagged_mxfp4_gemm.py
+++ b/wave_lang/kernel/wave/templates/tagged_mxfp4_gemm.py
@@ -17,7 +17,7 @@ Required tags: k_loop, read_a, read_a_scale, read_b, read_b_scale,
 bitcast_a, bitcast_a_scale, bitcast_b, bitcast_b_scale, scaled_mma.
 """
 
-from sympy import Piecewise, ceiling, floor
+from sympy import Piecewise, ceiling, floor, Max
 
 import wave_lang.kernel.lang as tkl
 import wave_lang.kernel.wave as tkw
@@ -57,7 +57,6 @@ def get_tagged_mxfp4_gemm(
     BLOCK_M = tkl.sym.BLOCK_M
     BLOCK_N = tkl.sym.BLOCK_N
     BLOCK_K = tkl.sym.BLOCK_K
-    GROUP_SIZE_N = tkl.sym.GROUP_SIZE_N
     A_ADDRESS_SPACE = tkl.sym.A_ADDRESS_SPACE
     B_ADDRESS_SPACE = tkl.sym.B_ADDRESS_SPACE
     C_ADDRESS_SPACE = tkl.sym.C_ADDRESS_SPACE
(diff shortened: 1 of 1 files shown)


wave_lang/kernel/wave/templates/tagged_mxfp4_gemm.py
```diff
@@ -287,3 +311,53 @@ def repeat(
...
+    main_new_wg1 = (
+        group_index // num_wg_0
+    ) * group_size_n + flat_wg_index % group_size_n
+
+    # Tailing case, when N tiles is not a multiple of GROUP_SIZE_N
+    full_tiles_n = floor(num_wg_1 / group_size_n) * group_size_n
+    tail_tiles_n = num_wg_1 - full_tiles_n
+    total_full = full_tiles_n * num_wg_0
+    tail_linear = flat_wg_index - total_full
+    tail_new_wg0 = tail_linear // tail_tiles_n
```
harsh-nod: This will fail if tail_tiles_n = 0. Will tail_tailes_n every be 0 and if so, should be instead divide by Max(tail_tiles_n, 1)?

wave_lang/kernel/wave/templates/tagged_mxfp4_gemm.py
```diff
@@ -53,6 +57,7 @@ def get_tagged_mxfp4_gemm(
     BLOCK_M = tkl.sym.BLOCK_M
     BLOCK_N = tkl.sym.BLOCK_N
     BLOCK_K = tkl.sym.BLOCK_K
+    GROUP_SIZE_N = tkl.sym.GROUP_SIZE_N
```
harsh-nod: If group_size_n is just a fixed constant, we dont need to create a new symbolic variable for it.

wave_lang/kernel/wave/schedules/gemm_mxfp4_double_buffer.py
```diff
@@ -259,7 +259,7 @@ def get_mxfp4_dbuf_pingpong_schedule(use_stagger: bool = True, shape: tuple = No
         use_stagger: Enable wave staggering + WorkgroupBarrier in cluster 0.
             Recommended for 8-wave configs; disable for 4-wave.
         shape: Tuple of (M, N, K) dimensions. If provided and bigger than
-            (1024, 1024, 16384), an extra WorkgroupBarrier will be added
+            (1024, 1024, 1024), an extra WorkgroupBarrier will be added
```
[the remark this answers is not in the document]
harsh-nod: Thats fine for now, but what was the reason for the change? And its okay to land minor targeted fixes, (in fact even encouraged :))

PR #930, opened 2026-02-19
[water] replace ReadWriteBoundsAttr with SymbolMappingAttr, NFC

ftynse:
This removes one level of indirection when accessing the bounds attribute and another level of indirection in using context-uniqued WaveSymbolAttr instead of raw strings.

Partially supersedes #730

### DIFF
--- a/water/lib/Dialect/Wave/Transforms/LowerReadWriteOps.cpp
+++ b/water/lib/Dialect/Wave/Transforms/LowerReadWriteOps.cpp
@@ -91,7 +91,7 @@ buildStartIndices(Location loc, DictionaryAttr indexDict,
 ///                  bound_d(elements_per_thread))
 ///          foreach d in dimensions.
 ///
-/// whenever a bounds dictionary is provided. When it is not provided, return a
+/// whenever a bounds mapping is provided. When it is not provided, return a
 /// null mask. If the vectorized dimension cannot be identified, return failure.
 static FailureOr<Value>
 buildMask(Location loc, wave::WaveSymbolMappingAttr boundsMapping,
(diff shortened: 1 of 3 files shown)


water/lib/Dialect/Wave/Transforms/LowerReadWriteOps.cpp
```diff
@@ -309,7 +307,7 @@ createMemoryIndicesAndMask(ConversionPatternRewriter &rewriter,
                            Type memoryTypeArg, VectorType vectorType) {
   int64_t elementsPerThread = vectorType.getNumElements();
 
-  wave::WaveReadWriteBoundsAttr boundsDict = op.getBoundsAttr();
+  wave::WaveSymbolMappingAttr boundsDict = op.getBoundsAttr();
```
martin-luecke: nit: local variable is still called `boundsDict` instead of `boundsMapping`

water/python/WaterExtensionNanobind.cpp
```diff
@@ -575,47 +576,79 @@ struct PyWaveReadWriteBoundsAttr
...
+      MlirAttribute value = mlirWaveSymbolMappingAttrLookup(self, keyAttr);
+      if (mlirAttributeIsNull(value)) {
+        throw nb::key_error("Key not found.");
+      }
+      return value;
+    });
+    c.def("__getitem__", [](MlirAttribute self, intptr_t index) {
+      if (index < 0 || index >= mlirWaveSymbolMappingAttrGetNumEntries(self)) {
+        throw nb::index_error("Index out of range.");
+      }
```
martin-luecke: I see now, this is actually code that is used in the fx_emitter:
https://github.com/iree-org/wave/blob/42b512635c1bb5319e1a268c3de91fbed0a17a59/wave_lang/kernel/wave/mlir_converter/fx_emitter.py#L202-L215
because in the absence of `__iter__` on this, Python falls back on __get_item__ with int. We could also consider explicitly implementing `__iter__`, but I am fine with this as well

PR #935, opened 2026-02-19
Add MXFP4 GEMM template with preshuffled B data and B scales

panditsa:
Adds `get_tagged_mxfp4_gemm_preshuffle_b`, a new tagged MXFP4 GEMM kernel template where B data is read directly from global memory using the `shuffle_weight` preshuffle mapping and B scales use the `e8m0_shuffle` preshuffle mapping, while A and A scales continue through LDS. The `b_preshuffle` and `e8m0_shuffle` host-side tensor transformation utilities are added to `mxfp_utils.py` as shared helpers. An end-to-end pytest (`testScaledGemmMXFP4PreshuffleB`) and an example test in `7.1_schedule.py` are included to validate correctness against the PyTorch reference.


harsh-nod (approved):
lgtm!

PR #936, opened 2026-02-20
[water] add reshape operation

ftynse:
This is an "internal" operation that is introduced to transpose
register-stored data in conjunction with `extract`. The design attempts
to mirror that of the python equivalent, which appears inconsistent
if not entirely broken (#942).

Type verification is intentionally lax on types because the operation
leaks the vector concepts into tensor land and there is no good way of
reconciling those without a redesign.

### DIFF
--- a/water/lib/Dialect/Wave/Transforms/LoweringPatterns.cpp
+++ b/water/lib/Dialect/Wave/Transforms/LoweringPatterns.cpp
@@ -1172,6 +1172,7 @@ public:
       return success();
     }
 
+    // Split case: extract a slice from a single operand.
     auto targetType = cast<VectorType>(
         getTypeConverter()->convertType(op.getResult().getType()));
     assert(static_cast<uint64_t>(targetType.getNumElements()) ==
--- a/wave_lang/kernel/wave/mlir_converter/water_emitter.py
+++ b/wave_lang/kernel/wave/mlir_converter/water_emitter.py
@@ -496,7 +496,7 @@ def _convert_vector_shapes(
     """Converts a dictionary of index symbols mapped to integers to a dictionary attribute."""
     resolved = _resolve_vector_shapes_for_attr(shapes, subs)
     i64 = ir.IntegerType.get_signless(64)
-    dict = {k.name: ir.IntegerAttr.get(i64, v) for k, v in resolved.items()}
+    dict = {k: ir.IntegerAttr.get(i64, v) for k, v in resolved.items()}
     return ir.DictAttr.get(dict)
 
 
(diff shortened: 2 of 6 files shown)


wave_lang/kernel/wave/mlir_converter/water_emitter.py
```diff
@@ -487,6 +490,16 @@ def _convert_to_wave_expr_list_tuple(
     return WaveExprListAttr.get(symbol_attrs, multi_result_map)
 
 
+def _convert_vector_shapes(
+    shapes: dict[IndexSymbol, int | IndexExpr], subs: dict[IndexExpr, Any] = {}
+) -> ir.DictAttr:
+    """Converts a dictionary of index symbols mapped to integers to a dictionary attribute."""
+    resolved = _resolve_vector_shapes_for_attr(shapes, subs)
+    i64 = ir.IntegerType.get_signless(64)
+    dict = {k.name: ir.IntegerAttr.get(i64, v) for k, v in resolved.items()}
```
martin-luecke: This is code motion, I think. But previously this was used like this:
`dict = {k: ir.IntegerAttr.get(i64, v) for k, v in resolved.items()}`.
`_resolve_vector_shapes_for_attr` returns `dict[str, int]`, so `.name` will be an issue

wave_lang/kernel/wave/mlir_converter/water_emitter.py
martin-luecke: Handling the 1-elem case as a sequence as well reduces duplicated code:
```Python
    mlir_operands = []
    for arg in fx_node.args:
        items = arg if isinstance(arg, Sequence) else [arg]
        for item in items:
            if (mlir_arg := get_single_mapped_value(item, allow_missing=True)) is not None:
                mlir_operands.append(mlir_arg)
    return mlir_operands
```

wave_lang/kernel/wave/mlir_converter/water_emitter.py
ftynse: Thanks!

water/lib/Dialect/Wave/Transforms/LoweringPatterns.cpp
```diff
@@ -1108,6 +1108,92 @@ class ShuffleOpLoweringPattern : public OpConversionPattern<wave::ShuffleOp> {
...
+    assert(static_cast<uint64_t>(targetType.getNumElements()) ==
+               static_cast<uint64_t>(
+                   cast<VectorType>(adaptor.getSource()[0].getType())
+                       .getNumElements()) /
+                   op.getNumSlices() &&
+           "vector size mismatch");
+    if ((cast<VectorType>(adaptor.getSource()[0].getType()).getNumElements() %
+         op.getNumSlices()) != 0) {
+      return rewriter.notifyMatchFailure(op,
+                                         "imperfectly divisible vector size");
```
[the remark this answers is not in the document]
ftynse: It will hide a problem: if targetNumElements = (numSlices + 1) * sliceNumElements + leftover, it will gracefully exit on non-divisible case even though there is a problem with sizes that should have been caught by verifiers.

PR #937, opened 2026-02-20
[water] tell cursor where the binaries are

ftynse:
It seems to be constantly confused otherwise.


martin-luecke (approved):

ftynse:
NFC, not waiting for CI

PR #955, opened 2026-02-23
Buffer load strength reduction

Hardcode84:
### Summary

- Add an explicit `soffset` (SGPR scalar offset) operand to all VMEM load ops in the WaveASM dialect
- Add a new `BufferLoadStrengthReduction` pass that eliminates per-iteration VALU address computation in loops by exploiting the hardware soffset field

### Motivation

In a typical GEMM inner loop, each `buffer_load` recomputes its voffset from the induction variable every iteration (e.g. `v_mov + v_add + v_lshlrev` = 3 VALU ops per load). With N loads sharing the same SRD, that's 3N VALU instructions per iteration just for address math.

Buffer instructions have `effective_address = SRD_base + voffset + soffset`, where `soffset` is an SGPR-only scalar offset added by the hardware at zero VALU cost. The pass exploits this by:

1. Precomputing the voffset at `iv=initial_value` (loop-invariant, captures the thread-specific part)
2. Computing the per-iteration stride via `v_readfirstlane(voff(iv+step) - voff(iv))`
3. Carrying one SGPR `soffset` per SRD group as a loop iter_arg, bumped by `s_add_u32` each iteration

This replaces **N VALU ops/iter per SRD group** with **1 SALU op/iter per SRD group**.

### Why buffer-load-specific (not a general transform + peephole)

A general loop strength reduction would produce `v_add(base, accumulated_stride)` -- still 1 VALU per iteration -- which a subsequent peephole could split into voffset/soffset for buffer_loads. But the only consumers of IV-dependent address chains in GEMM loops are buffer_loads and LDS ops (the latter handled by `LoopAddressPromotion` with a different strategy). Two passes for one consumer class isn't worth the abstraction boundary.

### Changes

**Dialect** (`WaveASMOps.td`): Added `WaveASM_SRegOrImm:$soffset` as 3rd operand to `VMEMLoadOp` class.

**Translation** (`TranslateFromMLIR.cpp`, `AMDGPUHandlers.cpp`, `MemRefHandlers.cpp`): All `BUFFER_LOAD_*::create()` calls now pass a zero constant as the soffset operand.

**Assembly emission** (`AssemblyEmitter.cpp`): `emitBufferLoad` resolves operand(2) as soffset and emits it instead of a hardcoded `0`.

**New pass** (`BufferLoadStrengthReduction.cpp`): Finds buffer_loads in `LoopOp` with IV-dependent voffsets, groups by SRD, precomputes stride, and rewrites the loop with soffset iter_args.

**Pipeline** (`waveasm-translate.cpp`): Pass runs unconditionally after peephole, before memory offset optimization.

### DIFF
--- a/wave_lang/kernel/wave/asm/wave_asm/lib/Transforms/BufferLoadStrengthReduction.cpp
+++ b/wave_lang/kernel/wave/asm/wave_asm/lib/Transforms/BufferLoadStrengthReduction.cpp
@@ -463,10 +463,12 @@ static void applyStrengthReduction(LoopOp loopOp) {
 
   // Patch buffer_loads: set voffset to precomputed value, soffset to iter_arg.
   for (unsigned i = 0; i < candidates.size(); ++i) {
-    // Find the cloned buffer_load.
+    // Find the cloned buffer_load via the mapping. The lookup must succeed
+    // because we cloned the entire loop body above.
     Operation *clonedLoad = nullptr;
     for (Value result : candidates[i].loadOp->getResults()) {
-      Value clonedResult = mapping.lookup(result);
+      Value clonedResult = mapping.lookupOrNull(result);
+      assert(clonedResult && "cloned load result not found in mapping");
       clonedLoad = clonedResult.getDefiningOp();
       break;
     }
@@ -487,6 +489,9 @@ static void applyStrengthReduction(LoopOp loopOp) {
   // and s_cmp -> s_cbranch must have no SCC-clobbering ops between them).
   // Find the cloned s_cmp/s_add for IV (the ops that produce the condition)
   // and insert soffset updates before them.
+  // NOTE: s_add_u32 wraps at 2^32. For typical GEMM loops (stride < 2^20,
+  // iterations < 2^12), soffset stays well within 32 bits. If the product
+  // stride*iterations overflows 2^32, the buffer address will be wrong.
   Value newCond = mapping.lookup(condOp.getCondition());
 
   // Insert soffset updates before the s_cmp that produces the condition.
--- a/wave_lang/kernel/wave/asm/wave_asm/test/Transforms/buffer-load-strength-reduction.mlir
+++ b/wave_lang/kernel/wave/asm/wave_asm/test/Transforms/buffer-load-strength-reduction.mlir
@@ -167,10 +167,13 @@ waveasm.program @shared_srd_group
   %init_iv = waveasm.s_mov_b32 %zero : !waveasm.imm<0> -> !waveasm.sreg
   %init_acc = waveasm.v_mov_b32 %zero : !waveasm.imm<0> -> !waveasm.vreg
 
-  // One SRD group (same stride) -> 1 extra soffset iter_arg.
+  // One SRD group (same stride=16) -> 1 extra soffset iter_arg.
   // Original: 2 iter_args. After: 3 (iv, acc, soff).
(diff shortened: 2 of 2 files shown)


wave_lang/kernel/wave/asm/wave_asm/lib/Transforms/BufferLoadStrengthReduction.cpp
```diff
@@ -0,0 +1,421 @@
...
+    Value strideSGPR; // stride as SGPR (from v_readfirstlane).
+  };
+  llvm::DenseMap<Value, unsigned> srdToGroupIdx;
+  SmallVector<SRDGroup> groups;
+  SmallVector<unsigned> candidateGroupIdx;
+
+  for (unsigned i = 0; i < candidates.size(); ++i) {
+    auto it = srdToGroupIdx.find(candidates[i].srd);
+    if (it == srdToGroupIdx.end()) {
+      // Compute stride for this SRD group using the first candidate.
```
harsh-nod: Same comment as below, is the stride uniform across all threads?

wave_lang/kernel/wave/asm/wave_asm/lib/Transforms/BufferLoadStrengthReduction.cpp
```diff
@@ -0,0 +1,421 @@
...
+      Value voff0 =
+          cloneChainBeforeLoop(candidates[i].deps, candidates[i].voffset,
+                               ivInit, loopOp, body, builder);
+      Value voff1 =
+          cloneChainBeforeLoop(candidates[i].deps, candidates[i].voffset,
+                               ivPlusStep, loopOp, body, builder);
+      Value strideVGPR =
+          V_SUB_U32::create(builder, loc, vregType, voff1, voff0);
```
harsh-nod: The stride is computed as `voff(iv+step) - voff(iv)` in VGPR space, then `v_readfirstlane_b32` extracts lane 0's value. This is correct **only if the stride is uniform across all threads**.

For typical GEMM patterns like `(tid + iv) << shift`, the stride `((tid + step) << shift) - (tid << shift)` cancels the tid term and equals `step << shift` — uniform. Good.

But `isAddressVALU` accepts `V_MUL_LO_U32`. Consider `v_mul_lo_u32(tid, iv)`:
- stride = `tid * (iv + step) - tid * iv = tid * step` — **non-uniform** across threads.
- `v_readfirstlane_b32` would give `tid_0 * step`, incorrect for all other threads.

Maybe verify uniformity by checking that the stride VGPR has no tid-dependency (conservative: check that the dep chain, with IV constants, doesn't reference any precolored VGPR or block arguments)

wave_lang/kernel/wave/asm/wave_asm/lib/Transforms/BufferLoadStrengthReduction.cpp
```diff
@@ -0,0 +1,421 @@
...
+      continue;
+    bodyBuilder.clone(op, mapping);
+  }
+
+  // Patch buffer_loads: set voffset to precomputed value, soffset to iter_arg.
+  for (unsigned i = 0; i < candidates.size(); ++i) {
+    // Find the cloned buffer_load.
+    Operation *clonedLoad = nullptr;
+    for (Value result : candidates[i].loadOp->getResults()) {
+      Value clonedResult = mapping.lookup(result);
```
harsh-nod: This will crash if lookup fails?

PR #958, opened 2026-02-24
Implement mxfp4 split-k gemm

willghatch:
The core things added are split-k gemm, and it is tested for (1) generation of the `buffer_atomic_pk_add_bf16` instruction that we wanted to use, and (2) for gemm correctness.

Overview of changes unrelated to wave_asm:

- `remove_global_indexing` in `general_utils.py`: Zeroes out tiling constraint starts (e.g. `K_SPLIT_OFF`) alongside workgroup IDs before dimension scaling, so that the subtraction of the start offset doesn't mix scaled and unscaled units (K vs K/32 for MXFP4 scales).

- Fixing spurious bounds on split-K tiling that prevented scale vector merging: TilingConstraint.get_index_bound was conservatively generating bounds for the split-K case because sympy could not prove that ceiling(Min(K, f(wg)) / tile)
  * tile <= K.  These bounds prevented merge_contiguous_reads from combining scalar scale reads into vector<4xi8> loads (it skips reads that already have bounds).  Add _work_may_exceed_dim() to structurally detect the aligned split-k pattern and prove no overshoot, avoiding the spurious bound.  (This was necessary to get scale_preshuffle to have 4x vector loads when combined with split-k.)


wave_lang/kernel/wave/utils/general_utils.py
```diff
@@ -173,20 +173,28 @@ def remove_global_indexing(
     tiling_constraints = [c for c in constraints if isinstance(c, TilingConstraint)]
     workgroup_ids = [WORKGROUP_0, WORKGROUP_1, WORKGROUP_2]
     subs = {w: 0 for w in workgroup_ids}
+    # Zero out tiling constraint starts (e.g. K_SPLIT_OFF) alongside
```
harsh-nod: Do we need this?

tests/kernel/wave_gemm_test.py
```diff
@@ -3614,3 +3624,47 @@ def testSplitKGemm(
...
+    torch_ref = torchScaledGemmMXFP4(x, w, x_scales, w_scales)
+
+    x_gpu = x.cuda()
+    w_t_gpu = w.T.contiguous().cuda()
+    x_scales_gpu = x_scales.cuda()
+    w_scales_gpu = w_scales.cuda()
+    c_gpu = device_zeros(m, n, dtype=torch.bfloat16)
+
+    splitk_gemm(x_gpu, x_scales_gpu, w_t_gpu, w_scales_gpu, c_gpu)
+    assert_close(c_gpu.cpu().to(torch.float32), torch_ref, rtol=5e-2, atol=5e-2)
```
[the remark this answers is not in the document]
willghatch: Because we round to bf16 from f32, which has some error (especially at high exponents), then we sum splits which also sums the error beyond the normal f32->bf16 rounding error.

wave_lang/kernel/wave/constraints.py
```diff
@@ -760,6 +760,52 @@ def get_index_bound(self, vector_shape: Optional[int]) -> Optional[IndexExpr]:
...
+    # The Min caps the numerator at dim, so
+    #   Min(dim, x) <= dim
+    #   ceiling(Min(dim, x) / tile) <= ceiling(dim / tile) = dim / tile
+    #   tile * ceiling(Min(dim, x) / tile) <= dim
+    # as long as dim is evenly divisible by tile.
+    if isinstance(dim_bound, (int, Integer)):
+        dim_int = int(dim_bound)
+        tile, ceil_expr = _extract_tile_and_ceiling(work_bound)
+        if tile is not None and dim_int % tile == 0 and ceil_expr is not None:
+            numerator = (ceil_expr.args[0] * tile).simplify()
```
harsh-nod: Is the use of .simplify important here? Are you relying sympy to transform this to a canonical form?

examples/python/7.1_schedule.py
```diff
@@ -373,6 +381,40 @@ def test_dbuf_4wave_mxfp_dynamic_preshuffle_b_gemm(
     print("MXFP GEMM preshuffle-B 4-wave (WaveASM backend) test passed!")
 
 
+def test_splitk_preshuffle_scales_gemm_cpp(
```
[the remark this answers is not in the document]
willghatch: It now has an option for splitk in the other tests, but it is not yet supported in the ones using waveasm backend.  This one test is still there to have something using waveasm.  The others have issues with register allocation in waveasm.  I can look at it more tomorrow.

harsh-nod (changes requested):

wave_lang/kernel/wave/constraints.py
```diff
@@ -760,6 +760,53 @@ def get_index_bound(self, vector_shape: Optional[int]) -> Optional[IndexExpr]:
...
+        return int(work_bound) > int(dim_bound)
+
+    # The Min caps the numerator at dim, so
+    #   Min(dim, x) <= dim
+    #   ceiling(Min(dim, x) / tile) <= ceiling(dim / tile) = dim / tile
+    #   tile * ceiling(Min(dim, x) / tile) <= dim
+    # as long as dim is evenly divisible by tile.
+    if isinstance(dim_bound, (int, Integer)):
+        dim_int = int(dim_bound)
+        tile, ceil_expr = _extract_tile_and_ceiling(work_bound)
```
harsh-nod: The `work_bound` property returns `start + count * tile_size` — a sympy `Add`. The function's `_extract_tile_and_ceiling` expects a pure `Mul` with 2 args, so it always returns `(None, None)` for split-K. Is this expected?

wave_lang/kernel/wave/templates/tagged_mxfp4_gemm.py
```diff
@@ -573,6 +579,341 @@ def repeat(
...
+        B_ADDRESS_SPACE: (GLOBAL_ADDRESS_SPACE if preshuffle_B else a_address_space),
+        BLOCK_M: block_shape[0],
+        BLOCK_N: block_shape[1],
+        BLOCK_K: block_shape[2],
+        BLOCK_S: 1,
+        M: m,
+        N: n,
+        K: k,
+        S: num_splits,
+        K_SPLIT_OFF: WORKGROUP_2 * k_per_split,
```
harsh-nod: Should we add an assert that this only works when k_per_split % BLOCK_K = 0?

harsh-amd (approved):

harsh-nod (approved):

PR #967, opened 2026-02-24
[asm] Memory offset optimization: split oversized constants

Hardcode84:
## Summary

- Extend the `MemoryOffsetOptimization` pass to handle constants that exceed the hardware offset limit by splitting them into an aligned high part (`K_hi`) kept in the `v_add_u32` and a low residual (`K_lo`) folded into the static offset field.
- Before: `v_add_u32(base, 68608)` with `offset:0` — full constant in VALU.
- After: `v_add_u32(base, 65536)` with `offset:3072` — residual absorbed into hardware.
- The downstream `ScopedCSE` pass merges `v_add_u32` ops that now share the same `K_hi` and base, eliminating redundant VALU instructions across nearby loads.


ftynse (commented):
It is much better when the commit message explains _why_ a change is made, we can see _what_ it is from reading the code.

PR #968, opened 2026-02-24
[wave2water] fix diagnostics emission

ftynse:
Diagnostics were not properly reported when coming from generated MLIR
verification or parsing. This was due to MLIR Python bindings
intercepting emitted diagnostics before they could hit our handler and
wrapping them into an exception that was mostly ignored.

### DIFF
--- a/wave_lang/kernel/wave/mlir_converter/water_emitter.py
+++ b/wave_lang/kernel/wave/mlir_converter/water_emitter.py
@@ -48,6 +48,7 @@ from wave_lang.kernel.wave.mlir_converter.diagnostics import (
     LocationFrame,
     MLIRDiagnostic,
     NameLocation,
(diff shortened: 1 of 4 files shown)


wave_lang/kernel/wave/mlir_converter/water_emitter.py
```diff
@@ -1088,12 +1120,22 @@ def _flush_output(
     sys.stdout.flush()
 
 
+from enum import Enum
+
+
+class WaterDiagTestingMode(Enum):
+    NO = "no"
+    DIRECT = "direct"
+    VERIFIER = "verifier"
```
martin-luecke: We could move this to diagnostics.py, where other related stuff (WaterError, MLIRDiagnostic, ...) is defined

PR #969, opened 2026-02-24
[asm] Add ScalePackElimination pass

Hardcode84:
- Add a new `ScalePackElimination` pass that eliminates redundant BFE->iter_arg->LSHL_OR round-trips for B-scale values in scaled MFMA loops.
- Before: the loop carries 4 separate byte iter_args, repacks them with 3 `v_lshl_or_b32` ops every iteration, and re-extracts with 4 `v_bfe_u32` ops at the bottom — 7 wasted VALU ops per iteration.
- After: the loop carries the dword directly as a single iter_arg, feeding `v_mfma_scale` without any pack/unpack overhead.
- Wire the pass into `waveasm-translate` via `--waveasm-scale-pack-elimination`, placed after peephole and before buffer load strength reduction.
- Add lit tests covering the basic transformation, a negative case (byte arg has extra uses), and a negative case (yield BFEs source from different dwords).
- NFC: relax `LoopAddressPromotionPass` from `OperationPass<ModuleOp>` to `OperationPass<>` so it can run on any root operation.

### DIFF
--- a/wave_lang/kernel/wave/asm/wave_asm/lib/Transforms/LoopAddressPromotion.cpp
+++ b/wave_lang/kernel/wave/asm/wave_asm/lib/Transforms/LoopAddressPromotion.cpp
@@ -306,8 +306,13 @@ struct LoopAddressPromotionPass
   }
 
   void runOnOperation() override {
-    getOperation()->walk(
-        [](LoopOp loopOp) { applyLoopAddressPromotion(loopOp); });
+    // Post-order so inner loops are processed before outer ones.
+    // Collect first since transformation invalidates walk iterator.
+    SmallVector<LoopOp> loops;
+    getOperation()->walk<WalkOrder::PostOrder>(
+        [&](LoopOp loopOp) { loops.push_back(loopOp); });
+    for (auto loopOp : loops)
+      applyLoopAddressPromotion(loopOp);
   }
 };
 


wave_lang/kernel/wave/asm/wave_asm/lib/Transforms/ScalePackElimination.cpp
```diff
@@ -0,0 +1,408 @@
...
+    return "waveasm-scale-pack-elimination";
+  }
+
+  StringRef getDescription() const override {
+    return "Eliminate BFE/LSHL_OR round-trips for B-scale iter_args";
+  }
+
+  void runOnOperation() override {
+    ModuleOp module = getOperation();
+    // Collect loops first (transformation invalidates walk iterator).
```
harsh-nod: Should this walk be in post-order so inner loops are processed first? This is to handle nested loops, where if you process the outer loop first, the nested inner loop will get erased. Also do this also affect LoopAddressPromotion?

wave_lang/kernel/wave/asm/wave_asm/lib/Transforms/ScalePackElimination.cpp
```diff
@@ -0,0 +1,408 @@
...
+//     %new_dword = buffer_load_dword ...
+//     %nb0 = v_bfe_u32(%new_dword, 0, 8)
+//     %nb1 = v_bfe_u32(%new_dword, 8, 8)
+//     %nb2 = v_bfe_u32(%new_dword, 16, 8)
+//     %nb3 = v_bfe_u32(%new_dword, 24, 8)
+//     yield ..., %nb0, %nb1, %nb2, %nb3, ...
+//
+// After optimization:
+//   Loop carries the dword directly. The LSHL_OR chain and BFE extractions
+//   are eliminated, saving 6 VALU + 8 BFE instructions per iteration.
```
harsh-nod: Shouldnt this say saving 3 lhsl_or + 4 bfe = 7 valu instructions per chain per iteration?

wave_lang/kernel/wave/asm/wave_asm/lib/Transforms/ScalePackElimination.cpp
```diff
@@ -0,0 +1,408 @@
...
+          auto offsetImm = postBuilder.getType<ImmType>(offset);
+          auto offsetConst =
+              ConstantOp::create(postBuilder, loc, offsetImm, offset);
+          auto widthImm = postBuilder.getType<ImmType>(8);
+          auto widthConst = ConstantOp::create(postBuilder, loc, widthImm, 8);
+          Value dwordResult = newLoop.getResult(chainNewArgIdx[ci]);
+          auto bfe = V_BFE_U32::create(postBuilder, loc, vregType, dwordResult,
+                                       offsetConst, widthConst);
+          loopOp.getResult(i).replaceAllUsesWith(bfe);
+          goto nextResult; // Break out of both inner loops.
```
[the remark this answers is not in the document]
Hardcode84: done

PR #970, opened 2026-02-25
[asm] New peephole opts and asm LICM pass

Hardcode84:
New peephole patterns, a new LICM pass, extraction of M0 redundancy elimination into its own pass, and a fix in the assembly emitter for VGPR swap cycles.

**New peephole patterns** (in `Peephole.cpp`):
- `BufferLoadLDSSoffsetPattern`: extract SGPR shifts from `buffer_load_lds` voffset into soffset (3 cases: direct shift, nested add, buried in `v_lshl_add_u32` addend).
- `BFEPackIdentityPattern`: eliminate redundant BFE-extract-then-repack of 4 consecutive bytes from the same dword.
- `LshlAddPattern` guard: skip fusion when SGPR shift feeds a `buffer_load_lds` chain, letting the soffset pattern handle it instead.

**New pass — `waveasm-licm`** (`LICM.cpp`): loop-invariant code motion — hoists VALU address-computation ops whose operands all dominate the loop out of `LoopOp` bodies. Iterative fixpoint with post-order walk for nested loops. Runs after peephole so fused ops (e.g. `v_lshl_add_u32`) get hoisted as a single unit.

**Extracted pass — `waveasm-m0-redundancy-elim`** (`M0RedundancyElimination.cpp`): M0 redundancy elimination previously lived inline in the peephole pass. Extracted into its own pass for independent pipeline ordering. Fixed a latent bug: now conservatively resets tracking across region-bearing ops (loops, ifs) that may clobber M0 in their bodies.

**`waveasm-peephole`** (`Peephole.cpp`): now purely local pattern-based rewrites via `applyPatternsGreedily`.

All three passes are declared in `Passes.td` and use `OperationPass<>` (not anchored to `ModuleOp`).

**Pipeline order:** peephole -> LICM -> M0 elim.

**Assembly emitter fix** (`AssemblyEmitter.cpp`): VGPR swap cycles in loop iter_args are now supported. Previously only SGPR swaps were handled; VGPR swaps hit an assertion. LICM can increase register pressure enough to trigger VGPR swap cycles at the loop back-edge, so the emitter now handles them with a 3-move swap via a temporary VGPR (mirroring the existing SGPR logic).

### DIFF
--- a/wave_lang/kernel/wave/asm/wave_asm/lib/Transforms/M0RedundancyElimination.cpp
+++ b/wave_lang/kernel/wave/asm/wave_asm/lib/Transforms/M0RedundancyElimination.cpp
@@ -17,6 +17,9 @@
 
 #include "mlir/Pass/Pass.h"
 #include "llvm/ADT/STLExtras.h"
+#include "llvm/Support/DebugLog.h"
+
+#define DEBUG_TYPE "waveasm-m0-redundancy-elim"
 
 namespace waveasm {
 #define GEN_PASS_DEF_WAVEASMM0REDUNDANCYELIM
@@ -34,7 +37,8 @@ struct M0RedundancyEliminationPass
   using WAVEASMM0RedundancyElimBase::WAVEASMM0RedundancyElimBase;
 
   void runOnOperation() override {
-    getOperation()->walk([](Block *block) {
+    unsigned numEliminated = 0;
+    getOperation()->walk([&](Block *block) {
       Value lastM0Source;
 
       for (Operation &op : llvm::make_early_inc_range(*block)) {
@@ -50,12 +54,15 @@ struct M0RedundancyEliminationPass
         Value src = m0Op.getSrc();
 
         // If M0 already holds this value, the write is redundant.
-        if (src == lastM0Source)
+        if (src == lastM0Source) {
           op.erase();
-        else
+          ++numEliminated;
+        } else {
           lastM0Source = src;
+        }
       }
     });
+    LDBG() << "removed " << numEliminated << " redundant M0 writes.";
   }
 };
 


wave_lang/kernel/wave/asm/wave_asm/lib/Transforms/LICM.cpp
```diff
@@ -0,0 +1,92 @@
...
+          return defOp->getParentRegion() != loopRegion;
+        // Block arguments of the loop body do NOT dominate (they're
+        // loop-carried values that change each iteration).
+        if (auto blockArg = dyn_cast<BlockArgument>(val))
+          return blockArg.getOwner()->getParentOp() != loopOp.getOperation();
+        return false;
+      };
+
+      // Check if an op is a VALU (address computation) that's safe to hoist.
+      auto isHoistableVALU = [](Operation *op) -> bool {
```
harsh-nod: Rather than hardcoding this, can we use an op trait Pure or NoMemoryEffect?

wave_lang/kernel/wave/asm/wave_asm/lib/Transforms/M0RedundancyElimination.cpp
```diff
@@ -0,0 +1,81 @@
...
+  StringRef getArgument() const override {
+    return "waveasm-m0-redundancy-elim";
+  }
+
+  StringRef getDescription() const override {
+    return "Eliminate redundant M0 register writes in WAVEASM IR";
+  }
+
+  void runOnOperation() override {
+    getOperation()->walk([](ProgramOp program) {
```
harsh-nod:
```cpp
getOperation()->walk([](ProgramOp program) {
  program.walk([](Block *block) { ... });
});
```

The outer walk finds `ProgramOp`s, then the inner walk finds blocks. Since the pass is `OperationPass<>` (anchored on any op), if the root operation *is* a `ProgramOp`, the walk is slightly redundant. A simpler approach:

```cpp
getOperation()->walk([](Block *block) { ... });
```

This works because blocks within non-`ProgramOp` parents would have the same semantics. The current code is correct but has an unnecessary layer.

wave_lang/kernel/wave/asm/wave_asm/lib/Transforms/LICM.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+#include "mlir/Pass/Pass.h"
+#include "llvm/ADT/SmallVector.h"
+
+using namespace mlir;
+using namespace waveasm;
+
+namespace {
+
+struct LICMPass : public PassWrapper<LICMPass, OperationPass<>> {
```
ftynse: This defines a pass _separate_ from the one defined in `Passes.td` and will be highly confusing.

wave_lang/kernel/wave/asm/wave_asm/lib/Transforms/M0RedundancyElimination.cpp
```diff
@@ -0,0 +1,81 @@
+// Copyright 2025 The Wave Authors
```
ftynse: Happy new year

wave_lang/kernel/wave/asm/wave_asm/lib/Transforms/M0RedundancyElimination.cpp
```diff
@@ -0,0 +1,81 @@
...
+    getOperation()->walk([](ProgramOp program) {
+      program.walk([](Block *block) {
+        Value lastM0Source;
+        SmallVector<Operation *, 8> toErase;
+
+        for (Operation &op : *block) {
+          auto m0Op = dyn_cast<S_MOV_B32_M0>(&op);
+          if (!m0Op) {
+            // Conservatively reset tracking across region-bearing ops
+            // (loops, ifs) — their bodies may clobber M0.
```
ftynse: Not for here, but this is something that should be modeled via side effects.

wave_lang/kernel/wave/asm/wave_asm/lib/Transforms/M0RedundancyElimination.cpp
```diff
@@ -0,0 +1,81 @@
...
+          // If M0 already holds this value, the write is redundant.
+          if (src == lastM0Source) {
+            toErase.push_back(m0Op);
+          } else {
+            lastM0Source = src;
+          }
+        }
+
+        for (auto *op : toErase)
+          op->erase();
```
ftynse: Why can't we erase as we go? `llvm::make_early_inc_range` in the loop should help with the iterators.

wave_lang/kernel/wave/asm/wave_asm/lib/Transforms/Peephole.cpp
```diff
@@ -780,12 +942,102 @@ struct FoldMovConstIntoAndPattern : public OpRewritePattern<V_AND_B32> {
...
+    auto bfe3 = outerOp.getSrc0().getDefiningOp<V_BFE_U32>();
+    if (!bfe3)
+      return failure();
+
+    // Middle: v_lshl_or_b32(byte2, 16, inner_result).
+    auto middleOp = outerOp.getSrc2().getDefiningOp<V_LSHL_OR_B32>();
+    if (!middleOp)
+      return failure();
+
+    auto shiftMiddle = getConstantValue(middleOp.getSrc1());
```
ftynse: MLIR has an "almost always *avoid* auto" style, don't use auto unless it improves readability (RHS is a cast, a lambda or an iterator).

PR #971, opened 2026-02-25
Add post-scheduling buffer count optimization

panditsa:
Adds a check that reduces shared memory buffer counts in pipelined loops by detecting the consume-before-write pattern.

This targets the asymmetric MXFP4 GEMM (3-stage pipeline: GatherToLDS at stage 0, shared reads at stages 1-2), reducing A's LDS usage from ~102KB to ~68KB.


panditsa:
Dealing with the lit test I realized that the optimization needs more work. I have a workaround for mxfp4 gemm. 


Closing this for now!

PR #972, opened 2026-02-25
Minimal functional WaveASM backend for MXFP4 preshuffle-B GEMM

panditsa:
## Wave ASM (functional correctness only, no optimizations)
- **AssemblyEmitter.cpp**: Fix multi-register loop back-edge copies so iter_args wider than 1 dword (e.g. `dwordx4`) don't silently corrupt after iteration 0.
- **AffineHandlers.cpp**: Generalize the `floor(x/d)==0` constant fold from thread-ID-specific upper bounds to BitRange-based analysis, enabling MXFP4 index patterns.
- **VectorHandlers.cpp**: Add sub-register packed extraction via `V_BFE_U32` for types where multiple sub-32-bit elements pack into one VGPR (e.g. `vector<4xi8>`).

## Compilation infrastructure
- **compile.py / compile_options.py**: Integrate the WaveASM backend into `wave_compile()` via `use_wave_asm_backend` option, with `print_mlir`/`print_mlir_file` support for the extracted kernel MLIR.
- **7.1_schedule.py / utils.py**: Simplify C++ backend tests from ~220 lines of manual orchestration to ~15 lines using the standard `wave_compile()` flow.

## Python scheduling/codegen
- **read_write.py**: Three-way index splitting (workgroup / uniform / thread) so the backend can place induction-variable contributions in `buffer_load` soffset.

### DIFF
--- a/lit_tests/kernel/wave/water_host_wrapper.py
+++ b/lit_tests/kernel/wave/water_host_wrapper.py
@@ -90,18 +90,15 @@ def test_read_write():
     print(read_write.asm)
 
     # CHECK-LABEL:    test_read_write
-    # CHECK-DAG:        #[[MAP0:.*]] = affine_map<()[s0] -> (s0 - (s0 floordiv 64) * 48)>
     # CHECK:          gpu.module @gpu_module
     # CHECK:          gpu.func @read_write
     # CHECK-SAME:       (%[[D0:.*]]: memref<f16> {llvm.inreg}, %[[D1:.*]]: memref<f16> {llvm.inreg})
     # CHECK-SAME:       kernel attributes {known_block_size = array<i32: 64, 1, 1>}
-    # CHECK-DAG:        %[[C0:.*]] = arith.constant 0 : index
     # CHECK:            %[[thread_id_x:.*]] = gpu.thread_id  x
-    # CHECK:            %[[S0:.*]] = memref.reinterpret_cast %[[D0]] to offset: [0], sizes: [16, 16], strides: [16, 1] : memref<f16> to memref<16x16xf16, strided<[16, 1]>>
-    # CHECK:            %[[S1:.*]] = memref.reinterpret_cast %[[D1]] to offset: [0], sizes: [16, 16], strides: [16, 1] : memref<f16> to memref<16x16xf16, strided<[16, 1]>>
-    # CHECK:            %[[I0:.*]] = affine.apply #[[MAP0]]()[%[[thread_id_x]]]
-    # CHECK:            %[[V:.*]] = vector.load %[[S0]][%[[I0]], %[[C0]]] : memref<16x16xf16, strided<[16, 1]>>, vector<16xf16>
-    # CHECK:            vector.store %[[V]], %[[S1]][%[[I0]], %[[C0]]] : memref<16x16xf16, strided<[16, 1]>>, vector<16xf16>
+    # CHECK:            memref.reinterpret_cast %[[D0]] to offset: [0], sizes: [{{.*}}], strides: [1] : memref<f16> to memref<{{.*}}xf16, strided<[1]>>
+    # CHECK:            memref.reinterpret_cast %[[D1]] to offset: [0], sizes: [{{.*}}], strides: [1] : memref<f16> to memref<{{.*}}xf16, strided<[1]>>
+    # CHECK:            vector.load {{.*}} : memref<{{.*}}xf16, strided<[1]>>, vector<16xf16>
+    # CHECK:            vector.store {{.*}} : memref<{{.*}}xf16, strided<[1]>>, vector<16xf16>
     # CHECK:            return
 
     # CHECK-LABEL:    func.func @isolated_benchmark


wave_lang/kernel/wave/asm/wave_asm/lib/Transforms/AssemblyEmitter.cpp
```diff
@@ -604,7 +604,13 @@ std::optional<std::string> KernelGenerator::generateOp(Operation *op) {
                 auto [dstPhys, dstIsSGPR] = getPhysRegInfo(body.getArgument(i));
 
                 if (srcPhys >= 0 && dstPhys >= 0 && srcPhys != dstPhys) {
-                  pendingCopies.push_back({dstPhys, srcPhys, isSGPR});
+                  // Multi-register iter_args (e.g. dwordx4) need one copy per
+                  // sub-register, otherwise only the first register is copied
+                  // and the remaining lanes silently corrupt after iteration 0.
+                  int64_t width = getRegSize(body.getArgument(i).getType());
```
harsh-nod: The expanded sub-register copies are appended in ascending order and later emitted sequentially. If the destination range overlaps the source range and `dstPhys > srcPhys`, earlier copies will overwrite source registers that later copies still need to read.

**Concrete example:** Suppose an iter_arg maps `v[2:5] → v[4:7]` (width 4). The expansion produces:
```
{dst=4, src=2}, {dst=5, src=3}, {dst=6, src=4}, {dst=7, src=5}
```
Emitted sequentially:
- `v_mov_b32 v4, v2` — writes v4
- `v_mov_b32 v5, v3` — writes v5
- `v_mov_b32 v6, v4` — **reads v4, which now holds v2, not the original v4**
- `v_mov_b32 v7, v5` — **reads v5, which now holds v3, not the original v5**

The result is that v6 and v7 receive copies of v2 and v3, not the original v4 and v5. The existing swap-detection logic (lines 619–645) only handles the specific case where two copies form a complete swap cycle (`A→B` paired with `B→A`); it does not detect write-before-read hazards within a partially-overlapping copy sequence.

**Recommendation:** Either:
- (a) Add an assertion that source and destination register ranges do not overlap:
  ```cpp
  assert(!(srcPhys < dstPhys + width && dstPhys < srcPhys + width) ||
         srcPhys == dstPhys &&
         "overlapping src/dst ranges in multi-register iter_arg copy");
  ```
- (b) Copy in reverse order when `dstPhys > srcPhys` (the standard `memmove` strategy):
  ```cpp
  if (dstPhys > srcPhys) {
    for (int64_t r = width - 1; r >= 0; --r)
      pendingCopies.push_back({dstPhys + r, srcPhys + r, isSGPR});
  } else {
    for (int64_t r = 0; r < width; ++r)
      pendingCopies.push_back({dstPhys + r, srcPhys + r, isSGPR});
  }
  ```

wave_lang/kernel/wave/asm/wave_asm/include/waveasm/Dialect/WaveASMOps.td
```diff
@@ -680,7 +680,19 @@ def WaveASM_S_SEXT_I32_I16 : SALUUnaryOp<"s_sext_i32_i16">;
 //===----------------------------------------------------------------------===//
 
 // Arithmetic
-def WaveASM_S_ADD_U32 : SALUBinaryOp<"s_add_u32">;
+// s_add_u32 and s_addc_u32 set SCC (carry flag) as a side effect,
+// so they must NOT be marked Pure (prevents DCE of SRD base adjustments
+// that write to precolored registers without SSA users).
+def WaveASM_S_ADD_U32 : WAVEASMOp<"s_add_u32", [WaveASM_ArithmeticOp]> {
```
harsh-nod: Rather than ignore SCC, can you follow the SALUCmpOp and return SCC as an additional result? This way SCC is explicitly modeled in the IR and the verifier will prevent reordering ops that corrupt the carry chain.

wave_lang/kernel/wave/asm/wave_asm/lib/Transforms/TranslateFromMLIR.cpp
```diff
@@ -1271,7 +1275,70 @@ LogicalResult handleVectorStore(Operation *op, TranslationContext &ctx) {
...
+    // Check for pending per-workgroup SRD base adjustment (from linearized
+    // reinterpret_cast for >4GB output buffers). Emitted inline here so the
+    // SALU ops survive DCE — their results go straight into the precolored
+    // SRD that is immediately consumed by buffer_store.
+    if (auto *adj = ctx.getPendingSRDBaseAdjust(storeOp.getBase())) {
+      int64_t N = ctx.getNextSwizzleSRDIndex();
+      auto *mlirCtx = builder.getContext();
+
+      // Copy source SRD base to new SRD
+      std::string copyBase = "s_mov_b64 s[" + std::to_string(N) + ":" +
```
harsh-nod: We don't want to be emitting strings here as this bypasses register allocation and all downstream passes directly. Model the entire SRD setup sequence using typed WaveASM ops. If `s_mov_b64` is needed, add it to the dialect. If precolored register writes are needed, use `PrecoloredSRegOp` + proper SALU ops.

wave_lang/kernel/wave/asm/wave_asm/lib/Transforms/handlers/MemRefHandlers.cpp
```diff
@@ -110,28 +110,76 @@ LogicalResult handleMemRefView(Operation *op, TranslationContext &ctx) {
 }
 
 /// Handle memref.reinterpret_cast - track memref identity
-/// CRITICAL: This propagates LDS offset and updates SRD buffer size
+/// CRITICAL: This propagates LDS offset and updates SRD buffer size.
```
harsh-nod: Can we remove CRITICAL from the comments? Probably artifacts from the LLM

PR #973, opened 2026-02-25
Add persistent emitter for MLIR roundtrip subprocess reuse

### LINKED ISSUE #964: Emitter hits Python recursion limit on vanilla attention pattern
`emit_wave_dialect` exceeds Python's default recursion limit when converting the vanilla attention kernel. Currently worked around in the test by bumping `sys.setrecursionlimit(10000)`.

```python
# TODO: investigate why the recursion limit is hit in the emitter.
limit = sys.getrecursionlimit()
sys.setrecursionlimit(10000)

with Context(), Location.unknown():
    mlir_output, diagnostics, _ = emit_wave_dialect(trace, constraints, options)
sys.setrecursionlimit(limit)
```

The emitter likely has deep recursion over the trace graph that should be converted to an iterative approach.

martin-luecke:
## Summary

- Add `PersistentEmitter` context manager that keeps `water_emitter` and `fx_emitter` subprocesses alive across multiple conversion calls, communicating over stdin/stdout instead of spawning a fresh process each time.
- Extract shared helpers (`_prepare_water_request`, `_unpack_water_response`, `_unpack_fx_response`, `_start_emitter`) to deduplicate the standalone and persistent code paths in `mlir_converter.py`.
- Use `PersistentEmitter` in all three MLIR lit tests (`mlir_roundtrip_pipeline.py`, `mlir_converter.py`, `mlir_to_fx.py`).

## Performance

Each emitter subprocess spends ~2s on Python imports (water_mlir bindings). Previously every `emit_wave_dialect` / `mlir_to_fx` call spawned a fresh process. Approximate savings:

| Test file | Before | After | Speedup |
|---|---|---|---|
| `mlir_roundtrip_pipeline.py` (94 spawns) | ~180s | ~17s | ~10x |
| `mlir_converter.py` (24 spawns) | ~60s | ~14s | ~4x |
| `mlir_to_fx.py` (8 spawns) | ~20s | ~6s | ~3x |

Total MLIR converter test time: ~260s → ~37s (~7x faster, saving ~3.5 min in CI). All measurements done on my local machine (Apple M3 Max)

- The attention test in `mlir_converter.py` previously had a manual `sys.setrecursionlimit(10000)` workaround around the entire `emit_wave_dialect` call. This has been removed. The recursion limit bump is now handled internally, scoped specifically to the `dill.dumps` and `dill.loads` calls that serialize/deserialize the `CapturedTrace` object graph, and restored via try/finally immediately after. We should eventually explore changing the structure of `CapturedTrace` to avoid this entirely

Closes #964

### DIFF
--- a/tests/kernel/wave/test_subprocess_protocol.py
+++ b/tests/kernel/wave/test_subprocess_protocol.py
@@ -1,4 +1,4 @@
-# Copyright 2025 The Wave Authors
+# Copyright 2026 The Wave Authors
 #
 # Licensed under the Apache License v2.0 with LLVM Exceptions.
 # See https://llvm.org/LICENSE.txt for license information.
--- a/wave_lang/kernel/wave/mlir_converter/fx_emitter.py
+++ b/wave_lang/kernel/wave/mlir_converter/fx_emitter.py
@@ -1064,7 +1064,7 @@ def _process_single_request(mlir_text: str) -> bytes:
     return dill_util.dumps(response)
 
 
-def _server_mode() -> int:
+def _server_mode() -> None:
     """Run as a persistent server, reading length-prefixed requests from stdin.
 
     Instead of spawning a fresh process per request, the parent keeps this
@@ -1092,22 +1092,28 @@ def _server_mode() -> int:
         try:
             request = dill_util.loads(raw_request)
         except Exception as e:
-            sys.stderr.write(f"FATAL: failed to unpickle: {e}\n")
-            break
+            response = FxEmitterResponse()
+            response.diagnostics.append(
+                WaterError(message=f"Failed to unpickle request: {e}")
+            )
+            send_message(sys.stdout.buffer, dill_util.dumps(response))
+            continue
 
         mlir_text = request.get("mlir") if isinstance(request, dict) else None
         if not isinstance(mlir_text, str):
-            sys.stderr.write(
-                f"FATAL: expected 'mlir' string in request, got: {type(mlir_text)}\n"
+            response = FxEmitterResponse()
+            response.diagnostics.append(
+                WaterError(
+                    message=f"Expected 'mlir' string in request, got: {type(mlir_text)}"
+                )
             )
-            break
+            send_message(sys.stdout.buffer, dill_util.dumps(response))
+            continue
 
         response_data = _process_single_request(mlir_text)
 
         send_message(sys.stdout.buffer, response_data)
 
-    return 0
-
 
 if __name__ == "__main__":
-    sys.exit(_server_mode())
+    _server_mode()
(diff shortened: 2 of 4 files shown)


wave_lang/kernel/wave/mlir_converter/fx_emitter.py
```diff
@@ -1041,7 +1041,108 @@ def _diagnostics_handler(d: ir.Diagnostic) -> bool:
...
+    while True:
+        # Read the 4-byte length header for the next request.
+        header = sys.stdin.buffer.read(header_size)
+        if len(header) < header_size:
+            break  # Parent closed stdin -- shut down.
+
+        (length,) = struct.unpack(header_fmt, header)
+        raw_request = sys.stdin.buffer.read(length)
+        if len(raw_request) < length:
+            break  # Truncated payload -- parent died mid-write.
```
ftynse: Should we warn about this somehow?

wave_lang/kernel/wave/mlir_converter/mlir_converter.py
```diff
@@ -364,18 +443,117 @@ def mlir_to_fx(
...
+    alone.
+
+    Usage::
+
+        with PersistentEmitter() as emitter:
+            mlir, diags, attrs = emitter.emit_wave_dialect(trace, constraints, options)
+            trace2, cons2, opts2, diags2 = emitter.mlir_to_fx(mlir)
+    """
+
+    def __init__(self) -> None:
```
ftynse: Should we further parameterize this to indicate whether we want a bidirectional conversion or not and, when not, only start one of the processes?

ftynse (commented):
Nice! Overall design comments:

 - Do we actually need non-server mode? If so, we should have a test for it.
 - Can we extract the protocol logic into a separate package that only depends on dill and sys, so it's importable everywhere without mlir/iree dependencies and testable separately?

wave_lang/kernel/wave/mlir_converter/fx_emitter.py
martin-luecke: There is no parent process anymore to report anything to. I clarified the error message and print it to stderr now

wave_lang/kernel/wave/mlir_converter/fx_emitter.py
```diff
@@ -1041,7 +1041,108 @@ def _diagnostics_handler(d: ir.Diagnostic) -> bool:
...
+    parser = argparse.ArgumentParser(description="FX emitter (MLIR -> FX)")
+    parser.add_argument(
+        "--server",
+        action="store_true",
+        help="Run in persistent server mode (length-prefixed protocol on stdin/stdout)",
+    )
+    args = parser.parse_args()
+
+    if args.server:
+        sys.exit(_server_mode())
```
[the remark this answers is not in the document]
martin-luecke: removed non-server mode and simplified all of this away

tests/kernel/wave/test_subprocess_protocol.py
```diff
@@ -0,0 +1,89 @@
+# Copyright 2025 The Wave Authors
```
ftynse:
```suggestion
# Copyright 2026 The Wave Authors
```

PR #975, opened 2026-02-25
Epilogue elimination in MXFP4

xintin:
With epilogue elimination, the loop simply runs for N extra iterations instead. Those extra iterations load OOB data, but 
- VGPR buffer loads return 0 (hardware guarantee via SRD validBytes)
- GatherToLDS uses branchless SRD zeroing: `validBytes = select(iv + prefetch_offset < max_iv, real_validBytes, 0)`. When OOB, `validBytes=0` makes `NUM_RECORDS=0` in the SRD, so the hardware DMA becomes a no-op without any branch.

 Since the OOB loads produce zeros, the extra MFMAs are `0 x something + accumulator = accumulator`, no-ops that don't change the result. The epilogue code is eliminated.

* Add a parameter eliminate_epilogue to get_mxfp4_asymmetric_schedule
* Pass it through to tkw.pipeline(k_loop, eliminate_epilogue=eliminate_epilogue)

### DIFF
--- a/wave_lang/kernel/compiler/wave_codegen/read_write.py
+++ b/wave_lang/kernel/compiler/wave_codegen/read_write.py
@@ -217,7 +217,8 @@ def _get_max_buffer_size(elem_type: IrType) -> int:
 
     Buffer ops offsets are i32, return maximum memref size in elements.
     """
-    return ((1 << 31) - 1) // (elem_type.width // 8)
+    # Use _elem_bytes to avoid division-by-zero for sub-byte types (e.g. mxfp4).
+    return ((1 << 31) - 1) // _elem_bytes(elem_type)
 
 
 def _get_strides_from_memref(mem: Value) -> list[Value]:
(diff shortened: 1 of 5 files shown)


wave_lang/kernel/compiler/wave_codegen/read_write.py
```diff
@@ -362,25 +385,99 @@ def _get_constant_value(candidate: Value):
...
+    cond_val = gen_sympy_index(add_emitter_subs(emitter), guard_condition)
+    i1 = IntegerType.get_signless(1)
+    if cond_val.type != i1:
+        zero_idx = arith_d.constant(cond_val.type, 0)
+        cond_val = arith_d.cmpi(arith_d.CmpIPredicate.ne, cond_val, zero_idx)
+
+    return arith_d.select(cond_val, real_valid, zero_valid)
+
+
 def _cast_buffer_and_encode_stride(
```
harsh-nod: Could you remove is_read from this as the functionality of this function is going beyond its original definition to something like

```
mem = _cast_buffer_and_encode_stride(
      mem, strides, element_type, emitter,
      valid_bytes_override=_compute_valid_bytes(
          mem, element_type,
          symbolic_shape if is_read else None,
          emitter,
      ),
  )
```

and then compute valid bytes becomes
```
  def _compute_valid_bytes(ptr, elem_type, symbolic_shape, emitter) -> Value:
      use_real_bounds = (
          emitter.options.eliminate_epilogue and symbolic_shape is not None
      )
      total_bytes = _compute_total_valid_bytes(elem_type, symbolic_shape, use_real_bounds)
      uint64 = IntegerType.get_signless(64)

      if use_real_bounds:
          metadata = memref_d.extract_strided_metadata(ptr)
          offset_elements = metadata[1]
          offset_bytes = arith_d.index_cast(uint64, offset_elements)
          elem_bytes_val = arith_d.constant(uint64, get_constant_attr(elem_type.width // 8, uint64))
          offset_bytes = arith_d.muli(offset_bytes, elem_bytes_val)
          total_val = arith_d.constant(uint64, get_constant_attr(total_bytes, uint64))
          return arith_d.subi(total_val, offset_bytes)

      return arith_d.constant(uint64, get_constant_attr(total_bytes, uint64))
```

PR #977, opened 2026-02-25
Three-way index split for GatherToLDS uniform offset folding

panditsa:
## Summary

- Introduces `_split_index_three_way()` in `read_write.py` to decompose index expressions into **workgroup**, **uniform (IV)**, and **thread-dependent** components
- Adds `_linearize_uniform_offset()` helper to linearize the IV part into a single scalar offset suitable for `buffer_load` soffset (SGPR)
- Updates `handle_gather_to_lds` to use the three-way split when induction variables are present, keeping the uniform IV offset as a separate `arith.addi` operand
- Updates `gather_to_shared.py` CHECK lines and adds a new `uniform_offset_split.py` lit test


harsh-nod (approved):
lgtm! just a minor comment

PR #980, opened 2026-02-26
Mxfp4 8 wave kernel with shuffling

adedespirlet:
This PR introduces 8 wave kernels for MXFP4 along with their associated schedules. Reaching 3.4 PFLOP for 16384x16384x16384 without reordering.

**Two variants are implemented:**

**Normal ping-pong**: each phase consists exclusively of either compute operations or memory operations
**Mixed ping-pong:** each phase can include both compute and memory operations

### **Motivation**
The mixed ping pong schedule addresses a memory dependency issue between the 2 wave clusters.

### **Problem**
With staggering the two clusters read from global memory and write to LDS at different times. The cluster that starts later issues its global loads later. Under memory bus congestion, it may happen that the global loads from the later cluster do not arrive in LDS before the other cluster attempts to read from it.
To prevent incorrect results, we must introduce a second workgroup barrier to realign the timing between clusters.
However this additional barrier introduces significant latency.
### **Solution:** Mixed Ping-Pong
The mixed ping-pong schedule mitigates this by overlapping the second barrier with useful work that is independent of the other cluster’s global load operations. This helps hide the added synchronization cost.

### **Additional Notes**
Both kernels pre shuffle A and B scales for improved memory coalescing. These are loaded directly into VGPRs.
B cannot be shuffled and loaded directly into VGPRs because prefetching would require twice the number of VGPRs, leading to register spilling.

It may be interesting to investigate whether using AGPRs could be a solution to that?

### DIFF
--- a/wave_lang/kernel/wave/schedules/gemm_mxfp4_double_buffer.py
+++ b/wave_lang/kernel/wave/schedules/gemm_mxfp4_double_buffer.py
@@ -379,11 +379,6 @@ def get_mxfp4_dbuf_pingpong_schedule(use_stagger: bool = True, shape: tuple = No
         # If the bus gets congested and cluster memory dependency are affected, we must add a second barrier to fix the timing and prevent incorrect output results.
         # In case a second a second workgroup barrier is needed, another schedule is created to hide the latency of that second barrier, by scheduling safe ds_read ops before the second barrier (see get_mxfp4_dbuf_mixed_pingpong_schedule).
         use_extra_barrier = False
-<<<<<<< HEAD
-     
-=======
-
->>>>>>> 79cd92cb (get rid of dead code)
         # Build cluster 0: first K-partition loads + bitcasts + GatherToLDS
         cluster_0_ops = [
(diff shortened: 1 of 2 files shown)


wave_lang/kernel/wave/schedules/gemm_mxfp4_double_buffer.py
```diff
@@ -381,40 +367,25 @@ def mxfp4_dbuf_schedule():
...
+        # In case a second a second workgroup barrier is needed, another schedule is created to hide the latency of that second barrier, by scheduling safe ds_read ops before the second barrier (see get_mxfp4_dbuf_mixed_pingpong_schedule).
         use_extra_barrier = False
-        if shape is not None and len(shape) >= 3:
-            threshold = (1024, 1024, 1024)
-            use_extra_barrier = (
-                shape[0] > threshold[0]
-                and shape[1] > threshold[1]
-                and shape[2] > threshold[2]
-            )
+        # if shape is not None and len(shape) >= 3:
```
harsh-nod: dead code?

PR #981, opened 2026-02-26
[compiler] Support dynamic M, N, K dimensions in MXFP4 scaled GEMM test

harsh-nod:
The MXFP4 kernel uses packed memory types (K/2, K/32) which produced unresolved _Rational expressions when K was a dynamic symbol. This fixes the codegen and host dispatch to handle compound symbolic dimension expressions, enabling fully dynamic M, N, K for MXFP4.

- emitter: resolve _Rational results via floor division in gen_sympy_index
- host_codegen: infer dynamic symbol values from compound buffer dims (e.g. K from K/2), compute compound dim runtime values for dispatch
- compile: extend get_dynamic_dimension_actual with the same inference
- test: add dynamic_dims param_bool to testScaledGemmMXFP4


suryajasper (approved):
lgtm

PR #985, opened 2026-02-26
Handle pre-canonical FX forms in trace equivalence checker

martin-luecke:
## Summary
- Auto-detect pre-canonical FX graphs (Iterate present, GetResult absent) and
  filter GetResult nodes from the round-tripped side during comparison.
- Add directional TilingConstraint comparator that accepts None for
  induction_var/iters on the source side (similar to how we already do with HardwareConstraint)
- All 47 MLIR roundtrip passes now succeed with no expected failures.

### DIFF
--- a/wave_lang/kernel/wave/utils/graph_utils.py
+++ b/wave_lang/kernel/wave/utils/graph_utils.py
@@ -692,9 +692,11 @@ def _check_graphs_equivalent(
     lhs_nodes = list(lhs.nodes)
     rhs_nodes = list(rhs.nodes)
 
-    # Auto-detect pre-canonical FX: if lhs has Iterate but no GetResult nodes,
-    # the source graph predates the add_get_results pass. Filter
-    # GetResult from both sides so they can be compared structurally.
+    # Auto-detect pre-canonical FX: if lhs has Iterate but no GetResult,
+    # add_get_results has not run yet. Filter GetResult from rhs so the
+    # two sides can be compared structurally.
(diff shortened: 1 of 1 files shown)


wave_lang/kernel/wave/utils/graph_utils.py
```diff
@@ -663,6 +692,14 @@ def _check_graphs_equivalent(
     lhs_nodes = list(lhs.nodes)
     rhs_nodes = list(rhs.nodes)
 
+    # Auto-detect pre-canonical FX: if lhs has Iterate but no GetResult nodes,
+    # the source graph predates the add_get_results pass. Filter
+    # GetResult from both sides so they can be compared structurally.
+    lhs_has_iterate = any(isinstance(get_custom(n), Iterate) for n in lhs_nodes)
+    lhs_has_getresult = any(isinstance(get_custom(n), GetResult) for n in lhs_nodes)
+    if lhs_has_iterate and not lhs_has_getresult:
```
ftynse: Should we guard against situations where we have both iterate and getresult?

martin-luecke: Added a comment to clarify. When both Iterate and GetResult are present (in the proper canonical form), the condition is False, and no filtering occurs, so both sides are compared with their GetResult nodes as usual. add_get_results is all-or-nothing, so a partial state shouldn't arise, but even if it did, the check is conservative: any GetResult on the source side disables filtering.

PR #990, opened 2026-02-26
[water] rework broadcast operation

ftynse:
The initial design for broadcast assumed only trailing dimensions may be
broadcasted based on the intent behind the pywave counterpart. However,
the practice shows that, even in pywave, broadcast can apply to
_arbirary_ dimensions, not necessarily the trailing ones. Redesign the
operation to support that by requiring the result type to be always
provided. This is consistent with having the list of indexing symbols in
the frontend. Consequently, disable type inference and EPT propagation
since they cannot be achieved without knowing the positions of
broadcasted dimensions. Index inference forward is identity to ensure
non-broadcasted dimensions match, and backward is near-identity that
drops broadcasted dimensions.

Add a lowering for broadcast operations to either `vector.broadcast` or
noop depending on EPT.

### DIFF
--- a/water/lib/Dialect/Wave/IR/WaveOps.cpp
+++ b/water/lib/Dialect/Wave/IR/WaveOps.cpp
@@ -2486,8 +2486,7 @@ FailureOr<ChangeResult> wave::BroadcastOp::propagateIndexExprsBackward(
     wave::EmitErrorFn emitError) {
   auto sourceTensorType = dyn_cast<WaveTensorType>(getSource().getType());
   if (!sourceTensorType) {
-    emitError() << "expected source tensor type, got "
-                << getOperand().getType();
+    emitError() << "expected source tensor type, got " << getSource().getType();
     return failure();
   }
 


water/lib/Dialect/Wave/IR/WaveOps.cpp
```diff
@@ -2531,78 +2525,6 @@ LogicalResult wave::BroadcastOp::verify() {
   return success();
```
martin-luecke: technically, our verifier and lowering allows for Broadcasts ops that reorder the original dimensions, similar to Permute

water/lib/Dialect/Wave/IR/WaveOps.cpp
```diff
@@ -2513,13 +2509,11 @@ LogicalResult wave::BroadcastOp::verify() {
...
 
-  if (!sourceType.getFullySpecified() || !resultType.getFullySpecified())
-    return success();
-
-  if (getBroadcastDims()) {
-    return emitOpError("does not expect explicit dims when source and result "
-                       "types are fully specified");
-  }
+  // When result is a tensor, require both source and result to be fully
+  // specified (vectors are unchanged).
```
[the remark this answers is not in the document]
ftynse: We don't have that requirement actually, source can be underspecified and its type may be inferred later at the definition point. We only need the result type to be fully-specified because we cannot infer it.

water/lib/Dialect/Wave/IR/WaveOps.cpp
```diff
@@ -2484,7 +2484,7 @@ FailureOr<ChangeResult> wave::BroadcastOp::propagateIndexExprsBackward(
     llvm::MutableArrayRef<wave::IndexExprsLatticeStorage> operandIndexExprs,
     llvm::ArrayRef<wave::IndexExprsLatticeStorage> resultIndexExprs,
     wave::EmitErrorFn emitError) {
-  auto sourceTensorType = dyn_cast<WaveTensorType>(getOperand().getType());
+  auto sourceTensorType = dyn_cast<WaveTensorType>(getSource().getType());
   if (!sourceTensorType) {
     emitError() << "expected source tensor type, got "
                 << getOperand().getType();
```
martin-luecke: tiny nit: update `getOperand()` in error message as well to avoid confusion

PR #995, opened 2026-02-27
[water] lowering read/write with mappings

ftynse:
Make read/write lowering handle permutations in the mapping by swapping the order of index expressions following the inverse permutation (because the mapping is memory->value and the index expressions are for value). Practically this means reordering `ordered_syms` since the MLIR dictionary attribute order is always alphabetical. Note that this is partly inconsistent with the python counterpart as it only permutes start expressions, which it get can away with by not using step/stride after permutation.

Change `ordered_sym` to track value shape instead of memory shape; it wasn't relevant until now because the two were verifiably identical, but will matter with permutation mappings. The order of index in pywave is that of the `indexing_dims`, which is taken from the non-memory part of the mapping (output for reads, input for writes), which is now verified to match the value shape when present. Update the distributed shape resolution pass accordingly. Add verifiers to guarantee this.

Change the logic of generating transfer_read/write, which should be dependent on the which dimension is vectorized according to the index expression _not_ the strides of the memref, which transfer read should be capable of handling by itself. Be defensive about it and bail out from patterns should memref layout have non-consecutive strides. We don't have such cases currently.

We should consider revising the read/write lowerings holistically since pywave doesn't seem to use transfer_read/write and the logic around mapping and masking is questionable.

### DIFF
--- a/docs/wave/ir_design_notes.rst
+++ b/docs/wave/ir_design_notes.rst
@@ -89,36 +89,106 @@ are represented.  As a result, the MLIR-imported trace may list more
 captures than the source trace.
 
 
+Index Expressions and Index Sequence Analysis
+---------------------------------------------
+
+Index expressions describe how elements of a tensor are distributed across
+multiple workitems (threads) and workgroups (blocks), as well as across multiple
+copies of the operation processing the tensor after the _expansion_ (unrolling)
+process. They are recorded per dimension and consist of three components:
+
+  - `start` (also `offset`) - position of the first element "owned" by the
+    workitem along this dimension;
+  - `step` (also `size`) - number of elements "owned" by the workitem along this
+    dimension;
+  - `stride` - number of elements in linear indexing of this dimension or,
+    **inconsistently**, of the canonical row-major storage needed to get to the
+    next element "owned" by the workitem.
+
+Note the inconsistency in stride definition, which depends on the kind of
+operation that the index expression was propagate from: per-dimension strides if
+propagated from MMA-style operations, row-major Torch-style strides otherwise.
+
+In Python, each component of the index expression is a sympy `Expr` using
+specially-named symbols to refer to workgroup and item identifiers.
+Per-dimension expressions are organized in a dictionary where their order is
+load-bearing and matches that of the `indexing_dims` property of the parent
+operation. Typically, this order is that of unique non-reduction dimensions of
+the consecutive operands followed by reduction dimensions. For memory access
+operations, the order is that of the non-memory operand. Index expressions are
+associated with the operation via the `index` attribute, which may contain
+multiple entries; one entry typically corresponds to the operation's result;
+multiple entries typically correspond to the operation's operands.
+
+In MLIR, each component of the index expression is an `AffineMap` using
+exclusively symbols. The entirety of the components uses the same list of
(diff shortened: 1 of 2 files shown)


water/lib/Dialect/Wave/Transforms/LowerReadWriteOps.cpp
```diff
@@ -291,6 +260,61 @@ static void buildVectorWrite(Location loc, PatternRewriter &rewriter, Value mem,
...
+
+  // XXX: step/stride are not permuted similarly to pywave. For step, this
+  // appears necessary to avoid losing track of the dimension that is
+  // vectorized. For stride, not sure, it may be unused at all at this stage.
+  SmallVector<NamedAttribute> newIndexDictEntries;
+  for (size_t i = 0, e = updatedOrderedSyms.size(); i < e; ++i) {
+    auto updatedIndexMapping = cast<wave::WaveIndexMappingAttr>(
+        indexDict.get(orderedSyms[i].getName()));
+    auto originalIndexMapping = cast<wave::WaveIndexMappingAttr>(
+        indexDict.get(updatedOrderedSyms[i].getName()));
```
martin-luecke: these names are confusing. Why does the `originalIndexMapping` come from `updatedOrderedSyms` and above vice versa?

water/test/Dialect/Wave/ops-invalid.mlir
```diff
@@ -344,39 +344,39 @@ func.func @mismatch_shape_write(%lhs: !wave.tensor<[@A, @B] of f32, <register>>,
...
 // -----
 
-// ordered_syms symbols must match memory tensor shape in order (wave.write).
+// ordered_syms symbols must match value tensor shape in order (wave.write).
 func.func @write_ordered_syms_symbol_mismatch(%val: !wave.tensor<[@M, @K, @N] of f32, <register>>, %mem: !wave.tensor<[@M, @K, @N] of f32, <global>>) {
-  // expected-error @below {{'ordered_syms' symbol at index 1 ('N') does not match memory tensor shape symbol ('K')}}
+  // expected-error @below {{'ordered_syms' symbol at index 1 ('N') does not match value tensor shape symbol ('K')}}
   wave.write %val, %mem {ordered_syms = [#wave.symbol<"M">, #wave.symbol<"N">, #wave.symbol<"K">]}
     : !wave.tensor<[@M, @K, @N] of f32, <register>>, !wave.tensor<[@M, @K, @N] of f32, <global>>
   return
```
martin-luecke: We should consider adding a negative test where a mapping is present and ordered_syms matches the memory shape instead of the value shape, e.g.:
```mlir
func.func @read_ordered_syms_mismatch_with_mapping(%mem: !wave.tensor<[@N, @M] of f32, <global>>) {
  // expected-error @below {{'ordered_syms' symbol at index 0 ('N') does not match value tensor shape symbol ('M')}}
  %0 = wave.read %mem {
    ordered_syms = [#wave.symbol<"N">, #wave.symbol<"M">],
    mapping = #wave.expr_list<[] (d0, d1) -> (d1, d0)>
  } : (!wave.tensor<[@N, @M] of f32, <global>>) -> !wave.tensor<[@M, @N] of f32, <register>>
  return
}
```

PR #996, opened 2026-02-27
[water] use vectorized dim in read/write mask generation

ftynse:
Existing implementation was assuming the trailing dimension is the vectorized one, which is not always the case. The vectorized dimension is the one for which index expression step is greater than 1. We already compute that, but were not passing it into the mask generator. Do so.

Remark: it is unclear why transfer_read/write accepts a 1D mask for an nD memref, should be investigated separately.

### DIFF
--- a/water/lib/Dialect/Wave/Transforms/LowerReadWriteOps.cpp
+++ b/water/lib/Dialect/Wave/Transforms/LowerReadWriteOps.cpp
@@ -394,7 +394,7 @@ createMemoryIndicesAndMask(ConversionPatternRewriter &rewriter,
          "NYI: only permutation mappings are currently supported");
   FailureOr<Value> mask =
       buildMask(op->getLoc(), boundsMapping, memoryShape, rewriter, hyper,
-                startIndices, elementsPerThread);
+                startIndices, elementsPerThread, *vectorizedDim);
   if (failed(mask))
     return rewriter.notifyMatchFailure(op, "couldn't build the required mask");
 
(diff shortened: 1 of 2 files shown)


water/lib/Dialect/Wave/Transforms/LowerReadWriteOps.cpp
```diff
@@ -85,17 +85,18 @@ buildStartIndices(Location loc, DictionaryAttr indexDict,
...
 /// whenever a bounds mapping is provided. When it is not provided, return a
-/// null mask. If the vectorized dimension cannot be identified, return failure.
+/// null mask. elements_per_thread is considered to be 1 for any dimension
+/// other than `vectorizedDim` and the given value for `vectorizedDim`.
 static FailureOr<Value>
 buildMask(Location loc, wave::WaveSymbolMappingAttr boundsMapping,
           ArrayRef<wave::WaveSymbolAttr> orderedSyms, PatternRewriter &rewriter,
           wave::WaveHyperparameterAttr hyper, ArrayRef<Value> startIdx,
-          int64_t elementsPerThread) {
+          int64_t elementsPerThread, uint64_t vectorizedDim) {
```
martin-luecke: we should be consistent in picking `uint64_t` or `int64_t`. 
`vectorizedDim` in `createMemoryIndicesAndMask` is `std::optional<int64_t>` and `getPositionOfVectorizedDim` returns `std::optional<uint64_t>`

PR #999, opened 2026-02-27
[asm] MXFP4 fixes for C++ backend

panditsa:
- Add SALUBinaryWithCarryOp for s_add/s_sub ops that set SCC
- Fix S_ADD_U32 type arguments in BufferLoadStrengthReduction
- Add ticketed-waitcnt pass option
- Fix multi-register iter_arg copy ordering in AssemblyEmitter
- Add VOP3 constant bus restriction checks in Peephole
- Convert RawOp strings to typed WaveASM ops in SRD prologue
- Add pending SRD base adjustment tracking for >4GB buffers
- Fix stride=0 rejection in strength reduction
- Update lit tests for new op signatures

### DIFF
--- a/wave_lang/kernel/wave/asm/wave_asm/test/Dialect/region-control-flow.mlir
+++ b/wave_lang/kernel/wave/asm/wave_asm/test/Dialect/region-control-flow.mlir
@@ -25,11 +25,11 @@ waveasm.program @simple_loop
   // CHECK:      %[[INIT:.*]] = waveasm.s_mov_b32 %{{.*}} : !waveasm.imm<0> -> !waveasm.sreg
   // CHECK-NEXT: %{{.*}} = waveasm.loop (%[[IV:.*]] = %[[INIT]]) : (!waveasm.sreg) -> !waveasm.sreg {
   %final = waveasm.loop(%i = %init) : (!waveasm.sreg) -> (!waveasm.sreg) {
-    // CHECK-NEXT:   %[[NEXT:.*]]:2 = waveasm.s_add_u32 %[[IV]], %{{.*}} : !waveasm.sreg, !waveasm.imm<1> -> !waveasm.sreg, !waveasm.sreg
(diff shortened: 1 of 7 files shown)


wave_lang/kernel/wave/asm/wave_asm/lib/Transforms/Ticketing.cpp
```diff
@@ -490,22 +491,24 @@ struct InsertWaitcntPass
...
-      ++st.opIndex;
+    if (ticketedWaitcnt) {
+      for (Operation *op : allOps) {
+        ++st.opIndex;
 
-      if (observeExistingWaitcnt(op, st))
-        continue;
-      if (handleBarrier(op, st))
-        continue;
+        if (observeExistingWaitcnt(op, st))
```
harsh-nod: minor: tabs?

wave_lang/kernel/wave/asm/wave_asm/include/waveasm/Dialect/WaveASMOps.td
```diff
@@ -98,6 +98,17 @@ class SALUBinaryOp<string mnemonic, list<Trait> traits = []>
   let assemblyFormat = "$src0 `,` $src1 attr-dict `:` type($src0) `,` type($src1) `->` type($dst)";
 }
 
+// SALU Binary with carry/SCC: dst, scc = op(src0, src1)
+// s_add_u32 and s_addc_u32 set SCC (carry flag) as a side effect.
+// We model SCC as an explicit sreg result so the verifier can prevent
+// reordering ops that corrupt the carry chain.
```
harsh-nod: I think we will need to sync about sec propagation more carefully as it could be the case that the acc produced by one of these operators might be used in a subsequent operator creating a dependency chain. Right now this is not possible because the acc value is not really fed into any of these operators. I think it's OK to skip this for this full request, but let's add a TODO for it.

PR #1000, opened 2026-02-27
[compiler] MXFP4 frontend and scheduling changes

panditsa:
- Add ticketed_waitcnt option to compile options
- Add multi_buffering parameter to schedule ops
- Update MXFP4 double buffer schedule for preshuffle B
- Fix lit tests for updated codegen patterns
- Add 7.1_schedule.py test improvements
- Update read_write.py for MXFP4 support

### DIFF
--- a/lit_tests/kernel/wave/water_host_wrapper.py
+++ b/lit_tests/kernel/wave/water_host_wrapper.py
@@ -87,6 +87,8 @@ def test_read_write():
         tkw.write(res, b)
 
     read_write = wave_compile(get_wave_compile_options(canonicalize=True), read_write)
+    with open("read_write.mlir", "w") as f:
+        f.write(read_write.asm)
     print(read_write.asm)
 
     # CHECK-LABEL:    test_read_write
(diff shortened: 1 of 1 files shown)


wave_lang/kernel/ops/wave_schedule_ops.py
```diff
@@ -183,7 +183,7 @@ def reorder_graph(loop: Any, clusters: Any): ...
 
 
 @define_schedule_op
-def pipeline(iterate: Sequence[fx.Node]): ...
+def pipeline(iterate: Sequence[fx.Node], multi_buffer_count: Optional[int] = None): ...
```
harsh-nod: Why does this take multi buffer count as an input when it's not used anywhere in the function?

wave_lang/kernel/ops/wave_schedule_ops.py
```diff
@@ -601,53 +601,123 @@ def handle(
...
 
-        # Apply the reordering to the subgraph
+        # Determine target graph based on pipeline stage
+        is_parent_graph_stage = isinstance(loop, PipelineStageRef) and loop.stage in (
+            PipelineStage.EPILOGUE,
+            PipelineStage.PROLOGUE,
+        )
+
+        if is_parent_graph_stage:
+            return cls._reorder_parent_graph(kernel_trace, loop.stage, cluster_nodes)
```
harsh-nod: Why do we need a separate function for re ordering the parent graph? Can we not use reorder_graph?

lit_tests/kernel/wave/water_host_wrapper.py
```diff
@@ -87,21 +87,20 @@ def read_write(
         tkw.write(res, b)
 
     read_write = wave_compile(get_wave_compile_options(canonicalize=True), read_write)
+    with open("read_write.mlir", "w") as f:
+        f.write(read_write.asm)
```
tgymnich: @panditsa why do we write this file to disk? Is this a left over from debugging?

PR #1003, opened 2026-02-27
Support MXFP4 MT_64x192x256

xintin:
**To fix `M=64`,**
`b_index = {0, 1, 2, 3}` was for the 4 waves. But there are only 2 valid m-blocks (0 and 1), `BLOCK_M/32 from preshuffle scale formula`. Waves 2 and 3 computed b_index = 2, 3, which exceeded num_m_blocks = 2, causing:
Out-of-bounds global memory reads (DMA from invalid addresses)
Out-of-bounds LDS writes (lds_base = 2*256 = 512 when total LDS was only 512 bytes)


wave_lang/kernel/wave/preshuffle_scale_to_shared.py
```diff
@@ -257,7 +257,9 @@ def _transform_scale_memory(
     elements_per_thread = 4
     num_waves = total_threads // threads_per_wave
     num_m_blocks = m_count // 32
-    num_dma_ops = max(1, num_m_blocks // num_waves)
+
+    # ensures we schedule enough DMA ops to cover all m-blocks even when num_m_blocks doesn't divide evenly by num_waves
+    num_dma_ops = max(1, (num_m_blocks + num_waves - 1) // num_waves)
```
harsh-nod: just a suggestion, we have an existing ceildiv implementation that you could use.

harsh-nod (approved):
LGTM!

PR #1008, opened 2026-02-28
[asm] Replace amdclang++ with clang++ from LLVM build

Hardcode84:
Build clang as part of the LLVM build and install it alongside waveasm-translate, removing the dependency on an external ROCm toolkit for assembly compilation.


harsh-nod (approved):
lgtm!

PR #1032, opened 2026-03-03
[water] Support IndexMapping roundtrip in MLIR <-> FX conversion

martin-luecke:
## Summary

- Reconstruct `IndexMapping` from the `WaveExprListAttr` permutation map on `wave.read`/`wave.write` during MLIR-to-FX import, enabling mapping roundtrip for permutation-only mappings.
- Add `IndexMapping.__eq__` to check for structural equality rather than object identity
- Add some tests for roundtripping mappings and generalize testing infrastructure a bit

### DIFF
--- a/wave_lang/kernel/wave/mlir_converter/fx_emitter.py
+++ b/wave_lang/kernel/wave/mlir_converter/fx_emitter.py
@@ -226,10 +226,10 @@ def _convert_mapping_attr(
     value_type: WaveTensorType,
     *,
     is_read: bool,
-) -> IndexMapping:
+) -> IndexMapping | None:
     """Reconstruct an IndexMapping from a Water MLIR permutation map attribute.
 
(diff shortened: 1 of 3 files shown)


wave_lang/kernel/wave/mlir_converter/fx_emitter.py
```diff
@@ -218,6 +220,56 @@ def _convert_read_write_bounds(
...
+    from the memory and value types.
+
+    For reads, input_mapping describes memory dims; for writes, output_mapping
+    describes memory dims (matching the convention in water_emitter.py).
+
+    Note: dynamic_val_mappings are not represented in the MLIR attribute and
+    are not reconstructed here. Roundtripping mappings with dynamic values
+    is not currently supported.
+    """
+    affine_map = mapping_attr.map
```
ftynse: Better to be safe and check for `affine_map` not being None, at least the MLIR side may have a null map.

PR #1037, opened 2026-03-04
[water] Fix water emitter crash on iterate blocks with reduction loop-carried values

martin-luecke:
## Summary

The water emitter crashed when emitting iterate blocks whose loop-carried
values include reductions (e.g. attention's online-softmax loop carrying
acc, max, and sum). This blocked MLIR emission for all passes before
`infer_types` (passes 1–7).

- Derive iterate result types from `init_args` (already-emitted enclosing
  graph) instead of calling `infer_type()` on body output nodes whose
  types are not yet available.
- Collect `IterArg`s in graph order instead of via `iter_args()` which
  sorts by `iter_idx` — a field that is `None` before `initialize_iter_args`.
- Propagate captured outer-node types to implicit-capture placeholders so
  downstream body nodes can infer their types before `infer_types` runs.

### DIFF
--- a/lit_tests/kernel/wave/mlir_converter.py
+++ b/lit_tests/kernel/wave/mlir_converter.py
@@ -21,6 +21,7 @@ from wave_lang.kernel.wave.compile import (
     build_graph_passes,
     wave_compile,
 )
+from wave_lang.kernel.wave.type_inference import infer_types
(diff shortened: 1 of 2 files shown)


ftynse (approved):
This is what one gets for not having canonical forms in the IR :(

PR #1042, opened 2026-03-04
minor compile time optimizations

willghatch:
piecewise_aware_subs:

Add is_number early return before computing free_symbols. Use frozenset and intersection() instead of set and & for the subs key matching.

_split_index:

Skip the expensive simplify/expand when the difference between the original expression and the WG-zeroed expression is already zero, meaning the expression has no workgroup symbols.


willghatch:
This one shows some improvement locally, but it's within the noise.  I want to see if it makes a difference in CI.  If not, I'm going to close this PR.

harsh-nod:
looks like its within the noise

willghatch:
Yeah, it looks like this doesn't really make a difference.  I think we are out of big slam dunk optimizations, at least by looking at the dynamic mxfp4 compilation benchmark I've been using.

PR #1053, opened 2026-03-05
[water] Support variadic reduction ops in Water dialect and add corresponding simplification pass

martin-luecke:
Extends the Water dialect reduction ops (wave.sum, wave.max_element) to accept variadic inputs, matching the PyWave representation, where expand_graph tiles reduction inputs into a list of slices. This simplifies FX <-> MLIR roundtrips by allowing the dialect to directly represent the intermediate form, rather than requiring the Python side to decompose reductions before emission, track which reductions stem from this, and fuse them again for the roundtrip.

A new ExpandVariadicReductions pass chains N variadic inputs into N single-input reductions, each feeding its result as the next accumulator — a partial port of the logic in PyWave's decompose_reduce_ops pass. Both the Water emitter and FX importer have been updated to handle variadic forms in both directions.
A normal-form annotation for expanded reductions could be added to indicate where in the pipeline single-input reductions are expected, though currently this would only be relevant for codegen, I think.

### DIFF
--- a/water/include/water/Dialect/Wave/Transforms/Passes.td
+++ b/water/include/water/Dialect/Wave/Transforms/Passes.td
@@ -198,6 +198,10 @@ def WaterWaveExpandVariadicReductionsPass
     supports variadic inputs for faithful roundtripping with the Python
     representation. This pass normalizes them before lowering, which
     requires single-input reductions.
+
+    This corresponds to the "source reduce" step (step 1) of the Python-side
+    `decompose_reduce_ops` pass, which additionally performs local element
+    reduction, cross-thread butterfly shuffles, and accumulator combination.
   }];
 }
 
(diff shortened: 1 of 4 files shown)


water/include/water/Dialect/Wave/Transforms/Passes.td
```diff
@@ -177,4 +177,28 @@ def WaterWaveResolveDistributedAllocationsPass
...
+      %0 = wave.sum %a init(%init) <scope>
+      %1 = wave.sum %b init(%0) <scope>
+      %r = wave.sum %c init(%1) <scope>
+
+    Variadic inputs arise from the Python-side graph expansion pass, which
+    tiles the reduction dimension into multiple slices. The Wave dialect
+    supports variadic inputs for faithful roundtripping with the Python
+    representation. This pass normalizes them before lowering, which
+    requires single-input reductions.
+  }];
```
ftynse: Could you document the differences between this pass and its python counterpart?

ftynse (approved):

PR #1054, opened 2026-03-05
[water] Add FX importer handlers for arithmetic, unary, and attention-specific ops

martin-luecke:
This adds MLIR-to-FX import handlers for the ops needed by attention kernels: binary arithmetic (add, sub, mul, max, min, select), unary (exp2, reciprocal), cast, permute, self_index, apply_expr, extract, and reshape.

The apply_expr handler reconstructs the original sympy expression lambda from the MLIR WaveExprListAttr and combinator attribute, folding the affine map results back through the appropriate sympy constructor.
As the reconstructed lambda is a new Python object, the existing trace equivalence checker cannot compare it by identity. A `_check_callable_equivalent` helper is added that evaluates both lambdas with fresh symbolic inputs and verifies the resulting sympy expressions are equivalent via simplify(a - b) == 0. This allows roundtrip tests to confirm that ApplyExpr nodes carry semantically identical expressions despite being distinct function objects.

### DIFF
--- a/wave_lang/kernel/wave/mlir_converter/fx_emitter.py
+++ b/wave_lang/kernel/wave/mlir_converter/fx_emitter.py
@@ -39,10 +39,10 @@ try:
         AllocateOp,
         ApplyExprOp,
         BroadcastOp,
-        CastOp as CastMlirOp,
+        CastOp,
         DivOp,
         Exp2Op,
-        ExtractOp as ExtractMlirOp,
+        ExtractOp,
         ExtractSliceOp,
         IterateOp,
         MaxElementOp,
@@ -54,8 +54,8 @@ try:
         ReadOp,
         ReciprocalOp,
         RegisterOp,
-        ReshapeOp as ReshapeMlirOp,
-        SelectOp as SelectMlirOp,
+        ReshapeOp,
+        SelectOp,
         SelfIndexOp,
         SubOp,
         SumOp,
@@ -108,7 +108,7 @@ from wave_lang.kernel.ops.wave_ops import (
     Allocate,
     ApplyExpr,
     Broadcast,
-    CastOp as CastFx,
+    CastOp as Cast,
     Exp2,
     Extract,
     ExtractSlice,
@@ -128,7 +128,7 @@ from wave_lang.kernel.ops.wave_ops import (
     Read,
     Reciprocal,
     Reshape,
-    SelectOp as SelectFx,
+    SelectOp as Select,
     SelfIndex,
     SharedMemoryBarrier,
     Sub,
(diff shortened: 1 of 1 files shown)


wave_lang/kernel/wave/mlir_converter/fx_emitter.py
```diff
@@ -836,6 +877,301 @@ def _handle_reduction_op(
...
+        combinator_fn = _combinator_to_sympy_fn(combinator_attr.value)
+
+    operand_syms = [
+        index_symbol(OPERAND_SYMBOL_NAME_WAVE_PREFIX + str(i))
+        for i in range(num_operands)
+    ]
+
+    def expr_lambda(*args):
+        subs_map = {operand_syms[i]: args[i] for i in range(len(args))}
+        results = [r.subs(subs_map) if hasattr(r, "subs") else r for r in sympy_results]
```
ftynse: Please add a comment as to when `hasattr` may fail. I suspect we may have plain integers as a result, but better to double-check. I'd also prefer stricter typing via `isinstance` over duck typing. Adding a type annotation on the `sympy_results` variable would also improve readability for me here.

wave_lang/kernel/wave/utils/graph_utils.py
```diff
@@ -239,6 +264,64 @@ def _nodes_match(expected: fx.Node, actual: fx.Node) -> bool:
     return False
 
 
+def _sympy_equiv(a: sympy.Basic, b: sympy.Basic) -> bool:
+    """Check symbolic equivalence of two sympy expressions.
+
+    For numeric/algebraic expressions this checks `simplify(a - b) == 0`.
+    For relational expressions (e.g. `x < 5`) and other non-numeric sympy
```
ftynse: For relational exprs, we can rewrite them to the canonical form `lhs - rhs < 0` (or whatever comparison operator)`, and then use `simplify(lhs_a - rhs_a - lhs_b + rhs_b) == 0` for a better equivalence check

martin-luecke: Good point!

wave_lang/kernel/wave/mlir_converter/fx_emitter.py
```diff
@@ -34,24 +35,40 @@
     from water_mlir.water_mlir import ir
     from water_mlir.water_mlir.dialects import wave, arith, amdgpu, func
     from water_mlir.water_mlir.dialects.wave import (
+        AddOp,
         AllocateOp,
+        ApplyExprOp,
         BroadcastOp,
+        CastOp as CastMlirOp,
```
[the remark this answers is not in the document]
martin-luecke: renamed to adhere to the implicit conversion for now

PR #1055, opened 2026-03-05
waveasm: add more ops

willghatch:
Mostly scf.if.

Stuff towards getting dynamic mxfp4 working.  This was the main piece to get a simple example compiling but not passing.  Just adopts the loop handling for register liveness, which will not be enough, but was the simplest thing to split this off.

Not clear how useful this is until I actually get stuff working on top of it, but this passes the basic tests.

### DIFF
--- a/waveasm/lib/Transforms/Liveness.cpp
+++ b/waveasm/lib/Transforms/Liveness.cpp
@@ -367,11 +367,9 @@ LivenessInfo computeLiveness(ProgramOp program) {
       if (parentOp && isa<LoopOp>(parentOp)) {
         Block *body = blockArg.getOwner();
         Operation *terminator = body->getTerminator();
-        if (terminator) {
-          auto termIt = opToIdx.find(terminator);
-          if (termIt != opToIdx.end()) {
-            range.end = std::max(range.end, termIt->second);
-          }
+        auto termIt = opToIdx.find(terminator);
+        if (termIt != opToIdx.end()) {
+          range.end = std::max(range.end, termIt->second);
         }
 
         // Extend to the next sibling op after the LoopOp. This bridges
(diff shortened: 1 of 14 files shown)


waveasm/lib/Transforms/Liveness.cpp
```diff
@@ -368,46 +368,63 @@ LivenessInfo computeLiveness(ProgramOp program) {
...
-                it->second.end = std::max(it->second.end, termIt->second);
+          return false;
+        };
+
+        // Extend end to the last terminator in any region of the op.
+        auto extendToRegionEnd = [&](Operation *regionOp) {
+          for (Region &region : regionOp->getRegions()) {
+            for (Block &block : region) {
+              Operation *terminator = block.getTerminator();
+              if (terminator) {
```
ftynse: IIRC, `getTerminator` will assert if the block does not have a terminator so its result should never be null, but may be worth doubl-checking.

waveasm/lib/Transforms/Liveness.cpp
```diff
@@ -540,6 +557,113 @@ LivenessInfo computeLiveness(ProgramOp program) {
...
+    auto thenYield = dyn_cast<YieldOp>(thenBlock.getTerminator());
+    if (!thenYield)
+      return;
+
+    YieldOp elseYield = nullptr;
+    if (Block *elseBlock = ifOp.getElseBlock()) {
+      elseYield = dyn_cast<YieldOp>(elseBlock->getTerminator());
+    }
+
+    for (unsigned i = 0; i < ifOp->getNumResults(); ++i) {
```
ftynse: https://llvm.org/docs/CodingStandards.html#don-t-evaluate-end-every-time-through-a-loop

willghatch:
This became outdated and some of the functionality was ultimately landed in other PRs.

PR #1058, opened 2026-03-06
Coalesce B-scale reads for dynamic-dim MXFP4 preshuffle kernels

Hardcode84:
### Problem

When M, N, K are dynamic, the read coalescer in `partition_strided_operators.py` fails to merge the 16 per-thread B-scale byte reads into a single contiguous `vector<16xi8>` load. Instead it produces fragmented loads of mixed widths ({2, 16, 8, 4}) stitched together with `vector.from_elements`.

Root cause: the pairwise merge verifies candidate pairs by computing per-dim offset diffs at multiple numeric probe points. For B-scale buffers with shape `[N, K/2]`, the 2D decomposition `row = offset floordiv K/2, col = offset mod K/2` gives inconsistent diffs when probe values (e.g. K=137 → K/2=68) don't respect the kernel's divisibility constraints (K % 256 == 0). The verification then rejects perfectly valid merge candidates.

With static dims this doesn't happen because K is concrete and the decomposition is trivially consistent.

### Solution

1. **Divisibility substitutions before probing**: plumb `get_divisibility_subs(constraints)` into the merge pipeline. Before numeric probing, apply forward subs like `K → 256*K'` so that `floordiv`/`Mod` evaluate consistently across all probe sets. This is the key fix — with it, B-scale reads coalesce into 8 clean `vector<16xi8>` loads identical to the static-dims case.

2. **Numeric probing for pairwise merge**: replace the old symbolic diff approach (which exploded on complex preshuffle index expressions) with concrete numeric evaluation using three linear generators that avoid pathological values.

3. **Re-merging across mask levels**: allow reads that already carry precomputed masks (from prior merge rounds) to participate in further merging. The mask is extended to cover the wider result and the precomputed condition is preserved.

4. **Bounds pre-flattening**: pre-compute each read's bounds check as a flat sympy boolean before merging, so the coalescer can reason about mask compatibility without re-deriving bounds at each merge step.

### DIFF
--- a/lit_tests/kernel/wave/scaled_gemm.py
+++ b/lit_tests/kernel/wave/scaled_gemm.py
@@ -10,6 +10,7 @@ from wave_lang.kernel.wave.constraints import (
     ScaledMMAType,
 )
 from wave_lang.kernel.wave.scheduling.schedule_enums import SchedulingType
+from wave_lang.kernel.wave.templates import get_tagged_mxfp4_gemm_preshuffle_b
 from wave_lang.kernel.wave.templates.test_kernels import (
     get_broadcasted_scale_gemm_mxfp4,
 )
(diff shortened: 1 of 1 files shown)


wave_lang/kernel/wave/analysis/partition_strided_operators.py
```diff
@@ -477,13 +613,402 @@ def _merge_contiguous_reads_once(trace: CapturedTrace, hw_constraint) -> bool:
...
-            # Skip reads that have bounds: the merged read would lose the
-            # mapping and source→target index, making mask generation incorrect.
-            if custom.bounds is not None:
-                continue
             key = (custom.memory, custom.elements_per_thread, region_id)
             groups[key].append(node)
+    return groups
+
+
+def _resolve_symbolic_diff(raw_diff, has_complex_mapping, expected_vals=None):
```
harsh-nod: is this dead code?

wave_lang/kernel/wave/analysis/partition_strided_operators.py
```diff
@@ -477,13 +613,402 @@ def _merge_contiguous_reads_once(trace: CapturedTrace, hw_constraint) -> bool:
...
+def _multiway_coalesce(
+    read_infos,
+    merged,
+    reads,
+    symbolic_dims,
+    symbolic_shape,
+    hw_constraint,
+    divisibility_fwd=None,
+):
+    """Coalesce unmerged ept==1 reads whose flat offsets fall in an aligned window.
```
harsh-nod: The probe-based approach computes `num_offs` once (O(n)), but then for each anchor, iterates over all probes:
```python
for anchor_idx in range(len(unmerged_infos)):
    ...
    for probe_idx in range(len(unmerged_infos)):
```
With the same dict-lookup pattern used for `_pairwise_merge`, the inner loop could check `offset_map[target]` for each window position instead of scanning all probes.

wave_lang/kernel/wave/analysis/partition_strided_operators.py
```diff
@@ -477,13 +613,402 @@ def _merge_contiguous_reads_once(trace: CapturedTrace, hw_constraint) -> bool:
...
+                    vp = _eval_expr(resolved_offs[probe_idx], ep)
+                    if vp - va != diff_val:
+                        consistent = False
+                        break
+                except (TypeError, ValueError, ZeroDivisionError):
+                    consistent = False
+                    break
+            if not consistent:
+                continue
+            _, custom_p, node_p = (
```
harsh-nod: Why not just unpack unmerged_infos directly?

PR #1061, opened 2026-03-06
[WIP] Fix overflow when shape >= 32Kx32Kx32K, buffer overflow

xintin:
* Buffer offset calculations used signed 32-bit limits (2^31 - 1), capping addressable memory at ~2 GB. This patch switches to unsigned 32-bit limits (2^32 - 1) to support up to ~4 GB, and drops the nsw (no-signed-wrap) overflow flag on offset arithmetic so the compiler doesn't misoptimize offsets above the signed range.

* Updated buffer size constants for i8 type (2147483646 to 4294967294), OOB index dense constant (2147483647 to 4294967295), validBytes constant (2147483646 to 4294967294). Updated f16, f32, i32 buffer sizes.


ftynse (commented):
The size of the output memref looks completely off, before and after this patch. You may want to find the root cause of that rather than trying to increase the size for it to fit.

PR #1077, opened 2026-03-08
Initial LLVM dialect -> WaveASM translation pipeline

Hardcode84:
Adds a pipeline for translating LLVM dialect IR (produced by the water backend) into WaveASM, assembling it, and running it on GPU. This is a drop-in replacement for the real LLVM backend for simple kernels. No control flow or mfmas yet.

### Pipeline

```
water-opt (lower to LLVM dialect)
  → waveasm-translate --waveasm-translate-from-llvm
  → waveasm-translate --waveasm-gpu-module-to-binary
  → water-opt (water-gpu-to-gpu-runtime)
```

### What's implemented

- **TranslateFromLLVM pass**: Translates `llvm.func` kernels with ROCDL intrinsics into `waveasm.program` ops
  - Scalar LLVM ops: `constant`, `poison`, `add`, `mul`, `sext`, `zext`, `trunc`, `icmp`, `select`
  - ROCDL intrinsics: `workitem.id.x`, `workgroup.id.{x,y,z}`, `make.buffer.rsrc`
  - Buffer operations: `getelementptr` (bare-pointer, buffer, and chained GEPs), `load`, `store`
  - Wide buffer ops: `dwordx2/x3/x4` for vectorized loads/stores
  - SRD setup: prologue generation with kernarg preloading, proper SRD descriptor patching (stride bits, flags from `make.buffer.rsrc`)
- **gpu-module-to-binary pass**: Assembles WaveASM programs into HSACO binaries using `llvm-mc` + `lld`
- **water+waveasm pipeline**: Orchestrates the full flow from water IR to GPU execution
- **Symbol coexistence**: `llvm.func` stays alive alongside `waveasm.program` (with `__waveasm` suffix) for `gpu.launch_func` verification

### Bug fixes

- **Workgroup ID SGPR allocation**: AMDGPU only allocates system SGPRs for enabled workgroup IDs. If only `workgroup_id_y` was used, the hardware placed it at `s[base+0]` but we read `s[base+1]`. Fixed by always enabling all three IDs when any is used (matches real LLVM backend).
- **SRD descriptor corruption**: `s_mov_b64` copies raw pointer bits into SRD word 1, corrupting the stride/swizzle fields. Fixed by clearing upper bits and patching flags from `make.buffer.rsrc`.

Depends on https://github.com/iree-org/wave/pull/1074

### DIFF
--- a/waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
+++ b/waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
@@ -169,7 +169,9 @@ static LogicalResult handlePoison(LLVM::PoisonOp op, LLVMTranslationState &st) {
   auto &builder = ctx.getBuilder();
   auto loc = op.getLoc();
 
-  // Poison is undefined — materialize as zero.
+  // Poison is undefined -- materialize as zero.
+  // TODO: Assumes b32. A legalization pass should ensure all poison values
+  // are i32/f32 before this point.
   auto immTy = ctx.createImmType(0);
   auto imm = ConstantOp::create(builder, loc, immTy, int64_t{0});
   auto vregTy = ctx.createVRegType();
@@ -191,7 +193,8 @@ static LogicalResult handleConstant(LLVM::ConstantOp op,
   else
     return op->emitOpError("unsupported constant type");
 
-  // Materialize as v_mov_b32 into a VGPR.
+  // TODO: Assumes i32. A legalization pass should reject or widen non-b32
+  // constants before this point.
   auto immTy = ctx.createImmType(intVal);
   auto immOp = ConstantOp::create(builder, loc, immTy, intVal);
   auto vregTy = ctx.createVRegType();
@@ -237,7 +240,8 @@ static LogicalResult handleWorkgroupId(OpTy op, LLVMTranslationState &st,
   return success();
 }
 
-// i32<->i64 casts are identity on a 32-bit GPU.
+// TODO: Assumes all integer values fit in 32 bits. A legalization pass
+// should lower i64 arithmetic to pairs of i32 ops before this point.
 static LogicalResult handleSext(LLVM::SExtOp op, LLVMTranslationState &st) {
   st.ctx.getMapper().mapValue(op.getResult(), resolve(op.getOperand(), st.ctx));
   return success();
@@ -261,7 +265,8 @@ static LogicalResult handleICmp(LLVM::ICmpOp op, LLVMTranslationState &st) {
   Value lhs = resolve(op.getLhs(), ctx);
   Value rhs = resolve(op.getRhs(), ctx);
 
-  // V_CMP_* ops set VCC implicitly (no SSA result).
+  // TODO: Assumes i32 operands. A legalization pass should ensure all
+  // integer comparisons are on i32 values before this point.
   using Pred = LLVM::ICmpPredicate;
   switch (op.getPredicate()) {
   case Pred::slt:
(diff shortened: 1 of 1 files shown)


waveasm/include/waveasm/Transforms/Passes.td
```diff
@@ -239,4 +239,45 @@ def WAVEASMMemoryOffsetOpt : Pass<"waveasm-memory-offset-opt"> {
...
+  let description = [{
+    Final pass in the WaveASM pipeline.  Expects gpu.module ops containing
+    register-allocated waveasm.program ops.  Emits AMDGCN assembly, assembles
+    + links to HSACO via LLVM MC, and replaces each gpu.module with a
+    gpu.binary holding the code object.
+  }];
+
+  let options = [
+    Option<"targetArch", "target", "std::string", "\"gfx942\"",
+           "Target GPU architecture">,
```
ftynse: Upstream has a better alternative with the target attribute on GPU modules, we should consider using that instead. Especially when this flag has to be repeated for multiple passes, allowing for mismatches.

waveasm/lib/Transforms/GPUModuleToBinary.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+    module.walk([&](waveasm::ProgramOp program) {
+      if (failed(waveasm::writeAssembly(program, mapping, asmStream)))
+        signalPassFailure();
+    });
+
+    if (asmText.empty()) {
+      module.emitError("no assembly generated");
+      return signalPassFailure();
+    }
```
ftynse: Does this happen? Or is this an intercept of the error above, at which point it should happen there. Otherwise it is entirely possible that no assembly was generated for one program, but something was generated for another program, and the overall result is not empty.

waveasm/lib/Transforms/GPUModuleToBinary.cpp
```diff
@@ -0,0 +1,132 @@
...
+    OpBuilder builder(module.getContext());
+    StringAttr binaryAttr =
+        builder.getStringAttr(StringRef(hsaco->data(), hsaco->size()));
+
+    for (auto gpuModule : gpuModules) {
+      builder.setInsertionPointAfter(gpuModule);
+
+      Attribute target;
+      if (gpuModule.getTargetsAttr() && !gpuModule.getTargetsAttr().empty())
+        target = gpuModule.getTargetsAttr()[0];
```
ftynse: Report a warning that extra targets were ignored.

waveasm/lib/Transforms/TranslateFromLLVM.cpp
```diff
@@ -0,0 +1,672 @@
...
+  // Store the original kernel name for assembly emission.
+  std::string programName = (func.getName() + "__waveasm").str();
+  auto program =
+      ProgramOp::create(builder, loc, programName, targetAttr, abiAttr,
+                        /*vgprs=*/int64_t{256},
+                        /*sgprs=*/int64_t{104},
+                        /*workgroup_size=*/builder.getArrayAttr(sizes),
+                        /*lds_size=*/IntegerAttr{});
+
+  program->setAttr("kernel_name", builder.getStringAttr(func.getName()));
```
ftynse: Already mentioned elsewhere about string attribute names. Now it starts looking like this should have been an optional attribute on the op.

waveasm/lib/Transforms/TranslateFromLLVM.cpp
```diff
@@ -0,0 +1,672 @@
...
+
+  if (program.getBody().empty())
+    program.getBody().emplaceBlock();
```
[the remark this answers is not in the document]
Hardcode84: we can assert instead of doing the check but whatewer

Hardcode84:
closing in favor of https://github.com/iree-org/wave/pull/1078

PR #1078, opened 2026-03-08
Arithmetic legalization for WaveASM backend

Hardcode84:
Introduces a proper legalization stage between LLVM translation and optimization in the WaveASM pipeline. Instead of making register file (SGPR/VGPR) and width (i32/i64) decisions during translation, we now emit generic pseudo-ops and legalize them in a dedicated pass.

### What changed

**Design doc** (`waveasm/docs/legalization.md`)
- Documents the three-phase legalization strategy (type → register → instruction selection).
- Compares our approach with the real LLVM AMDGPU backend.

**Generic arithmetic pseudo-ops** (`WaveASMArithOps.td`, `WaveASMTypes.td`)
- 7 new ops: `arith.add`, `arith.mul`, `arith.cmp`, `arith.select`, `arith.trunc`, `arith.sext`, `arith.zext`.
- Single `arith.cmp` op with `CmpPredicate` enum (10 predicates) instead of per-predicate ops.
- Register-file-agnostic: accept any operand type, produce any register type.

**Arithmetic legalization pass** (`ArithLegalization.cpp`)
- Demand-driven divergence: VGPR operand → VALU path; all SGPR/imm → SALU path.
- Constant bus enforcement: inserts `v_mov_b32` when VALU has two SGPR operands.
- i64 narrowing: wide precolored SGPRs narrowed to low sub-register.
- Walk-based (not greedy rewriter) to avoid DCE of pure ops with no uses.

**LLVM translation updated** (`TranslateFromLLVM.cpp`)
- `handleAdd`, `handleMul`, `handleICmp`, `handleSelect`, `handleTrunc`, `handleSext`, `handleZext` now emit pseudo-ops.
- Removed `ensure32Bit`/`ensureVGPR` helpers — legalization pass handles this.

**Pipeline** (`water.py`)
- Added `--waveasm-arith-legalization` after `--waveasm-translate-from-llvm`.

**Also includes** (from prior commits on this branch):
- Scalar kernel arg support and dynamic dims in LLVM→WaveASM translation.
- Copy kernel e2e test covering the full WaveASM pipeline.

depends on https://github.com/iree-org/wave/pull/1077

### DIFF
--- a/waveasm/include/waveasm/Transforms/Passes.td
+++ b/waveasm/include/waveasm/Transforms/Passes.td
@@ -331,7 +331,7 @@ def WAVEASMTranslateFromLLVM : Pass<"waveasm-translate-from-llvm"> {
   }];
 
   let options = [
-    Option<"targetArch", "target", "std::string", "\"gfx942\"",
+    Option<"targetArch", "target", "std::string", "\"gfx950\"",
            "Target GPU architecture">
   ];
 
--- a/waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
+++ b/waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
@@ -678,6 +678,11 @@ static LogicalResult translateLLVMModule(Operation *rootOp,
   if (!target)
     return rootOp->emitError() << "unknown target: " << targetId;
 
+  if (!isa<GFX950TargetAttr>(target))
+    return rootOp->emitError()
+           << "LLVM->WaveASM translation only supports gfx950, got "
+           << targetId;
+
   SmallVector<LLVM::LLVMFuncOp> kernels;
   rootOp->walk([&](LLVM::LLVMFuncOp func) {
     if (func->hasAttr("gpu.kernel") || func->hasAttr("rocdl.kernel"))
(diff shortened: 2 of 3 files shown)


waveasm/test/Transforms/arith-legalization.mlir
```diff
@@ -111,20 +111,215 @@ waveasm.program @test_wide_narrowing
...
+  %b = waveasm.precolored.vreg 1 : !waveasm.vreg
+  %va = waveasm.pack %a, %a : (!waveasm.vreg, !waveasm.vreg) -> !waveasm.vreg<2>
+  %vb = waveasm.pack %b, %b : (!waveasm.vreg, !waveasm.vreg) -> !waveasm.vreg<2>
+
+  // VALU schoolbook: mul_lo, mul_hi, 2x mul_lo, v_add3_u32.
+  // CHECK: waveasm.v_mul_lo_u32
+  // CHECK: waveasm.v_mul_hi_u32
+  // CHECK: waveasm.v_mul_lo_u32
+  // CHECK: waveasm.v_mul_lo_u32
+  // CHECK: waveasm.v_add3_u32
```
ftynse: Would be reasonable to check that the right values go into the operations

PR #1080, opened 2026-03-08
[asm] Replace RawOp string hacks with typed SALUPhys ops for physical…

harsh-nod:
… register writes

RawOp was used to emit s_mov_b32/s_mov_b64/s_and_b32/s_or_b32 as raw assembly strings because the Pure trait on SALUUnaryOp/SALUBinaryOp causes DCE to eliminate ops whose SSA results have no consumer — which is exactly the case for SRD setup writes to physical registers.

Introduce SALUPhysUnaryOp and SALUPhysBinaryOp base classes that use the SpecialRegOp trait (non-Pure, prevents DCE/CSE) and take the destination physical register as an input operand rather than an SSA result. Add four concrete ops: S_MOV_B32_PHYS, S_MOV_B64_PHYS, S_AND_B32_PHYS, S_OR_B32_PHYS.

Replace 15 RawOp::create calls across emitSRDPrologue(), handleVectorStore() SRD adjustment, and handleFatRawBufferCast() with the new typed ops. Assembly output is identical — same mnemonics, operands, and instruction count.

Three RawOps remain for s_branch, .p2align, and label directives which have no typed op equivalent.

Add lit test verifying SALUPhys ops survive CSE while Pure variants are still eliminated.

### DIFF
--- a/waveasm/include/waveasm/Dialect/WaveASMOps.td
+++ b/waveasm/include/waveasm/Dialect/WaveASMOps.td
@@ -129,6 +129,17 @@ class SALUCmpOp<string mnemonic, list<Trait> traits = []>
   let assemblyFormat = "$src0 `,` $src1 attr-dict `:` type($src0) `,` type($src1) `->` type($result)";
 }
 
+// SALU Physical-Register Write (unary): write src to a physical SGPR dst.
+// The destination is an input operand (a PrecoloredSReg reference), not an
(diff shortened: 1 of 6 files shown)


Hardcode84:
Actually, you can just introduce something like `compose_reg` op, which takes multiple individual reg operands and returns the wide one. The op itself can be pure and also you can get rid of `newSrdBase` and delegate sgpr selection to regalloc completely.

PR #1084, opened 2026-03-09
Sort lists in cmake files in pre-commit

Hardcode84:
They were hurting my aesthetic feelings for way too long.


kuhar (commented):
Can you land the cleanup separately from the pre-commit hook?

ftynse (approved):
LGTM. Looks like useful hook elsewhere, but perhaps with a better repo name.

PR #1091, opened 2026-03-10
Dynamic stride support through waveasm

suryajasper:
This PR adds support for dynamic strides through the waveasm backend.

# 4 Cases for Dynamic Strides
There are 4 main cases that needed to be addressed to ensure complete support.

## Case 1: waveasm + dynamic strides

The loads were already using correct static offsets. The only issue was `num_records` being too small for non-contiguous layouts, causing the hardware to return zero for valid accesses. Fix: `computeBufferSizeFromMemRef`.

## Case 2: waveasm + dynamic strides + dynamic dims

Before the loads could use dynamic strides, the runtime stride values had to actually *reach* the kernel. This is where the cascade of prologue issues hit:

1. **Preload SGPR limit exceeded** — the 9 arguments couldn't all be preloaded, so the kernel got garbage values. Fix: partial preloading strategy.
2. **Scalar reservations DCE'd** — even after loading remaining scalars via `s_load_dword`, the register allocator reused those SGPRs. Fix: `scalar_sgpr_base`/`scalar_sgpr_count` attributes + explicit reservation in `LinearScanPass`.
3. **Mask ops aborted translation** — `vector<Nxindex>` arithmetic in the loop body caused a hard failure. Fix: non-fatal loop body failures in `RegionBuilder`.

## Case 3: waveasm + dynamic strides + buffer ops

Even with correct stride math, the loads were broken because:

1. **RawOps dropped in loop bodies** — the SRD fill instructions inside loops were silently discarded. Fix: `AssemblyEmitter` handles `RawOp` in nested regions.
2. **SRD adjustment inside the loop** — lazy emission placed it in the loop body; LICM partially hoisted it, corrupting registers on iteration > 0. Fix: eager adjustment in `handleFatRawBufferCast`.
3. **`handleMemRefLoad` hardcoded the wrong SRD** — it used `s[8:11]` regardless of which buffer was being accessed, and fell back to literal `0` for voffset. Fix: rewrite to use `lookupSRD` + `computeVOffsetFromIndices`.

SRD construction throughout (`emitSRDPrologue`, `emitSRDBaseAdjustment`) now uses origin/main's `PackOp`-based pattern: `ExtractOp` to decompose the source SRD, typed SALU ops (`S_ADD_U32`/`S_ADDC_U32`, `S_MOV_B32`/`S_MOV_B64`) on virtual registers, and `PackOp` to reassemble — letting the register allocator handle physical assignment. Prologue SRD fills use `DCEProtectOp` to prevent elimination. This eliminates all manual SGPR index arithmetic (`getNextSwizzleSRDIndex`) and `RawOp` string manipulation for SRD words.

## Case 4: waveasm + dynamic strides + dynamic dims + buffer ops

The Python codegen emitted per-element `memref.load` with `vector<8xindex>` OOB-index-selection, which waveasm has no handlers for. The loads were dead on arrival. Fix: allow `vector.maskedload` with buffer ops on the waveasm backend in `read_write.py`.

# Full Picture

```
Frontend generates translatable IR          → Case 4 fix (read_write.py)
  ↓
Kernel arguments reach the GPU correctly    → Case 2 fixes (preloading, SGPR reservation)
  ↓
SRD prologue fills correct registers        → Case 2 fixes (typed ops + DCEProtectOp)
  ↓
SRD base adjustment builds SSA SRD value    → Case 3 fixes (ExtractOp + SALU + PackOp)
  ↓
Loads/stores use correct SRD + voffset      → Case 3 fix (lookupSRD checks getSRDValue first)
  ↓
num_records doesn't falsely reject accesses → Case 1 fix (computeBufferSizeFromMemRef)
```

### DIFF
--- a/lit_tests/kernel/wave/gemm.py
+++ b/lit_tests/kernel/wave/gemm.py
@@ -2510,9 +2510,11 @@ def test_persistent_gemm():
     # CHECK-DAG:       %[[C304_I32:.+]] = arith.constant 304 : i32
     # CHECK-DAG:       %[[C128:.+]] = arith.constant 128 : index
     # CHECK:           %[[BLOCK_ID_X:.+]] = gpu.block_id  x upper_bound 304
+    # CHECK:           %[[CAST_I32:.+]] = arith.index_cast %[[BLOCK_ID_X]] : index to i32
+    # CHECK:           %[[CAST_IDX:.+]] = arith.index_cast %[[CAST_I32]] : i32 to index
 
     # Persistent loop
-    # CHECK:           %{{.*}} = scf.while (%[[ARG3:.+]] = %[[BLOCK_ID_X]]) : (index) -> index {
+    # CHECK:           %{{.*}} = scf.while (%[[ARG3:.+]] = %[[CAST_IDX]]) : (index) -> index {
     # CHECK:             %{{.*}} = arith.cmpi slt, %[[ARG3]], %[[C128]] : index
     # CHECK:             scf.condition(%{{.*}}) %[[ARG3]] : index
     # CHECK:           } do {
--- a/requirements-iree-pinned.txt
+++ b/requirements-iree-pinned.txt
@@ -7,5 +7,5 @@
 # Uncomment to skip versions from PyPI (so _only_ nightly versions).
 # --no-index
 
-iree-base-compiler==3.12.0rc20260407
-iree-base-runtime==3.12.0rc20260407
+iree-base-compiler==3.12.0rc20260408
+iree-base-runtime==3.12.0rc20260408


waveasm/lib/Transforms/handlers/HandlerUtils.cpp
```diff
@@ -62,14 +62,33 @@ int64_t getElementBytes(Type type) {
...
+        if (shape[i] > 1)
+          maxOffset += (shape[i] - 1) * std::abs(strides[i]);
+      }
+      return (maxOffset + 1) * elemBytes;
+    }
+  }
+  // Dynamic dims or dynamic strides — use a large-but-bounded value so that
+  // hardware OOB checking catches wild offsets (e.g., negative wraparound)
+  // and returns zero rather than faulting on unmapped memory.
+  return 0x20000000;
```
harsh-nod: this return value of 512MB is too small and will silently zero large buffers

tests/kernel/dynamic_strides_test.py
```diff
@@ -29,17 +30,27 @@ def _compile_gemm(shape, dynamic_dims=False, buffer_ops=False):
         dynamic_symbols=dynamic_symbols,
         wave_runtime=True,
         use_buffer_ops=buffer_ops,
+        backend="asm" if use_waveasm else "llvm",
     )
     options = set_default_run_config(options)
     return wave_compile(options, gemm), options
 
 
+def _xfail_waveasm_dynamic_strides(
```
harsh-nod: is this dead code?

harsh-nod (changes requested):
I'll start by gathering the PR data, existing review outputs, and source files to produce a thorough synthesis.Now let me read the key source files to verify the findings and cross-reference the reviewer outputs.Now let me check the key lines in `TranslateFromMLIR.cpp` and `HandlerUtils.cpp` for the critical findings.Let me verify a few more critical areas before writing the synthesis.Now I have all the data needed. Let me write the comprehensive synthesis.The consensus review synthesis has been written to `output.md`. Here's the summary:

**Verdict: REQUEST CHANGES** (confidence: 88/100)

**2 Critical Fixes Required (P0):**
1. **C1**: `0x20000000` fallback in `HandlerUtils.cpp:91` is too small — must revert to `0x7FFFFFFE`. Unanimous across all reviewers + harsh-nod.
2. **C2**: `handleVectorMaskedLoad` at `TranslateFromMLIR.cpp:1257` is missing the `getPendingSRDBaseAdjust` check that all three sibling handlers have. Will cause reads from wrong memory region when maskedload is the first consumer of a base-adjusted memref.

**2 Cleanup Items (P2, per maintainer):**
3. **C3**: Dead `_xfail_waveasm_dynamic_strides` function + unused `import pytest`
4. **C4**: Stale "LLVM backend" comment in `compile_options.py`

**2 Strongly Recommended (P1):**
5. **C6**: Scope the silent loop-body skip to specific op types or add `emitWarning()`
6. **NC1**: Assert passthrough is zero in masked load handler

**Key Resolved Disagreements:**
- Passthrough removal: downgraded from HIGH to P2 — hardware OOB zeroing provides the safety net
- GFX950 overflow SRD: all agree it's real but pre-existing — does not block this PR
- Silent skip pattern: kept at P1 — pragmatically needed but should be scoped tighter

PR #1094, opened 2026-03-10
Add MLIR verifier to check HardwareConstraint.vector_shapes consistency with tile sizes

### LINKED ISSUE #864: Add MLIR verifier to check HardwareConstraint.vector_shapes consistency with tile sizes
The `HardwareConstraint` attribute may carry an explicit `vector_shapes` dictionary mapping dimension symbols to concrete sizes. These values should be consistent with the resolved tile sizes from the corresponding `WorkgroupConstraint` and `TilingConstraint` entries on the same dimensions.

Currently there is no MLIR-side verification for this invariant. If the two diverge (e.g., `vector_shapes` says `M: 16` but the `WorkgroupConstraint` for `M` resolves to `32`), the mismatch is not caught until runtime or leads to silent miscompilation.

The existing `verifyConstraints` function in `WaveDialect.cpp` already evaluates constraint tile sizes against hyperparameters. It could be extended to cross-check resolved tile sizes against `vector_shapes` entries when both are present, emitting a diagnostic on mismatch.

tgymnich:
- **change waves_per_block to DenseI32ArrayAttr to allow to use struct(params) in assemblyFormat**
- **verify vector_shapes consistency with WorkgroupConstraint and TilingConstraint**

fixes #864

### DIFF
--- a/water/lib/Dialect/Wave/IR/WaveDialect.cpp
+++ b/water/lib/Dialect/Wave/IR/WaveDialect.cpp
@@ -222,6 +222,7 @@ verifyConstraints(ArrayAttr constraints,
   // * The number of workgroups should be greater than or equal to one.
   llvm::SmallDenseMap<wave::WaveSymbolAttr, int64_t> resolvedWorkgroupSizes(
       workgroupConstraints.size());
+  llvm::SmallDenseMap<wave::WaveSymbolAttr, int64_t> resolvedSizes;
   llvm::SmallDenseSet<wave::WaveWorkgroupDimAttr, 4> assignedDims;
   llvm::SmallDenseSet<wave::WaveWorkgroupDimAttr, 4> needsPrimaryDim;
   for (auto &&[symbol, constraint] : workgroupConstraints) {
@@ -250,6 +251,7 @@ verifyConstraints(ArrayAttr constraints,
 
     int64_t workgroupSize = evaluated->front();
     resolvedWorkgroupSizes[symbol] = workgroupSize;
+    resolvedSizes[symbol] = workgroupSize;
 
     std::optional<llvm::SmallVector<int64_t>> resolvedDims =
         wave::resolveSymbolNames(symbol, hyperparams);
@@ -310,6 +312,7 @@ verifyConstraints(ArrayAttr constraints,
                          << workgroupSize << " for dimension: " << symbol;
     }
     resolvedWaveCounts[symbol] = numWaves;
+    resolvedSizes[symbol] = resolvedWaveSize;
   }
 
   // verify consistency between wave constraints and waves_per_block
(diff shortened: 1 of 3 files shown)


docs/ir_design.rst
```diff
@@ -0,0 +1,159 @@
...
+In MLIR, ``vector_shapes`` entries must all be ``IntegerAttr`` values.  The
+verifier in ``WaveDialect.cpp`` enforces this.
+
+
+The special value 0
+^^^^^^^^^^^^^^^^^^^
+
+A vector shape of ``0`` marks a dimension as *scalar* — the wave does not tile
+along it.  This is used for dimensions like batch (``B``) that should not
+contribute to the intra-wave data distribution:
```
ftynse: I don't understand the "scalar" part even though it seems to be present in the comments. The fact that a dimension isn't tiled doesn't make it scalar, it is just going to be replicated/unrolled by expansion. What deserves an investigation here is whether one is allowed to set `0` as vector shape for a symbol that is mapped to blocks/threads and, symmetrically, whether one is allowed to set a non-zero as vector shape for a dimension that isn't mapped.

docs/ir_design.rst
```diff
@@ -0,0 +1,159 @@
...
+Relationship to workgroup and tiling constraints
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``vector_shapes`` and constraint tile sizes serve different purposes:
+
+* **Tile size** (from ``#wave.workgroup_constraint`` or
+  ``#wave.tiling_constraint``) is the total amount of work assigned to one
+  workgroup or one iteration of a reduction loop along a dimension.
+* **Vector shape** is the amount of work one wave handles in a single instance
+  of an operation along that dimension.
```
ftynse: What about wave constraints? It looks like they _also_ specify wave tile size. My (potentially incorrect) undesrtanding is that `vector_shape` and wave constraints are redundant, but I am easily convinced otherwise.

docs/ir_design.rst
```diff
@@ -0,0 +1,159 @@
...
+
+**When mma_type is present,** the vector shapes derive from the MMA geometry
+and are typically smaller than the constraint tile sizes.  The expansion pass
+(which runs on the Python/FX side) replicates each
+operation to cover the tile.  For example, with ``BLOCK_M = 64``,
```
ftynse: which of the tile sizes? :)

docs/ir_design.rst
```diff
@@ -0,0 +1,159 @@
...
+``waves_per_block = [2, 2, 1]``, and ``mma_type = f32_16x16x16_f16``
+(vector shape 16 for M):
+
+.. code-block:: text
+
+   expansion_count = ceil(64 / (2 × 16)) = 2
+
+The MLIR IR only sees the already-expanded result: two ``wave.mma`` ops along M
```
ftynse: This is not exactly true since it depends on where the conversion happens.

docs/ir_design.rst
```diff
@@ -0,0 +1,159 @@
...
+rather than one.  The ``vector_shapes`` remain on the
+``#wave.hardware_constraint`` for verification and for passes that need to
+reason about the per-wave tile.
+
+**When mma_type is absent,** the MLIR verifier enforces that each
+``vector_shapes`` entry **matches** the resolved tile size from the
```
ftynse: I'm not convinced that vector shapes must match _workgroup_ tile size. Not saying I'm right, but please convince me one way or another.

PR #1100, opened 2026-03-10
Add IfOp else-branch type coercion in waveasm translation

willghatch:
When an scf.if has a then branch that produces register values and an else branch that yields constants (which become immediates), the waveasm.if verifier requires both branches to yield type-compatible values. This implements the coercion and tests it.

### DIFF
--- a/waveasm/lib/Transforms/RegionBuilder.cpp
+++ b/waveasm/lib/Transforms/RegionBuilder.cpp
@@ -13,6 +13,35 @@
 using namespace mlir;
 using namespace waveasm;
 
+/// Insert a move instruction to coerce `src` into a register of the target
+/// type's register class.  Returns the coerced value, or a null Value if the
+/// target type is not a recognized register type.
+static Value coerceToRegType(OpBuilder &builder, Location loc,
+                             TranslationContext &ctx, Type targetType,
+                             Value src) {
+  auto regClass = getRegClass(targetType);
+  if (!regClass)
+    return {};
+  int64_t size = getRegSize(targetType);
+  Type regType;
+  switch (*regClass) {
+  case RegClass::AGPR:
+    regType = ctx.createARegType(size);
+    break;
+  case RegClass::VGPR:
+    regType = ctx.createVRegType(size);
+    break;
+  case RegClass::SGPR:
+    regType = ctx.createSRegType(size);
+    break;
+  default:
+    llvm_unreachable("unhandled register class");
+  }
+  if (*regClass == RegClass::SGPR)
+    return S_MOV_B32::create(builder, loc, regType, src);
+  return V_MOV_B32::create(builder, loc, regType, src);
+}
+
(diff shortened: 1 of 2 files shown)


waveasm/lib/Transforms/RegionBuilder.cpp
```diff
@@ -368,17 +377,43 @@ IfOp RegionBuilder::buildIfFromSCFIf(scf::IfOp ifOp) {
...
-    YieldOp::create(builder, loc, yieldVals);
+    // Coerce else-yield values to match then-yield types.  The then branch
+    // may produce AGPR/VGPR accumulators while the else branch yields bare
+    // immediates (e.g. zero constants).  The waveasm.if verifier requires
+    // both branches to yield type-compatible values.
+    for (unsigned i = 0; i < elseYieldVals.size(); ++i) {
+      if (i >= thenYieldVals.size())
+        break;
+      Type thenType = thenYieldVals[i].getType();
+      Type elseType = elseYieldVals[i].getType();
```
harsh-nod: Only else-to-then coercion is handled. what about then to else coercion?

waveasm/lib/Transforms/RegionBuilder.cpp
```diff
@@ -340,17 +342,24 @@ IfOp RegionBuilder::buildIfFromSCFIf(scf::IfOp ifOp) {
...
       }
     }
 
-    YieldOp::create(builder, loc, yieldVals);
+    YieldOp::create(builder, loc, thenYieldVals);
+
+    // Update the if-op result types to match actual then-yield types, since
+    // the initial result types were guessed before translating the then body.
+    for (unsigned i = 0; i < waveIfOp->getNumResults(); ++i) {
+      if (i < thenYieldVals.size()) {
```
ftynse: Is this not guaranteed by verifiers?

waveasm/lib/Transforms/RegionBuilder.cpp
```diff
@@ -386,33 +401,64 @@ IfOp RegionBuilder::buildIfFromSCFIf(scf::IfOp ifOp) {
...
+        continue;
+      // Coerce the immediate side to the register side.
+      if (getRegClass(thenType)) {
+        OpBuilder::InsertionGuard guard(builder);
+        builder.setInsertionPointToEnd(waveIfOp.getElseBlock());
+        Value coerced =
+            coerceToRegType(builder, loc, ctx, thenType, elseYieldVals[i]);
+        if (!coerced) {
+          ifOp.emitError("unsupported type coercion for yield result ")
+              << i << ": then type " << thenType << ", else type " << elseType;
```
ftynse: The common practice in mlir-based projects is to have a test for each user-visible diagnostic to make sure they are visible / consistent. Sometimes it helps identify diagnostics that cannot be triggered by valid inputs, which allows them to be converted to assertions and decrease compile time in release mode.

PR #1120, opened 2026-03-12
Introduce canonical Wave region formats and annotated pass adapters

martin-luecke:
## Summary

This is a step towards a canonical form for Wave FX graphs. This change only tackles region structure for now, not all graph structure more broadly.

The canonical region form chosen here is an isolated region representation: nested regions do not directly reference outer values, and all captures are represented explicitly as lifted `Placeholder`s at the start of the region signature. This gives us one stable structural form to target for verification, printing, and roundtrips.

We are introducing this gradually. Not all existing passes operate on that canonical region form yet, and this change does not require rewriting them all at once. Instead, passes can now declare which region form they expect, and the pass boundary will temporarily adapt the graph into that form before the pass runs and return to canonical form afterwards in the normal pipeline.

This change introduces four region forms:
- `ISOLATED`: the canonical form, with explicit lifted placeholder captures
- `LEGACY_PLACEHOLDERS`: the older placeholder-based capture form expected by several existing passes
- `DIRECT_OUTER_REF`: a legacy form where nested regions directly reference outer values
- `SCHEDULE_SIGNATURE_PLACEHOLDERS`: a hybrid form that keeps placeholders only for schedule-signature sources

## What this PR does
- introduce canonical region-capture handling and verification for nested Wave regions
- let passes declare their required region form directly via annotation through a decorator
- adapt the pipeline to canonicalize at pass boundaries while still supporting legacy region forms where needed
- update lit tests to request raw post-pass output explicitly at the call site

## Additional Change
`minimize_shared_allocs` now skips dead shared allocs instead of assuming every alloc still has a live first/last use. This became necessary once the new canonical-region migration flow exposed intermediate graphs where an alloc may survive temporarily after its uses are gone, making an old pass assumption explicit.

### DIFF
--- a/wave_lang/kernel/wave/minimize_global_loads.py
+++ b/wave_lang/kernel/wave/minimize_global_loads.py
@@ -320,6 +320,7 @@ def update_shared_memory_read(
     if custom_memory_shape != metadata.memory_shape:
         permutation = [custom_memory_shape.index(k) for k in metadata.memory_shape]
         custom_memory.update_arg("shape", metadata.memory_shape)
+        custom_memory.fx_node.type = custom_memory.type
         new_distributed_shape = []
         for i, perm in enumerate(permutation):
             offset = 0


wave_lang/kernel/wave/region_canonicalization.py
```diff
@@ -0,0 +1,528 @@
...
+        outer_source = region.get_outer_node(implicit_capture)
+        if outer_source.name == local_capture.name:
+            return outer_source
+
+    # Legacy placeholder-only regions may lose explicit outer-source links.
+    # When name matching also fails, fall back to capture-signature position.
+    if capture_idx < len(region.implicit_captures):
+        return region.get_outer_node(region.implicit_captures[capture_idx])
+
+    return region.get_outer_node(source)
```
ftynse: Is it safe to fall back to this?

wave_lang/kernel/ops/wave_ops.py
```diff
@@ -2313,32 +2366,219 @@ def is_contiguous_vec(self, constraints, target: str) -> bool:
...
+        graph: fx.Graph,
+        outer_node: fx.Node | CustomOp,
+        location: Optional[CapturedLocation] = None,
+    ) -> fx.Node:
+        """Insert a lifted placeholder that represents `outer_node` in `graph`."""
+        outer_node = cls.capture_source(outer_node)
+        placeholder = Placeholder(outer_node.name, outer_node.type)
+        with graph.inserting_after(cls._last_region_input_or_root(graph)):
+            placeholder_node = placeholder.add_to_graph(graph, loc=location)
+        if placeholder_node.name != outer_node.name:
```
ftynse: A comment that `add_to_graph` may force renaming would be useful here since we explicitly set the name 3 lines above.

wave_lang/kernel/ops/wave_ops.py
```diff
@@ -2313,32 +2366,219 @@ def is_contiguous_vec(self, constraints, target: str) -> bool:
...
+                    node
+                    for node in graph.nodes
+                    if node is not placeholder_node and node.name == outer_node.name
+                ),
+                None,
+            )
+            if conflicting_node is not None:
+                conflicting_node.name = placeholder_node.name
+            placeholder_node.name = outer_node.name
+        placeholder_node.type = outer_node.type
```
ftynse: Is this needed given that we passed in the type to `Placeholder` constructor?

wave_lang/kernel/ops/wave_ops.py
```diff
@@ -2313,32 +2366,219 @@ def is_contiguous_vec(self, constraints, target: str) -> bool:
...
+                    f"Cycle detected while resolving lifted capture source for {outer_node}"
+                )
+            seen.add(outer_node)
             outer_node = outer_node.meta["lifted"]
         return outer_node
 
+    def _build_capture_lookup(
+        self, graph: fx.Graph
+    ) -> NestedRegionOp._CaptureLookupSnapshot:
+        """Snapshot current outer-to-local capture bindings for `graph`."""
```
ftynse: Given that `by_outer` is a 1:1 dict, the logic below will capture only the first inner node "lifting" the outer node. This is the canonical node we choose elsewhere as well, but may be worth documenting this behavior and/or reusing the logic elsewhere so we don't accidentally diverge.

docs/wave/canonical_ir_format.md
```diff
@@ -0,0 +1,168 @@
...
+### `LEGACY_PLACEHOLDERS`
+
+This is the older placeholder-based capture form still expected by some
+pre-existing passes.
+
+Structural properties:
+
+- captured outer values are represented by region-local placeholders
+- the mapping from a local placeholder back to its outer source may still be
+  recovered with legacy heuristics, such as placeholder name or placeholder
```
ftynse: are these heuristics described somewhere we can point to?

lit_tests/kernel/wave/hoisting.py
```diff
@@ -130,4 +130,4 @@ def loop(acc: tkl.Register[M, N, tkl.f32]):
     # CHECK: %read_2_M:0_N:0_K:0
     # CHECK: %add_M:0_N:0_K:0
     # CHECK: %mma_M:0_N:0_K:0
-    # CHECK: %add_1_M:0_N:0_K:0
+    # CHECK: return [add_1_M:0_N:0_K:0]
```
ftynse: where did the definition of %add_1 go?

lit_tests/kernel/wave/index_sequence_analysis.py
```diff
@@ -340,7 +344,7 @@ def test_gemm():
     # CHECK-NEXT: read(memory=b, elements_per_thread=8,
     # CHECK-SAME: index={N: $WG1*BLOCK_N + BLOCK_N/2 + Mod(16*$T1 + 32*$T2 + floor($T0/8) + 32, 64) : 1 : 1, K: ARGK*BLOCK_K + 8*(Mod($T0, 8)) : 8 : 1})
     # CHECK-NEXT: write(register_=read_40, memory=allocate_1, elements_per_thread=8,
-    # CHECK-SMAE: index={N: BLOCK_N/2 + Mod(16*$T1 + 32*$T2 + floor($T0/8) + 32, 64), K: 8*(Mod($T0, 8)) : 8 : 1})
+    # CHECK-SAME: index={N: BLOCK_N/2 + Mod(16*$T1 + 32*$T2 + floor($T0/8) + 32, 64) : 1 : 1, K: 8*(Mod($T0, 8)) : 8 : 1})
```
ftynse: 🤦 this is what one gets for not using proper filecheck that would have warned about misspelled inputs

lit_tests/kernel/wave/expansion.py
```diff
@@ -78,10 +78,10 @@ def test_read_write_equal_sizes():
...
         graph = read_write_same_size()
         idxc.finalize()
-        add_get_results(graph)
-        infer_types(graph)
-        set_node_indices(graph, constraints)
-        expand_graph(graph, constraints)
+        add_get_results(graph, canonicalize_output=False)
+        infer_types(graph, canonicalize_output=False)
+        set_node_indices(graph, constraints, canonicalize_output=False)
+        expand_graph(graph, constraints, canonicalize_output=False)
```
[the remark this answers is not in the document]
martin-luecke: Yes, but that's a rather big refactor that should touch most of the tests.
Let's have this in a follow-up eventually

wave_lang/kernel/ops/wave_ops.py
```diff
@@ -951,12 +963,70 @@ def replacement_location_propagate(
...
+
+            fx.map_arg((user.args, user.kwargs), visit)
+            return found
+
+        for user in list(self.fx_node.users):
+            if graph is not None and user.graph is not graph:
+                continue
+            user._update_args_kwargs(
+                fx.map_arg(
+                    user.args, lambda arg: new_node if arg is self.fx_node else arg
```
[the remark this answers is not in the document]
martin-luecke: Also added some testing code for this functionality.

docs/wave/canonical_ir_format.md
martin-luecke: there's no documentation of these heuristics. They're ad-hoc patterns that evolved in the codebase.
Clarified the text

docs/wave/canonical_ir_format.md
```diff
@@ -0,0 +1,168 @@
...
+
+- nested region bodies do not directly reference outer graph nodes
+- every captured outer value used inside the region is represented by a region-local
+  `Placeholder`
+- these capture placeholders form the leading non-`IterArg` input prefix of the
+  subgraph
+- the parent `NestedRegionOp.implicit_captures` list is the authoritative
+  ordered capture signature
+- each local capture placeholder resolves to exactly one outer source via
+  `meta["lifted"]`
```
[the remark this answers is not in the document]
martin-luecke: In the current design both are needed because they live on different objects and serve different access patterns: implicit_captures is on the parent region op and defines the authoritative ordered capture signature.
meta["lifted"] is on each placeholder node and provides O(1) lookup of its outer source. Without meta["lifted"], resolving a placeholder's source would require finding its position in the prefix and indexing into implicit_captures. The verifier (verify_canonical_region_captures) enforces consistency between the two.

Though it is not run automatically yet.

PR #1126, opened 2026-03-13
[wip] [asm] Support for MT 224x160x256

xintin:
Enables block size 224x160x256 with wave_shape=(2,2) in the waveasm backend for shape `4*MT0, 4*MT1, 8192`.

BLOCK_M=224 could not run because three independent bugs were triggered:
1. Magic-number division bug
BLOCK_M=224 triggers the "add form" of Hacker's Delight unsigned division. LLVM's PostShift already incorporates a -1 reduction for this form, but the waveasm emitter subtracted 1 again, making floordiv 224 compute floordiv 112. This corrupted every LDS address that wraps around BLOCK_M.
2. Odd M-tile count
224/2/16 = 7 MFMA tiles along M. The schedule's partition_by_dim required even divisibility and would assert on 7 tiles split into 2 partitions. Replaced with a divmod-based scheme (4+3 split). Also clamped interleave offsets so they don't exceed the smaller partition size.
3. Negative constant bit-range tracking
BitRange::fromConstant used a signed loop (while (v>0)) that failed for large unsigned magic multipliers. This caused the OR-for-add optimization to miscompute bit ranges, producing wrong addresses for large problem shapes.

### DIFF
--- a/waveasm/test/Transforms/floor-ops.mlir
+++ b/waveasm/test/Transforms/floor-ops.mlir
@@ -24,7 +24,8 @@ func.func @div_by_7(%arg0: i32) -> i32 {
   // CHECK: waveasm.v_sub_u32
   // CHECK: waveasm.v_lshrrev_b32
   // CHECK: waveasm.v_add_u32
-  // CHECK: waveasm.v_lshrrev_b32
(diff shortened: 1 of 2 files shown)


willghatch (approved):
I've pushed an improvement to the tests -- they updated comments but were missing actual checks.

PR #1127, opened 2026-03-13
[wip][asm] Fix large shapes for MTs

xintin:
For large shapes, the bug was in `waveasm/include/waveasm/Transforms/TranslateFromMLIR.h`, in the `BitRange::fromConstant` method. When given a negative constant like -16, the while (v > 0) loop never executes (since -16 < 0), returning BitRange(0, -1), effectively empty range.

Wave compiler generates a workgroup stagger pattern for multi-workgroup kernels. It computes a flat workgroup ID (block_id_y * 4 + block_id_x) and uses an affine.apply map with the expression (flat_id floordiv 8) - 16 to compute the M-tile index for the "remainder" workgroups.

waveasm AffineHandlers.cpp compiles Add(lhs, rhs) and checks if the bit ranges overlap. If they don't overlap, it uses v_or_b32 (OR) instead of v_add_u32 (ADD), valid optimization when the operands truly occupy different bits.
Because fromConstant(-16) returned an empty range, overlaps() always returned false. So the addition (flat_id/8) + (-16) was compiled as (flat_id >> 3) | 0xFFFFFFF0 instead of (flat_id >> 3) - 16.

For flat_id < 128 (i.e., flat_id >> 3 is 0-15), the OR happens to give the same result as subtraction (the lower 4 bits fill in the gap). For flat_id >= 128 (i.e., flat_id >> 3 is 16+), bit 4 is set, and the OR clobbers it, producing -16 instead of 0.

This caused all workgroups with flat_id >= 128 (N-workgroup index >= 32) to compute a wildly wrong M-tile index, leading to out-of-bounds memory accesses and 100% incorrect results.

Changed fromConstant to use the unsigned 32-bit representation via llvm::countr_zero and llvm::countl_zero, so negative constants like -16 = 0xFFFFFFF0 correctly report bits 4-31 as occupied. This forces the Add handler to use v_add_u32 when adding a negative constant to a value whose bits overlap with the constant's set bits.

Locally tested:
| Shape (M, N, K) | Block | Status |
|---|---|---|
| 2048, 57344, 16384 | 224x160x256 | PASSED |
| 4096, 57344, 16384 | 224x160x256 | PASSED |
| 8192, 57344, 16384 | 224x160x256 | PASSED |
| 8192, 57344, 8192 | 224x160x256 | PASSED |
| 16384, 57344, 16384 | 224x160x256 | PASSED |
| 16384, 16384, 16384 | 224x160x256 | PASSED |
| 32768, 57344, 16384 | 224x160x256 | PASSED |
| 32768, 57344, 8192 | 224x160x256 | PASSED |
| 65384, 65384, 65384 | 224x160x256 | FAILED (64-bit overflow in magic constant) |

The `65384x65384x65384` shape fails because the magic-number constant for that shape overflows 32 bits (-425861943773571139 is a 64-bit value), producing an invalid v_mov_b32 operand. 

**TODO: add proper test cases and evaluate (with other MTs) if this is the right way to handle it.**

### DIFF
--- a/waveasm/test/Translate/affine-negative-const-and-normalization.mlir
+++ b/waveasm/test/Translate/affine-negative-const-and-normalization.mlir
@@ -1,0 +1,96 @@
+// RUN: waveasm-translate %s 2>&1 | FileCheck %s
+//
+// Test: correct handling of negative constants in affine expressions and
+// normalization of >32-bit constants produced by MLIR's affine canonicalizer.
+//
+// Bug 1 (negative constants): BitRange::fromConstant(-16) previously returned
+// an empty range because the old while(v > 0) loop never executed for negative
+// values.  This caused the OR-for-add optimisation to fire incorrectly: the
+// compiler emitted v_or_b32 instead of v_add_u32 for (tid floordiv 8) - 16,
+// producing wrong results when tid >= 128 (bit 4 of the quotient overlaps
+// with bit 4 of 0xFFFFFFF0).
+//
+// Bug 2 (large constants): MLIR's affine canonicalizer can compose nested
+// affine.apply ops, multiplying coefficients by a large LCM factor that
+// exceeds 32 bits.  The normalization pass divides all variable coefficients
+// and the divisor by their GCD to bring them back into 32-bit range.
+
+module {
+
+// --- Negative constant: must use v_add_u32, not v_or_b32 ---
+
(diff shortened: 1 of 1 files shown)


willghatch:
I've rebased the PR on main and added a lit test.

willghatch (approved):
I don't have a ton of context for the original problem, but this looks reasonable to me.

PR #1130, opened 2026-03-15
Ixsimpl integration

Hardcode84:
https://github.com/Hardcode84/4/tree/master

### DIFF
--- a/wave_lang/kernel/wave/utils/symbol_utils.py
+++ b/wave_lang/kernel/wave/utils/symbol_utils.py
@@ -10,6 +10,10 @@ import operator as op
 from typing import Callable, Optional
 
 import sympy
+import ixsimpl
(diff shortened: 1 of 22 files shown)


wave_lang/kernel/wave/utils/symbol_utils.py
```diff
@@ -58,6 +63,40 @@
 }
 
 
+####################################################################
+# ixsimpl roundtrip helper.
+####################################################################
+
+# Module-level context reused across calls to benefit from hash-consing.
+_ixs_ctx = ixsimpl.Context()
```
ftynse: Consider thread-safety.

Hardcode84: That's an interesting question, the C context itself is not thread safe, but it would be safe if called it under GIL. Otherwise, we can try to create thread-local contexts.

PR #1134, opened 2026-03-17
fix torch specs in requirements files

willghatch:
The requirements files had "<2.9", but our pyproject.yaml has "<2.10". Also, as an oversight the gen-pytorch-rocm-requirements.py script omitted an upper bound, this adds one at <2.10.


pytorch-rocm-requirements.txt
```diff
@@ -1,2 +1,2 @@
 --index-url https://download.pytorch.org/whl/rocm6.4
```
ftynse: should this point to something newer than rocm6.4 since we are here?

PR #1135, opened 2026-03-17
[water] Add BitcastOp and Scaled Dimensions

### LINKED ISSUE #922: bitcast operations

tgymnich:
Introduces wave.bitcast, a new Water dialect op that reinterprets the raw bits of a register-resident tensor as a different element type. Unlike a plain type cast, bitcast preserves the total bit count by scaling exactly one tensor dimension by the source-to-destination element bitwidth ratio.

New dialect pieces
- wave.bitcast op (WaveOps.td, WaveOps.cpp) with verifier, forward/backward type inference, index-expression propagation, and elements-per-thread propagation.
- Scaled dimensions: a dimension is "scaled" when its symbol maps to a WaveExprListAttr ceiling-division expression in the hyperparameters rather than a plain integer. The verifier enforces at most one such dimension per bitcast and validates the ratio against the hyperparameter values.

Hyperparameter extensions
- WaveExprListAttr — a new attribute wrapping a single-result affine map (ceiling division of a symbol by a constant) together with the referenced symbols. Encodes a derived dimension whose value is ceil(base_symbol / divisor).
- WaveHyperparameterAttr::getSymbolValue extended to resolve WaveExprListAttr entries transitively using an iterative worklist (topological-sort style), so derived symbols whose dependencies are also derived resolve correctly.
- verifyOperationAttribute extended to validate WaveExprListAttr entries: single-result, ceiling division of a symbol by a constant, divisor is an integer, dividend is evenly divisible by divisor, no cycles.

Emitter
- Detects sympy expressions of the form base / n (integer ceiling division) and emits a WaveExprListAttr so the Water type system can track the scaled dimension symbolically.

Lowering
- Pattern to lower wave.bitcast between fully-typed wave tensors to an arith.bitcast on the underlying vector.

fixes #922

### DIFF
--- a/water/lib/Dialect/Wave/IR/WaveDialect.cpp
+++ b/water/lib/Dialect/Wave/IR/WaveDialect.cpp
@@ -5,6 +5,7 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 #include "water/Dialect/Wave/IR/WaveDialect.h"
+#include "mlir/IR/AffineExpr.h"
 #include "mlir/IR/Attributes.h"
 #include "mlir/IR/BuiltinAttributes.h"
 #include "water/Dialect/Wave/IR/WaveAttrs.h"
@@ -21,6 +22,7 @@
 #include "llvm/ADT/StringExtras.h"
 #include "llvm/ADT/StringSet.h"
 #include "llvm/Support/LogicalResult.h"
+#include <cstdint>
 #include <optional>
 
 using namespace mlir;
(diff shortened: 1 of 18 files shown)


water/lib/Dialect/Wave/IR/WaveDialect.cpp
```diff
@@ -391,6 +392,84 @@ wave::WaveDialect::verifyOperationAttribute(Operation *op,
...
+               << " must be a single-result expr_list, but has "
+               << exprList.getMap().getNumResults() << " results";
+      }
+      for (Attribute symAttr : exprList.getSymbols()) {
+        wave::WaveSymbolAttr waveSym =
+            llvm::dyn_cast<wave::WaveSymbolAttr>(symAttr);
+        if (!waveSym) {
+          return op->emitError()
+                 << "hyperparameter " << entry.getName()
+                 << " expr_list contains non-WaveSymbolAttr symbol: "
```
ftynse: Nit: `WaveSymbolAttr` is a C++ class name, not something the user actually sees. Just say "wave symbol" or whatever other tests are using.

water/lib/Dialect/Wave/IR/WaveOps.cpp
```diff
@@ -2452,6 +2452,208 @@ LogicalResult wave::CastOp::verify() {
...
+  if (!dstIRType || !dstIRType.getFullySpecified())
+    return ChangeResult::NoChange;
+
+  // Build the expected result shape: leading dims from source, last dim from
+  // the result IR type (which carries its own symbolic dimension name).
+  SmallVector<wave::WaveSymbolAttr> newShape(srcType.getShape().drop_back(1));
+  newShape.push_back(dstIRType.getShape().back());
+
+  // Validate the last dimension ratio via hyperparameters when available.
+  wave::WaveHyperparameterAttr hyper = wave::getHyperparameters(getOperation());
```
ftynse: Similarly to above, I'm a bit reluctant to let analyses look up hyperparameters every time. InferIndexExprs analysis looks up once and passes around a reference in init functions.

water/lib/Dialect/Wave/IR/WaveOps.cpp
```diff
@@ -2452,6 +2452,208 @@ LogicalResult wave::CastOp::verify() {
...
+
+  if (hyper) {
```
ftynse: IMO this should be shared with verification. That being said, I'd like to keep this as a message from the analysis/inference pass, not just let the verifier let after the fact because passes are expected to produce valid IR.

ftynse (approved):

PR #1154, opened 2026-03-19
add upper bound to gen-pytorch-rocm-requirements.py

willghatch:
Re: https://github.com/iree-org/wave/pull/1150


tgymnich:
sounds good.

PR #1196, opened 2026-03-26
Linearize read indices pipeline

willghatch:
The main goal of this is to be able to have memory addresses for reads in a loop be simplified to `start + IV * stride`, with each of those values either being constant or at least able to be hoisted outside of the loop body.

Adds a pre-codegen pipeline that flattens N-dimensional read addresses into 1-dimensional LINEAR_INDEX accesses:
- flatten_read_indices: rewrites mapped reads to use a single flat offset
- annotate_iv_strides: extracts constant IV strides for loop-carried reads
- Codegen LINEAR_INDEX paths for both vector loads and GatherToLDS
- Removes the old _try_iv_split_offset codegen approach in favor of the new pipeline-based linearization
- Helper functions (mem_simplify, linearize_dims, _infer_floor_to_exact) in mapping_utils for symbolic floor/Mod cancellation
- Adjust bounds for linearized reads

This adds some new lit tests that show that with our mxfp4 shuffle layout we can generate linearized reads with constant stride.

This changes a ton of other lit tests.  Disclaimer:  I was asked to get this PR up ASAP, and this is a ton of churn in the lit tests, and I have not yet validated that they are all correct.

### DIFF
--- a/lit_tests/kernel/wave/mlir_converter.py
+++ b/lit_tests/kernel/wave/mlir_converter.py
@@ -1945,6 +1945,7 @@ def test_mxfp4_scaled_mma_256x256x256():
         K: 16384,
     }
     options = WaveCompileOptions(
+        linearize_reads=False,
         subs=subs,
         location_capture_config=LocationCaptureConfig(level=LocationCaptureLevel.NONE),
         enforce_locations=False,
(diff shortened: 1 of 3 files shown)


wave_lang/kernel/compiler/wave_codegen/read_write.py
```diff
@@ -988,6 +909,261 @@ def _build_mask_with_mapping(
...
+                    input_shape,
+                    emitter,
+                    use_real_bounds_override=True,
+                )
+                lin_src = _cast_buffer_and_encode_stride(
+                    lin_src,
+                    strides_vals,
+                    element_type,
+                    valid_bytes,
+                )
```
Hardcode84: We can it  into lambda to avoid copypaste

willghatch:
@Hardcode84 I've addressed your comments, let me know if you have any other comments.

PR #1202, opened 2026-03-27
[wave2water] roundtrip vector shapes

ftynse:
Support roundtripping vector shapes. This required special handling for
MMA ops where, on the MLIR side, we set vector shapes as a list for LHS,
RHS, Accumulator and Result similarly to index expressions and, on the
FX side, there is a single vector shapes entry that contains dimensions
from all values combined. Special handling is also needed got GetResult
nodes that have explicit vector shapes in addition to the IterateOp they
are getting the result of. Attach these to the IterateOp in the dialect
instead since GetResult are fundamentally absent from the MLIR
counterpart. An alternative solution could be to simply ignore GetResult
during comparison of vector shape entries.


martin-luecke (approved):
LGTM with the recent changes, just a few nits

PR #1206, opened 2026-03-30
[water] provenance-based propagation heuristic

ftynse:
It is unfortunately necessary to introduce this additional complexity to reflect the naive non-dataflow approach in the python counterpart: keep track of the vector shape of the operation from which the lattice originally propagates. This is required by the propagation-stopper heuristic that cannot infer this value from per-value vector shapes that water correctly cleans up to correspond to the shape. The shape is stored as a separate lattice field that is simply carried over based on priority. Since all Mma operations now have different priorities by construction, this never leads to lattice tops outside of synthetic test cases. For Write operations, the other source, the priority is equal but so are vector shapes since those are simply taken from the hardware constraint object.

This is highly dubious due to non-local propagation rules and blind overwriting of vector shapes in the python code, but necessary to match the current behavior. It must be revised as soon as possible to avoid priorities and non-local behavior altogether.


water/lib/Dialect/Wave/IR/WaveOps.cpp
```diff
@@ -1111,50 +1114,60 @@ joinIndexExprsLatticeInPlace(wave::IndexExprsLatticeStorage &lattice,
...
     const wave::IndexExprsLatticeStorage &to, Value toValue) {
-  if (from.isBottom() || from.isTop() || !from.getVectorShape() || to.isTop() ||
+  DictionaryAttr sourceVectorShapes = from.getSourceVectorShape();
+  if (from.isBottom() || from.isTop() || !sourceVectorShapes || to.isTop() ||
       to.isBottom())
     return true;
 
-  if (toValue.getDefiningOp<wave::MmaOp>())
+  if (toValue.getDefiningOp<wave::MmaOp>()) {
+    LLVM_DEBUG(LDBG() << "skipping update to mma");
```
martin-luecke:
```suggestion
    LLVM_DEBUG(LDBG() << "skipping update to mma: " << toValue);
```

PR #1211, opened 2026-03-31
MXFP4 8w Optimizations Dynamic

adedespirlet:
This PR:
- Optimizes MXFP 8w schedule with respect to counters and memory ops.

- Includes a handwritten MLIR snippet that performs swizzle and dword stores to global memory (instead of u shorts). This optmization brings approx 7% improprement. 

- Unrolls the kernel twice to remove forced vmcnt(0) + v_mov copies at the end of the loop
Without unrolling, scale loads and scale consumption overlap within the same iteration, forcing the compiler to load into temporary VGPRs and copy them back to the loop iter_args's registers at the end of the loop (vmcnt(0) + v_mov) . With 2x unrolling, odd/even iterations alternate scale register sets, so loads target already "dead" registers directly. This prevents copies and vmcnt(0) which breaks perf. Scale waits now happen right before the MFMAs, maximizing latency hiding.

- Adds tkw.assumptions constraints to optimize dynamic kernels. 
In this case, the assumptions allow the compiler to omit masking logic when generating dynamic kernels. Specifically, the assumption states that the shape dimension is a perfect multiple of the tile size. With this guarantee, the compiler can safely eliminate bounds checks associated with gather_to_lds operations. This avoids inserting costly masking logic in the dynamic case and improves performance.

TODO: implement a pass that automatically emits the optimized epilogue storing logic.


xintin:
> "TODO: implement a pass that automatically emits the optimized epilogue storing logic."

Nice. Let's add this TODO in the code.

xintin:
As discussed, test on: `shape=(1792, 5376, 4096), block=(256, 192, 256), dynamic=True` to ensure no race condition.

examples/python/7.1_schedule.py
```diff
@@ -208,37 +232,3904 @@ def test_dbuf_8wave_pingpong_mxfp_gemm_Bshuffle_lds(
...
         b_address_space=SHARED_ADDRESS_SPACE,
+        output_dtype=tkl.bf16,
     )
     options.specialize = True
     options.use_buffer_ops = True
     options.minimize_shared_allocs = False
     options.linearize_shared_access = True
-
+    options.wave_runtime = True
+    # options.override_mlir = mlir_256x192
```
ftynse: Is the override currently disabled? Meaning the big string above is basically dead code?

examples/python/7.1_schedule.py
```diff
@@ -208,37 +232,3904 @@ def test_dbuf_8wave_pingpong_mxfp_gemm_Bshuffle_lds(
...
         options.dynamic_symbols = [tkl.sym.M, tkl.sym.N, tkl.sym.K]
         for sym in options.dynamic_symbols:
             del options.subs[sym]
     schedule = get_mxfp4_dbuf_pingpong_schedule_Bshuffled_lds(
-        use_stagger=True, shape=shape
+        use_stagger=True, shape=shape, block=block
     )
+    UNROLL_FACTOR = tkl.sym.UNROLL_FACTOR
+    options.subs[UNROLL_FACTOR] = 2
+    options.postprocess = """
```
ftynse: Is the dance with subs needed or can we just substitute 2 into the string below at the Python level?

tests/kernel/wave_gemm_mxfp_test.py
```diff
@@ -1188,8 +1167,10 @@ def testScaledGemmMXFP48WavePingpongPreshuffleScalesAndB(
     )
     options.specialize = True
     options.use_buffer_ops = True
-    options.minimize_shared_allocs = True
+    options.minimize_shared_allocs = False
```
ftynse: A comment as to why this is turned off is welcome

tests/kernel/wave_gemm_mxfp_test.py
```diff
@@ -1255,6 +1231,17 @@ def testScaledGemmMXFP48WavePingpongPreshuffleScalesAndBLDS(
     schedule = get_mxfp4_dbuf_pingpong_schedule_Bshuffled_lds(
         use_stagger=True, shape=shape
     )
+    UNROLL_FACTOR = tkl.sym.UNROLL_FACTOR
+    options.subs[UNROLL_FACTOR] = 2
```
ftynse: Same as above

ftynse (approved):

examples/python/7.1_schedule.py
```diff
@@ -208,37 +232,3907 @@ def test_dbuf_8wave_pingpong_mxfp_gemm_Bshuffle_lds(
     B data is preshuffled and loaded to LDS (shared memory), not directly to VGPRs.
     A data is read from global memory directly to LDS.
     """
+
```
xintin, line 235: Do we need this big mlir ( ~3600-line MLIR string) to be part of this file? It makes it unreviewable.
Extract or defer the inline MLIR snippets. Either move the ~3700-line MLIR strings to separate .mlir files with a comment documenting their provenance (how did you generate them, or that they were hand-written and verified against which kernel shape), or defer the override_mlir epilogue optimization until the automated pass mentioned in the TODO is implemented. As-is, 87% of the diff (+4240 lines) is opaque MLIR that cannot be maintained.

wave_lang/kernel/wave/schedules/gemm_mxfp4_double_buffer.py
```diff
@@ -811,79 +813,228 @@ def mxfp4_dbuf_schedule():
             loop_bitcast_b, dim=K, num_partitions=2
         )
 
+        # Count only actual global->VGPR Read nodes from the scheduled scale groups.
+        loop_a_scale_reads = tkw.filter_nodes(loop_all_read_a_scale, node_type=tkw.Read)
+        loop_b_scale_reads = tkw.filter_nodes(loop_all_read_b_scale, node_type=tkw.Read)
+        number_outstanding_loads_to_vgpr = len(loop_a_scale_reads) + len(
+            loop_b_scale_reads
+        )
+        safe = 0
```
xintin, line 839: Let's document the block-specific safe padding constants.

wave_lang/kernel/wave/schedules/gemm_mxfp4_double_buffer.py
```diff
@@ -811,79 +813,228 @@ def mxfp4_dbuf_schedule():
...
+                        tkw.SchedulingBarrier([]),
+                        loop_all_read_a_scale,
+                        loop_all_read_b_scale,
+                        loop_shared_load_a_1,
+                        loop_shared_load_b_1,
+                        loop_bitcast_a_1,
+                        loop_bitcast_b_1,
+                        tkw.SchedulingBarrier([]),
+                        tkw.MemoryCounterWait(
+                            load=(number_outstanding_loads_to_vgpr + safe)
```
xintin, line 897: The variable name safe is not clear.
It appears to be an extra counter margin added to MemoryCounterWait in here to prevent (maybe?) hazards on scale VGPRs between adjacent unrolled iterations. This needs a comment explaining what hazard it guards against, how these values were derived (empirically?), and why the (256, 160, 256) branch structurally differs by moving scale reads from Cluster 0 to Cluster 2.

Please add comments wherever needed.

tests/kernel/wave_gemm_mxfp_test.py
```diff
@@ -1086,31 +1086,16 @@ def testScaledGemmMXFP4PreshuffleBWideStores(
 
 
 MACROTILES_PRESHUFFLE_8WAVE_PINGPONG = [
-    (256, 160, 256),
```
xintin, line 1089: are we reducing the coverage from 7 shapes to 4 shapes or is it covered elsewhere?

tests/kernel/wave_gemm_mxfp_test.py
```diff
@@ -1086,31 +1086,16 @@ def testScaledGemmMXFP4PreshuffleBWideStores(
...
-    (64, 192, 256),
-    (64, 128, 256),
+    ((1024, 1920, 8192), (256, 192, 256)),
+    ((1024, 1024, 8192), (256, 256, 256)),
+    ((1024, 1600, 8192), (256, 160, 256)),
+    ((1024, 1024, 8192), (64, 64, 256)),
 ]
 
 
-_DYNAMIC_ALLOWED_PRESHUFFLE_8WAVE_BLOCKS = {
```
xintin, line 1099: question: why are we removing these tests? If it is covered elsewhere, then it is fine.

wave_lang/kernel/wave/templates/tagged_mxfp4_gemm.py
```diff
@@ -180,6 +181,19 @@ def _get_tagged_mxfp4_gemm_preshuffle_scales_impl(
     constraints += [tkw.WaveConstraint(N, BLOCK_N / wave_shape[1])]
     constraints += [tkw.HardwareConstraint(threads_per_wave=64, mma_type=mfma_variant)]
 
+    # Divisibility assumptions for M, N, K (no effect for static shapes).
+    constraints += [tkw.Assumption(Eq(M % 32, 0))]
+    constraints += [tkw.Assumption(Eq(N % 32, 0))]
+    constraints += [tkw.Assumption(Eq(K % 256, 0))]
+
+    # Include assumption that K is divisible by BLOCK_K to allow gather_to_shared ops to omit masking predicates.
+    constraints += [tkw.Assumption(Eq(K % BLOCK_K, 0))]
```
xintin, line 190: Question: why do we need this assumption and `constraints += [tkw.Assumption(Eq(K % 256, 0))]` too, both?

xintin (changes requested):

wave_lang/kernel/wave/schedules/gemm_mxfp4_double_buffer.py
adedespirlet: Added an explanation for the usage of "safe". It wasnt really explained indeed. 
Regarding the structural difference in (256, 160, 256): this is purely a scheduling choice to balance cluster workload more evenly. There is not strict guided rule behind that choice. I was just trying to optimize the workload based on the traces for that specific block tile

wave_lang/kernel/wave/templates/tagged_mxfp4_gemm.py
adedespirlet: if BLOCK_K is different to 256 we still need it. I just kept it such that people can test it with other BLOCK_K values than 256 later on

tests/kernel/wave_gemm_mxfp_test.py
adedespirlet: I’ve streamlined the tests because the previous distinction between dynamic and static block tiles is no longer necessary following optimizations that were implemented, both now run on the same tile sizes.

I remember simplifying the coverage of shapes because the 8wave pingpong schedule didnt support the remaining shapes. While we could add support for them it would require more effort which is not priority anymore

wave_lang/kernel/wave/templates/tagged_mxfp4_gemm.py
xintin: i see. i think we can add a comments like below for each
`# MXFP4 scale-group alignment (independent of tiling)`
and
`# Tiling alignment (allows gather_to_shared to omit masking)
`as you want to put it.

tests/kernel/wave_gemm_mxfp_test.py
xintin: If it was working before this pr, that means although this optimization works, it is not generic until we have the pass in place.
Removing these tests will just make it lost.

What we can do instead:
xfail and document the regression (in the code) with the reason. eg:
`pytest.param((1024, 1024, 8192), (128, 128, 256), marks=pytest.mark.xfail(reason=" ...")),`

xintin (approved):
LGTM! Left two more minor comments.
Once the ci is green, we can merge it.

PR #1212, opened 2026-03-31
Optimize same-denominator fraction addition in gen_sympy_index

adedespirlet:
When the compiler emits IR for sympy expressions containing multiple fraction additions, it defaulted to cross-multiplication even when the denominator was the same across all fractions. This caused denominator explosion. This is now fixed : it identifies when denominators are identical and directly adds the corresponding numerators instead.

### DIFF
--- a/lit_tests/kernel/wave/merge_scale_reads.py
+++ b/lit_tests/kernel/wave/merge_scale_reads.py
@@ -237,3 +237,51 @@ def test_preshuffle_b_opsel():
 
     # No remaining scaled_mfma should have scalar B scale either.
     # CHECK-NOT: amdgpu.scaled_mfma {{.*}} : {{.*}}, vector<{{.*}}xf4E2M1FN>, f8E8M0FNU,
+
+
+@run_test
+def test_opsel_coalescing_asymmetric_2x2():
+    """Verify opsel coalescing for asymmetric block=(256,192,256) with wave_shape=(2,2).
+
+    With (2,2) wave shapes and asymmetric M!=N block dimensions, the opsel
+    pass must handle untraceable yield values where A-scale iter_args don't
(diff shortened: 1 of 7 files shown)


adedespirlet:
> You can still insert sympy logic just before the actual lowering https://github.com/iree-org/wave/blob/main/wave_lang/kernel/compiler/wave_codegen/emitter.py#L959

it makes it more complicated to handle nested stuf but i'll try

wave_lang/kernel/compiler/wave_codegen/emitter.py
```diff
@@ -907,6 +883,36 @@ def _floor(value):
 
         return floordiv_expr(value.numerator, value.denominator)
 
+    def _group_same_denom_fractions(expr):
```
Hardcode84: you can now extract this function and test it independently.

lit_tests/kernel/wave/gen_sympy_index_emission.py
```diff
@@ -0,0 +1,47 @@
+# RUN: python %s | FileCheck %s
```
Hardcode84: lets move this test to our pytest suite

tests/unittests/group_same_denom_fractions_test.py
```diff
@@ -0,0 +1,62 @@
...
+        self.assertEqual(result, expr)
+
+    def test_non_add_passthrough(self):
+        """Non-Add expressions should be returned as-is."""
+        expr = self.x * self.y
+        result = _group_same_denom_fractions(expr)
+        self.assertEqual(result, expr)
+
+
+if __name__ == "__main__":
```
Hardcode84: don't think we need it as we use pytest

Hardcode84 (approved):

PR #1215, opened 2026-03-31
[asm] Cherry pick opsel scaled mfma fix

xintin:
Fix opsel scaled_mfma pass for (2,2) wave shapes and asymmetric blocks

Squashed cherry-pick of 7 opsel/scale coalescing fixes from
`panditsa:4waveasm-256x192x256`:
    1. Handle untraceable yield values in _find_mergeable_groups (aa2a9754)
    2. Add _merge_scale_byte_loads for unrolled sub-word scale loads (418d3433)
    3. Fix _merge_scale_byte_loads grouping for asymmetric blocks (cacee941)
    4. Fix wide load insertion point: yield_op -> byte0_op (ac2cde97)
    5. Fix id() -> hash() in _merge_scale_byte_loads + re-enable v2xi8 (8dcbba38)
    6. Fix id() -> hash() in _coalesce_vector_iter_args (e4d39959)
    7. Handle address-shifted A-scale iter_args (92578f81)
On block=(256,192,256) wave_shape=(2,2): reduces scf.for iter_args from 76 to 73, eliminates all 72 scalar-scale scaled_mfma ops (replaced with vector<4xf8E8M0FNU> + opsel), saves 3 VGPRs.

-------------------------

Additionally, 
1. Eliminated import re; replaced regex string parsing of affine maps with proper AffineMapAttr / AffineAddExpr / AffineConstantExpr API decomposition
2. Extracted _make_extract_slice, module-level helper replaces 4 duplicate ExtractStridedSliceOp construction sites, with a size parameter for the n>1 fallback case
3. id() -> hash(), consistent with the identity fix applied elsewhere in the file
4. Untraceable yield recovery, now tries all group members to find an existing vector<4xi8> source, not just the first arbitrary one
5. Context-aware type caching, _init_types() reinitializes when the MLIR context changes (caught a real cross-test bug)
Added contiguity assumption comment on the wide load creation path

---------------------------
53 tests on main were failing: 

1. testScaledGemmMXFP4PreshuffleMacrotiles (BLOCK_K=128)
Cause: _merge_scale_byte_loads was merging vector<1xi8> LDS loads into vector<4xi8> wide loads even when only 2 scale bytes were contiguous. The extra 2 bytes were garbage from adjacent LDS slots. The opsel pass then treated the full 4-byte vector as valid scale data.
Fix: Removed _merge_scale_byte_loads entirely. The _coalesce_vector_iter_args pre-pass already handles loop-carried scale coalescing, and non-loop-carried loads are already vector<4xi8> from merge_contiguous_reads.

2. testScaledGemmMXFP48WavePingpongPreshuffleScales; crash with "scf.yield" must be a Value (contained a None item)
Cause: After removing _merge_scale_byte_loads, some groups in _coalesce_vector_iter_args couldn't resolve their yield source (stayed None). The None was passed to scf.YieldOp, which crashed.
Fix: Added a filter to drop unresolved groups (yield_source=None) before building the coalesce plan.

### DIFF
--- a/wave_lang/kernel/wave/opsel_scaled_mfma.py
+++ b/wave_lang/kernel/wave/opsel_scaled_mfma.py
@@ -275,11 +275,13 @@ def _find_mergeable_groups(
         eligible.append((i, init_off, init_src, yield_src))
 
     # Phase 2: Group eligible iter_args by their init source vector<4xi8>.
-    # Uses hash() for MLIR Value identity (id() can alias across bindings).
+    # Uses the Operation object directly as dict key (not id() which can
+    # alias across Python bindings, and not bare hash() which has
+    # theoretical collision risk — dict keys use both __hash__ and __eq__).
     by_init_src = defaultdict(list)
     for entry in eligible:
         _, _, init_src, _ = entry
-        by_init_src[hash(init_src.owner)].append(entry)
+        by_init_src[init_src.owner].append(entry)
 
     # Phase 3: Within each source group, greedily form mergeable groups
     # from distinct byte offsets.  A group needs >= 2 offsets to be worth
@@ -305,7 +307,7 @@ def _find_mergeable_groups(
                 init_source = isrc
                 if ysrc is not None:
                     yield_source = ysrc
-                    yield_owners.add(hash(ysrc.owner))
+                    yield_owners.add(ysrc.owner)
                 else:
                     has_untraceable_yield = True
             if has_untraceable_yield:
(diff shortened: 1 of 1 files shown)


wave_lang/kernel/wave/opsel_scaled_mfma.py
```diff
@@ -190,56 +224,98 @@ def _find_mergeable_groups(
...
+            )
         eligible.append((i, init_off, init_src, yield_src))
 
+    # Phase 2: Group eligible iter_args by their init source vector<4xi8>.
+    # Uses hash() for MLIR Value identity (id() can alias across bindings).
     by_init_src = defaultdict(list)
     for entry in eligible:
         _, _, init_src, _ = entry
-        by_init_src[id(init_src.owner)].append(entry)
+        by_init_src[hash(init_src.owner)].append(entry)
```
panditsa: The migration from id() to hash() fixes binding aliasing but introduces a theoretical collision risk. Two distinct Operations could hash to the same value, causing incorrect grouping. Using the Operation object directly as a dict key (which uses both __hash__ and __eq__) would be collision-safe:

```by_init_src[init_src.owner].append(entry)  # uses __eq__ fallback for collisions```

PR #1218, opened 2026-03-31
[asm] Rewrite RegPool with BitVector and add bidirectional VGPR allocation

panditsa:
- **Replace O(n) RegPool with O(1) BitVector and add bidirectional VGPR allocation** — Register operations use `llvm::BitVector` instead of sorted free-list scans. Multi-register VGPR ranges spanning >75% of the program are allocated from the top of expected usage, separating long-lived `buffer_load` prefetch destinations from short-lived `ds_read` values to reduce fragmentation in double-buffered GEMM kernels. Also updates max-sgprs default to 102 for GFX9+.

- **Add tests** — New `linear-scan-bidirectional.mlir` covers long-lived vs short-lived allocation, loop double-buffer pattern, and single-reg/SGPR bypass.

### DIFF
--- a/waveasm/include/waveasm/Transforms/RegAlloc.h
+++ b/waveasm/include/waveasm/Transforms/RegAlloc.h
@@ -168,6 +168,8 @@ public:
     assert(alignment > 0 && "alignment must be positive");
 
     int64_t cap = (ceiling > 0 && ceiling <= maxRegs) ? ceiling : maxRegs;
+    if (size > cap)
+      return -1;
     int64_t highestBase = ((cap - size) / alignment) * alignment;
     for (int64_t candidate = highestBase; candidate >= 0;
          candidate -= alignment) {


waveasm/lib/Transforms/LinearScanRegAlloc.cpp
```diff
@@ -70,31 +70,61 @@ static void insertActiveRange(llvm::SmallVectorImpl<ActiveRange> &active,
...
 
+//===----------------------------------------------------------------------===//
+// Concrete Allocation Strategies
+//===----------------------------------------------------------------------===//
+
+std::optional<int64_t>
+BidirectionalStrategy::allocate(RegPool &pool, const LiveRange &range,
+                                llvm::ArrayRef<LiveRange> allRanges,
+                                int64_t maxPressure) {
+  if (pool.getRegClass() != RegClass::VGPR || range.size <= 1)
```
xintin: The range.size <= 1 guard at the top guarantees range.size >= 2 by the time we reach the ternary. The range.size == 1 branch of allocSingleFromTop is dead code. Should just be pool.allocRangeFromTop(....) directly?

waveasm/include/waveasm/Transforms/RegAlloc.h
```diff
@@ -66,129 +69,196 @@ class PhysicalMapping {
...
+  int64_t allocRangeFromTop(int64_t size, int64_t alignment,
+                            int64_t ceiling = -1) {
+    if (size <= 0)
+      return -1;
+    assert(alignment > 0 && "alignment must be positive");
 
-      // Check if all registers in range are free
+    int64_t cap = (ceiling > 0 && ceiling <= maxRegs) ? ceiling : maxRegs;
+    int64_t highestBase = ((cap - size) / alignment) * alignment;
+    for (int64_t candidate = highestBase; candidate >= 0;
```
harsh-nod: If `size > cap`, this computes `(cap - size) / alignment` where `cap - size` is negative. C++ integer division truncates toward zero, so `highestBase` becomes 0 or negative, and the loop below doesn't iterate — returning -1. Not a memory safety issue since the BitVector is sized at `maxRegs`, and it's unreachable from the current call site (`size` is always a register range ≤ the pool).

Still, a one-line guard makes the intent explicit:

PR #1219, opened 2026-03-31
[asm] Explicit SCC modeling

Hardcode84:
Model the AMDGCN Scalar Condition Code (SCC) flag as a first-class MLIR type in WaveASM instead of treating it as an implicit side effect.

### Problem

Many SALU instructions (bitwise, shifts, adds, compares) clobber the hardware SCC flag, but WaveASM previously had no way to represent this in the IR. This caused:

- **Silent correctness bugs**: CSE could merge two `s_cmp` ops, extending the SCC live range across an SCC-clobbering op (e.g. `s_and_b32`), producing wrong branch conditions.
- **Fragile carry chains**: `s_add_u32` -> `s_addc_u32` carry dependency was only enforced by emission order, not the SSA graph. Any pass reordering ops could break 64-bit adds.

### Solution

**New type and traits:**
- `SCCType` (`!waveasm.scc`) -- 1-bit type representing the hardware SCC flag.
- `SCCDef` / `SCCUse` op traits -- mark ops that write/read SCC.

**New op classes:**
- `SALUUnaryWithSCCOp` / `SALUBinaryWithSCCOp` -- produce explicit `(dst, scc)` results. All SCC-clobbering ops (bitwise, shifts, min/max, bfe, etc.) migrated from `Pure` base classes.
- `SALUBinaryWithCarryInOp` -- `s_addc_u32` takes an explicit `scc_in` operand, creating a true SSA dependency for carry chains.
- `S_CSELECT_B32` -- conditional select reading SCC.
- `S_SCHED_BARRIER` -- scheduling barrier (unrelated but included in the migration).

**New passes:**
- `waveasm-scc-verifier` -- detects SCC-clobbering hazards between producer and consumer. Wired into the pipeline.
- `waveasm-scc-spill-reload` -- inserts `s_cselect_b32` (spill SCC to SGPR) and `s_cmp_ne_u32` (reload) when CSE extends SCC live ranges across clobbers.

**Updated call sites:**
- All SALU op creation sites across TranslateFromMLIR, TranslateFromLLVMDialect, ArithLegalization, ArithHandlers, Peephole, BufferLoadStrengthReduction, RegionBuilder, and AMDGPUHandlers updated for explicit SCC types.
- Assembly emitter skips SCC results/operands during emission.
- ScopedCSE excludes `SCCUse` ops from CSE eligibility.

**Bugfix:** `S_CSELECT_B32` src0 constraint relaxed from `AnySGPR` to `SRegOrImm` to match hardware and allow the spill/reload pass to use immediate operands.

### DIFF
--- a/wave_lang/kernel/wave/compile.py
+++ b/wave_lang/kernel/wave/compile.py
@@ -1365,6 +1365,7 @@ def _generate_asm_code(mb, options):
             else "--waveasm-insert-waitcnt=ticketed-waitcnt=false"
         )
         tail_passes = [
+            "--waveasm-scc-spill-reload",
             "--waveasm-scc-verifier",
             "--waveasm-linear-scan=max-vgprs=256 max-agprs=256",
             waitcnt_flag,
--- a/waveasm/lib/Transforms/SCCVerifier.cpp
+++ b/waveasm/lib/Transforms/SCCVerifier.cpp
@@ -16,8 +16,6 @@
 #include "waveasm/Dialect/WaveASMOps.h"
 #include "waveasm/Transforms/Passes.h"
 
-#include "mlir/IR/Builders.h"
-#include "mlir/IR/BuiltinOps.h"
 #include "mlir/Pass/Pass.h"
 #include "llvm/Support/Debug.h"
 #include "llvm/Support/DebugLog.h"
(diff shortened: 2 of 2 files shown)


harsh-nod (approved):
**Verdict: APPROVE with one mandatory fix.** Confidence: 82/100.

**Summary of findings:**

- **P0 (C1):** The `--waveasm-scc-spill-reload` pass is fully implemented but not wired into `compile.py`. One-line fix required before merge — without it, CSE-merged SCC ops will cause compilation failures at the verifier stage.

- **P1 (C2):** The SCC verifier produces false positives on SGPR-typed conditions (from i64 ordered comparisons). Needs an `isa<SCCType>` type guard — ~5 lines.

- **P2 (C3, C4):** Verifier checks for `S_CSELECT_B32`/`S_ADDC_U32` are incomplete (only detect "no preceding SCC writer" rather than clobbers between definition and use), and the verifier hard-codes 4 consumer ops instead of using the `SCCUse` trait generically.

- **P3 (C5, NC1, NC3):** Cross-block verifier gap, missing `Pure` on carry ops (intentional for safety), and fragile `_ne_` → `_lg_` string replacement.

The core architecture — `SCCType`, `SCCDef`/`SCCUse` traits, explicit `$scc_in` carry chain, CSE exclusion for SCC-reading ops — is sound and resolves real AMDGCN correctness bugs.

wave_lang/kernel/wave/compile.py
```diff
@@ -1351,6 +1351,7 @@ def _generate_asm_code(mb, options):
             else "--waveasm-insert-waitcnt=ticketed-waitcnt=false"
         )
         tail_passes = [
+            "--waveasm-scc-verifier",
```
harsh-nod: The verifier is wired in here but the spill/reload pass (`--waveasm-scc-spill-reload`) isn't. Without it, when CSE merges two SCC-producing ops and an SCC-clobbering instruction sits between the def and the second use, the verifier will catch it but there's no pass to fix it.

Add `"--waveasm-scc-spill-reload",` to the pipeline — it should run after canonicalize/CSE and before the verifier.

harsh-nod (dismissed):
The SCC carry-chain architecture is well-designed — explicit `SCCType`, `SCCDef`/`SCCUse` traits, and CSE exclusion for SCC-reading ops are the right way to handle AMDGCN's implicit SCC semantics.

One thing needs to be fixed before this lands:

### P0 — `--waveasm-scc-spill-reload` not wired into compile.py

The spill/reload pass is fully implemented and tested, but it's not in the pass pipeline. Without it, CSE-merged SCC ops will cause compilation failures at the verifier stage when two SCC consumers are separated by an SCC-clobbering op. One-line fix.

### P1 — SCC verifier false positives on SGPR conditions

The verifier flags SGPR-typed results from i64 ordered comparisons as SCC violations. Needs an `isa<SCCType>` type guard.

### P2 — Verifier consumer list is incomplete

Only 4 ops are checked instead of using the `SCCUse` trait generically. Also, the verifier only detects "no preceding SCC writer" but not clobbers between definition and use.

harsh-nod (approved):
All three items from the previous review are addressed in the latest commit:

1. **P0 fixed** — `--waveasm-scc-spill-reload` is now wired into the pipeline in `compile.py`, running after CSE and before the verifier.

2. **P1 fixed** — The SCC verifier now checks all SCC-typed operands generically instead of hardcoding four specific consumer ops. This eliminates false positives on SGPR-typed conditions from i64 ordered comparisons.

3. **P2 fixed** — Same generic rewrite handles the incomplete consumer list.

The architecture is sound — `SCCType`, `SCCDef`/`SCCUse` traits, explicit `$scc_in` carry chain, and CSE exclusion for SCC-reading ops resolve real AMDGCN correctness bugs. Ready to merge.

harsh-nod (approved):
**Verdict: APPROVE** (confidence: 95/100)

PR #1219 introduces first-class SCC (Scalar Condition Code) modeling for AMDGCN in WaveASM. After thorough analysis of the diff and rebuttals, the PR is in excellent shape:

- **No blocking issues found.** The original high-severity findings (P0/P1/P2) from earlier reviews were all factually incorrect — the passes are wired in, the verifier is generic with proper type guards, and clobber detection works correctly.
- **Two minor P3 findings** remain: (1) the verifier skips cross-block SCC checks, which is mitigated by the spill/reload pass's conservative cross-block handling, and (2) a fragile `_ne_` → `_lg_` string replacement that works but could use a `StringSwitch` for maintainability.
- **Key correctness wins:** explicit carry chain SSA dependencies prevent silent reordering bugs, SCCUse ops are correctly excluded from CSE, and the spill/reload + verifier pipeline ordering is sound.

PR #1220, opened 2026-03-31
[asm] Cherry pick: Generalize WAR hazard detection and improve liveness sort order + tests and fixes

xintin:
- Cherry-picks [125cdc2](https://github.com/iree-org/wave/commit/125cdc213593411768ea494088e2f12eb81849c7) ("Generalize WAR hazard detection and improve liveness sort order") which extends WAR hazard detection beyond buffer_load to cover any iter_arg/block_arg overlap, and improves live range sort order to reduce fragmentation during allocation.
- Adds 10 lit tests across two files covering the new behavior:
> - scalar-war-hazard.mlir — 6 register-allocation behavior tests (SGPR WAR, VGPR WAR, same-point no-hazard, mixed MFMA+scalar, multiple post-def uses, negative case)
> - war-hazard-emit.mlir — 4 emit-assembly tests that verify back-edge copy insertion and produce different output before vs after the fix
- Applies fixes on top of the cherry-pick:
> - Removes dead #include "WaveASMInterfaces.h"
> - Restores unconditional untie for single-element buffer loads (_ubyte/_sbyte/_ushort/_sshort) whose transitive uses through bitcast/extract are not captured by usePoints
> - Replaces dead if (!defOp) return false with assert(defOp) since a non-BlockArgument Value always has a defining op
> - Adds tie-break rationale to the sort comparator comment

### DIFF
--- a/waveasm/lib/Transforms/Liveness.cpp
+++ b/waveasm/lib/Transforms/Liveness.cpp
@@ -5,7 +5,6 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 #include "waveasm/Transforms/Liveness.h"
-#include "waveasm/Dialect/WaveASMInterfaces.h"
 #include "waveasm/Dialect/WaveASMOps.h"
 #include "waveasm/Dialect/WaveASMTypes.h"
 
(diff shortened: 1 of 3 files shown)


waveasm/lib/Transforms/Liveness.cpp
```diff
@@ -31,44 +31,50 @@ static bool isSwapPatternIterArg(Value iterArg, Block &bodyBlock, unsigned i) {
...
+  // at a point where block_arg still has subsequent uses (strictly after,
+  // since same-point means read-before-write), tying them creates a WAR
+  // hazard.
   auto iterDefIt = info.defPoints.find(iterArg);
   auto baUseIt = info.usePoints.find(blockArg);
   if (iterDefIt != info.defPoints.end() && baUseIt != info.usePoints.end()) {
-    int64_t loadDef = iterDefIt->second;
+    int64_t iterDef = iterDefIt->second;
     for (int64_t usePoint : baUseIt->second) {
-      if (usePoint >= loadDef)
```
xintin: under investigation: if this change is safe for the existing buffer_load multi-register overlap check!

PR #1225, opened 2026-04-01
[asm] Enhance peephole pass with compile-time constant folding for VALU add/mul

panditsa:
- **Add compile-time constant folding for `V_ADD_U32` and `V_MUL_LO_U32`** — When both operands are compile-time constants (including through `V_MOV_B32` indirections via `getConstantValue`), fold the operation into a `v_mov_b32` of the computed result with 32-bit wrapping. Multiplication uses `uint64_t` arithmetic to avoid signed overflow UB on large operands.

- **Upgrade identity/zero elimination to see through `V_MOV_B32`** — `AddZeroPattern`, `MulOnePattern`, and `MulZeroPattern` now use `getConstantValue` instead of requiring a direct `ConstantOp`, so patterns like `add(v_mov_b32(0), x)` and `mul(v_mov_b32(1), x)` are correctly simplified.

- **7 new lit tests** covering constant folding (add, mul), 32-bit overflow wrapping, large-operand truncation, and MOV-indirection for zero/one elimination.

### DIFF
--- a/waveasm/lib/Transforms/Peephole.cpp
+++ b/waveasm/lib/Transforms/Peephole.cpp
@@ -513,7 +513,9 @@ struct AddZeroPattern : public OpRewritePattern<V_ADD_U32> {
 
     // Fold: both operands constant -> compute result at compile time.
     if (val0 && val1) {
-      int64_t result = (*val0 + *val1) & 0xFFFFFFFF;
+      int64_t result = static_cast<int64_t>(
+          (static_cast<uint64_t>(*val0) + static_cast<uint64_t>(*val1)) &
+          0xFFFFFFFF);
       auto loc = addOp.getLoc();
       auto immType = rewriter.getType<ImmType>(result);
       auto constOp = ConstantOp::create(rewriter, loc, immType, result);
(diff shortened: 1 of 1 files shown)


harsh-nod (dismissed):
**Verdict: APPROVE with minor requested changes (confidence 82/100)**

I traced the code across `Peephole.cpp`, `Utils.h`, and `WaveASMTypes.h` to verify each finding. Four items, ordered by priority:

1. **P1 — Missing VGPR type guard in identity elimination** (`Peephole.cpp` lines 527–533, 553–558). `AddZeroPattern` and `MulOnePattern` can replace a VGPR result with an SGPR/immediate operand, producing malformed IR. The `BFEPackIdentityPattern` in the same file already has an `isVGPRType` guard for exactly this reason. This is pre-existing but the PR widens the trigger surface via `getConstantValue`. Fix: ~2 lines per pattern.

2. **P2 — Add fold signed arithmetic inconsistency** (line 516). The add path uses raw `int64_t` addition while the mul path correctly uses `uint64_t` intermediates. Provably safe for 32-bit operands, but the asymmetry is a code quality issue. One-liner fix for consistency.

3. **P2 — Pattern priority shadowing** (lines 1079–1101). `MulPow2ToShiftPattern` fires before the constant fold in `MulZeroPattern`, so `mul(4, 3)` produces a shift instead of `v_mov_b32(12)`. Correct but suboptimal.

4. **P3 — Test gaps**. Missing reversed-operand identity test, SGPR-input test, and negative constant test.

PR #1228, opened 2026-04-02
[wave2water] Add MXFP4 support to fx_emitter and fix roundtrip comparison

martin-luecke:
Teach the MLIR-to-FX importer to handle ops used by the MXFP4 GEMM kernel and fix the roundtrip comparison for MMA index expressions.

FX emitter:
- Add `_handle_mma_like_op` handling both `wave.mma` and `wave.scaled_mma`
- Add `_handle_bitcast_op` for `wave.bitcast`.
- Add derived dimension support: hyperparameter parsing separates base integers from WaveExprListAttr expressions (e.g. K32 = ceiling(K/32)), with `_resolve_dim` mapping MLIR dimension names back to symbolic expressions.
- Thread `parse_ctx` through all `_convert_supported_attrs` call sites.

Attribute converter:
- Replace `convert_index_mapping_array_to_sympy` (which dispatched on `op.name` strings) with `convert_index_mapping_dict_to_sympy` (single mapping) and `convert_mma_index_to_sympy` (MMA/ScaledMMA reconstruction with derived_dims support).
- Assert lhs_scale/rhs_scale index consistency for ScaledMMA.

Roundtrip comparison:
- Compare MMA/ScaledMMA indices per-operand (`lhs_index`, `rhs_index`, etc.)  rather than on the combined dict. The combined dict carries Piecewise expressions over operand-selector symbols that are dead for certain dtype configurations and lost during MLIR serialization.
- Add WaveConstraint comparator and thread `subs` into `assert_traces_equivalent`

### DIFF
--- a/wave_lang/kernel/wave/mlir_converter/attr_type_converter.py
+++ b/wave_lang/kernel/wave/mlir_converter/attr_type_converter.py
@@ -295,7 +295,7 @@ def _convert_index_mapping_attr_to_sympy(
     return IndexSequence(start, step, stride)
 
 
-def convert_index_mapping_dict_to_sympy(
+def convert_symbol_mapping_attr_to_sympy(
     mapping_attr: wave.WaveSymbolMappingAttr,
     derived_dims: dict[str, IndexExpr] | None = None,
 ) -> dict[IndexSymbol, IndexSequence]:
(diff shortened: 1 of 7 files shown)


wave_lang/kernel/wave/mlir_converter/attr_type_converter.py
```diff
@@ -376,11 +406,22 @@ def convert_index_mapping_array_to_sympy(
...
+    if has_scale_operands:
+        scale_shared = set(lhs_scale_index.keys()) & set(rhs_scale_index.keys())
+        for sym in scale_shared:
+            assert lhs_scale_index[sym] == rhs_scale_index[sym], (
+                f"lhs_scale and rhs_scale disagree on {sym}: "
+                f"{lhs_scale_index[sym]} vs {rhs_scale_index[sym]}"
+            )
+        scale_only = set(lhs_scale_index.keys()) - set(combined.keys())
+        for sym in scale_only:
+            combined[sym] = lhs_scale_index[sym]
```
ftynse: IIUC, this will put the scale-only dimensions at the end of the dictionary. Let's double-check it is the right location for them and have a comment explaining why, given that the ordering is load-bearing (it isn't for MMAs AFAIK, but may be for reads/writes when it gets propagated to those).

ftynse (approved):
LGTM % an ordering issue with index attrs.

PR #1230, opened 2026-04-02
[water] improve error message for source vector shape conflict

ftynse:
This is a separate field of the index expression lattice so it is helpful to highlight which field is causing conflicts.

### DIFF
--- a/water/lib/Dialect/Wave/IR/WaveInterfaces.cpp
+++ b/water/lib/Dialect/Wave/IR/WaveInterfaces.cpp
@@ -1609,7 +1609,8 @@ wave::IndexExprsLatticeStorage::getJoinedSourceVectorShape(
   int32_t rhsPri = rhs.getSourceVectorShapePriority();
 
   if (!lhsSVS && !rhsSVS)
-    return std::make_pair(DictionaryAttr(), wave::IndexExprsLatticeStorage::kLowestPriority);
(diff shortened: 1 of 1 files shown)


tgymnich (approved):

water/lib/Dialect/Wave/IR/WaveInterfaces.cpp
```diff
@@ -1600,6 +1600,29 @@ static DictionaryAttr getJoinedVectorShape(DictionaryAttr lhs,
...
+FailureOr<std::pair<DictionaryAttr, int32_t>>
+wave::IndexExprsLatticeStorage::getJoinedSourceVectorShape(
+    const IndexExprsLatticeStorage &lhs, const IndexExprsLatticeStorage &rhs) {
+  DictionaryAttr lhsSVS = lhs.getSourceVectorShape();
+  DictionaryAttr rhsSVS = rhs.getSourceVectorShape();
+  int32_t lhsPri = lhs.getSourceVectorShapePriority();
+  int32_t rhsPri = rhs.getSourceVectorShapePriority();
+
+  if (!lhsSVS && !rhsSVS)
+    return std::make_pair(DictionaryAttr(), int32_t(0));
```
tgymnich:
```suggestion
    return std::make_pair(DictionaryAttr(),  wave::IndexExprsLatticeStorage::kLowestPriority;);
```

PR #1236, opened 2026-04-03
[asm] Cherry-pick: Add VGPR compaction pass to eliminate register fragmentation

xintin:
Post-allocation pass that reassigns physical VGPRs using shortest-first greedy packing to eliminate fragmentation from interleaved `buffer_load` (long-lived) and `ds_read` (short-lived) destinations. Wired as `--waveasm-vgpr-compaction` in the pass pipeline after linear scan allocation.

## Current State:
The linear scan register allocator assigns VGPRs in instruction order (lowest-available-first). When 4-wide `buffer_load_dwordx4` destinations (long-lived, data in flight from global memory) and 1-wide `ds_read`/address computation results (short-lived, consumed within a few instructions) are interleaved in a scheduled GEMM loop body, the allocator assigns them adjacent register numbers. When the short-lived values expire, their registers become gaps between still-live 4-wide ranges. These alignment gaps inflate `.vgpr_count` (= highest register number + 1) beyond the true simultaneous pressure.
For example, on `block=(256,192,256)` MXFP4 preshuffle-B GEMM:
- **Without compaction:** `.vgpr_count`: 268
- **With compaction:** `.vgpr_count`: 263 (saved 5 VGPRs)
Every VGPR saved brings high-pressure kernels closer to the 256-VGPR hardware limit on gfx950.

Results with other block sizes:
| Kernel | Without compaction | With compaction | Saved |
|---|---|---|---|
| (256, 192, 256) | 268 | 263 | 5 |
| (128, 256, 256) | 252 | 251 | 1 |
| (128, 128, 256) | 170 | 169 | 1 |
| (64, 64, 256) | 107 | 105 | 2 |

## Why a post-pass (not in-allocator)?
We evaluated four in-allocator alternatives:
| Approach | Result | Why it failed |
|---|---|---|
| Length-aware heuristic (long-lived -> high regs) | 512 VGPRs | Inflates max register number |
| Stage 2 mapping compaction (post-allocate, pre-rewrite) | 271 VGPRs | Misses extract/insert/pack-derived registers |
| Alignment-aware partitioning (reserve low regs for 4-wide) | 309 VGPRs | Pushes 1-wide values to high registers |
| Partitioning + post-pass | 265 VGPRs | Still worse than post-pass alone |
The post-pass succeeds because it operates on the fully materialized physical IR, after all extract/insert/pack fixups, loop extensions, and `_iterArgPhysRegs` attribute creation. No single-pass in-allocator heuristic can replicate this global view. The allocator's lowest-first policy is already locally optimal for minimizing `.vgpr_count`; the post-pass closes the remaining alignment gaps.

### DIFF
--- a/waveasm/lib/Transforms/LinearScanPass.cpp
+++ b/waveasm/lib/Transforms/LinearScanPass.cpp
@@ -74,6 +74,139 @@ static Type makePhysicalType(MLIRContext *ctx, Type virtualType,
   return virtualType;
 }
 
+//===----------------------------------------------------------------------===//
+// Rematerialization
+//===----------------------------------------------------------------------===//
+
+/// Return true when \p op sits inside a LoopOp body at any nesting depth.
+static bool isInsideLoopBody(Operation *op) {
+  for (Operation *p = op->getParentOp(); p; p = p->getParentOp()) {
+    if (isa<LoopOp>(p))
+      return true;
+  }
+  return false;
+}
+
+/// Check if an operation can be cheaply rematerialized (cloned near each
+/// use to shorten its live range).
+///
+/// Accepted patterns:
+///   - V_MOV_B32 with immediate operands  (accumulator zero-init)
+///   - V_MOV_B32 with SGPR source         (scalar-to-vector address copy)
+///   - S_MOV_B32 with immediate operands  (scalar constant materialisation)
+static bool isRematerializableOp(Operation *op) {
+  if (op->getNumResults() != 1)
+    return false;
+
+  if (isa<V_MOV_B32>(op)) {
+    for (Value operand : op->getOperands()) {
+      auto *defOp = operand.getDefiningOp();
+      if (!defOp)
+        return false;
+      if (isa<ConstantOp>(defOp))
+        continue;
+      // SGPR source: the scalar value dominates the original def, so it
+      // also dominates every use site where we might place a clone.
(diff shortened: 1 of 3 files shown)


harsh-nod (changes requested):
**Verdict: REQUEST CHANGES** (confidence: 92/100)

The VGPR compaction pass is well-engineered overall — the shortest-first greedy algorithm is correct, sub-element mapping is clean, and real-kernel results validate the approach (5 VGPRs saved on 256x192x256 GEMM).

**One blocking issue:** `applyRemapping` double-remaps `LoopOp`/`IfOp` results. Walk 1 (generic `Operation*`) remaps all results including LoopOp/IfOp, then Walks 2/3 remap them *again* using the already-mutated type. When the `oldToNew` map contains chains (e.g., `{4→0, 0→4}`), this produces incorrect register assignments. I verified this by tracing through the `remap` lambda — it reads the *current* type, not the original, making it non-idempotent. The reference `LinearScanPass.cpp` avoids this entirely by using a `Value→physReg` mapping instead.

**Prioritized fixes:**
1. **P0** — Remove result-remapping from Walks 2/3 (block arg remapping stays)
2. **P1** — Add a LoopOp lit test that forces a chained remap
3. **P2** — Remove dead `rangeBaseToIdx`, deduplicate `kScratchVGPR`
4. **P3** — Add IfOp test, document alignment cap

PR #1242, opened 2026-04-03
[asm] Enhance buffer load strength reduction with local CSE, voffset dedup, soffset folding, and SGPR address chain support

xintin:
Add localCSERange for cloned address chain deduplication. Add voffset decomposition into (VGPR base, SGPR addend, instOffset constant) for cross-load deduplication. Add peepholeSoffsetFold to fold constant soffsets into instOffset. Support SGPR address chains (S_MUL_I32, S_ADD_U32, S_MOV_B32) in address analysis and static stride computation.

Chery-picked from [panditsa/wave@fb37585.](https://github.com/iree-org/wave/commit/fb375859323e4793527c46530e08749bc037bdcc)

---

Extends the buffer load strength reduction pass with four capabilities that eliminate redundant VALU address computation and reduce VGPR pressure in GEMM loop bodies:

- Local CSE for cloned address chains: After precomputing voffsets before the loop, independently cloned address chains for different buffer_load candidates share the same leaf operands but produce duplicate SSA ops. localCSERange merges structurally identical ops with commutative operand canonicalization, collapsing N identical chains into one.
- Voffset deduplication via SGPR addend extraction: Decomposes precomputed voffsets into (shared VGPR base, per-load SGPR addend, compile-time constant). Loads that differ only by an SGPR offset (e.g., N-tile index) and/or a constant (e.g., K-half byte offset) share one VGPR voffset. The SGPR part folds into per-load s_add_u32(group_soffset, sgpr_addend), and the constant folds into the hardware instOffset field (12-bit unsigned, max 4095).
- Peephole soffset fold: Post-transformation cleanup that detects buffer_load ops with `voffset = V_ADD_U32(vgpr, sgpr)` and `soffset = 0`, and moves the SGPR part into the hardware soffset field. This saves one VALU per load by using the free scalar offset path in the memory controller.
- SGPR address chain support: Extends isAddressVALU and computeStaticStride to handle scalar ops (S_MUL_I32, S_ADD_U32, S_SUB_U32, S_MOV_B32) that appear in buffer_load address chains when the IV is an SGPR (which it always is in WaveASM loops). This enables stride computation through mixed VALU/SALU address expressions like v_add_u32(tid, s_mul_i32(iv, element_bytes)).

**Why it's needed**
In MXFP4 preshuffle-B GEMM kernels on MI350X (gfx950), the loop body contains 8-12 buffer loads for A-data, B-data, and scale reads. Without this pass, each load recomputes its voffset from scratch every iteration, typically 3-5 VALU ops per load (add IV, shift for byte offset, add constant offset). That's 30-50 VALU ops per iteration on the critical path, competing with MFMA for execution resources.

**After this pass:**
All IV-dependent address computation is replaced by one s_add_u32 per SRD group per iteration (typically 2-3 groups)
B-data loads that differ only by N-tile SGPR offset share one VGPR voffset instead of each keeping their own
Constant byte offsets (K-half splits) use the hardware instOffset field at zero cost.

**Additional fixes applied**
- V_LSHL_OR_B32 delta computation now bails when the OR operand is IV-dependent, preventing incorrect stride calculation when bitwise OR and addition semantics diverge
- Added compile-time soffset overflow guard: extracts loop bound from s_cmp_lt_u32 and rejects candidates where abs(stride) * max_iterations > INT32_MAX
- decomposeVoffset now iteratively peels SGPR addends via a while loop, handling multi-level V_ADD_U32(V_ADD_U32(base, sgpr1), sgpr2) chains
- Documented localCSERange O(N*M) complexity and upgrade path 
- Updated all test targets from gfx942 to gfx950
- Added negative stride lit test (downward-counting IV with step=-1)

### DIFF
--- a/wave_lang/kernel/wave/compile.py
+++ b/wave_lang/kernel/wave/compile.py
@@ -1368,6 +1368,7 @@ def _generate_asm_code(mb, options):
             "--waveasm-scc-spill-reload",
             "--waveasm-scc-verifier",
             "--waveasm-linear-scan=max-vgprs=256 max-agprs=256",
+            "--waveasm-vgpr-compaction",
             waitcnt_flag,
             f"--waveasm-hazard-mitigation=target={options.target}",
             "--emit-assembly",
--- a/waveasm/include/waveasm/Transforms/AssemblyEmitter.h
+++ b/waveasm/include/waveasm/Transforms/AssemblyEmitter.h
@@ -122,6 +122,9 @@ public:
   int64_t getPeakAGPRs() const { return peakAGPRs; }
 
 private:
+  /// Scratch VGPR for loading non-inline literals (v15).
+  /// VGPRCompaction has a matching constant; keep them in sync.
+  static constexpr int64_t kScratchVGPR = 15;
   /// Resolve an SSA Value to its physical register string
   std::string resolveValue(mlir::Value value);
   std::string resolveScalarValue(mlir::Value value);
@@ -208,13 +211,6 @@ private:
 
   /// Counter for generating unique loop labels in assembly
   int loopLabelCounter = 0;
-
-  /// Scratch VGPR for loading non-inline literals.
-  /// We use a lower VGPR (v15) to avoid excessive VGPR allocation.
-  /// The Python backend uses v_mov_b32 to materialize large constants
-  /// during IR generation, which is better.
-  /// TODO: Handle this properly during MLIR translation instead.
-  static constexpr int64_t kScratchVGPR = 15;
 
   /// Cached value currently held in kScratchVGPR, used to avoid redundant
   /// v_mov_b32 materializations of the same literal.
--- a/waveasm/include/waveasm/Transforms/Passes.td
+++ b/waveasm/include/waveasm/Transforms/Passes.td
@@ -277,6 +277,26 @@ def WAVEASMSCCVerifier : Pass<"waveasm-scc-verifier"> {
 }
 
 //===----------------------------------------------------------------------===//
+// VGPR Compaction Pass
+//===----------------------------------------------------------------------===//
+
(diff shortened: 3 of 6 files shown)


harsh-nod (changes requested):
The buffer load strength reduction enhancements are well-motivated — eliminating 30-50 VALU ops per loop iteration via local CSE, voffset dedup, and soffset folding is a big win for GEMM throughput. The SGPR address chain support and `V_LSHL_OR_B32` bailout fix are also correct.

However, there's one blocking correctness issue:

### P0: `sgprAddends` computed but never consumed

During voffset decomposition, the SGPR addend is correctly extracted and stored in `sgprAddends[i]`. But when the patching code rewrites buffer_load soffsets, it only consumes `instOffsetDeltas` — the `sgprAddends` vector is never read. For loads with SGPR-dependent offsets (the primary B-data use case in GEMM), the SGPR addend is silently dropped from the effective address.

The lit test `@sgpr_addend_extraction` encodes the current (buggy) behavior in its CHECK lines, so it passes, but the address semantics are wrong: `EA = base + vgpr_base + inst_offset` instead of the correct `EA = base + vgpr_base + sgpr_addend + inst_offset`.

**Fix:** In the soffset patching loop, fold `sgprAddends[i]` into `soffset` via `s_add_u32(group_soffset, sgpr_addend)`, similar to how `instOffsetDeltas` is folded into `instOffset`.

### P1: Signed overflow in `computeStaticStride`

`int64_t diff = strideDelta * elementBytes` can overflow for large strides. Use unsigned intermediates and check before casting.

### Positive notes

- The local CSE with commutative canonicalization is clean and well-bounded (O(N*M) per range)
- The `V_LSHL_OR_B32` bailout when OR operand is IV-dependent is a correct and important fix
- The soffset overflow guard using loop bound extraction is a good safety measure
- The peephole soffset fold is EA-preserving under modular 32-bit arithmetic

PR #1266, opened 2026-04-06
Add wave-dev-setup.sh for worktree-aware build and environment setup

willghatch:
This is based on a build script that I've been using to streamline working with worktrees. IE I wanted a single command to set up a worktree (main worktree OR extra worktrees), for myself and for agents to easily get going. The main way to use it is to run `./build_tools/wave-dev-setup.sh --worktree-initialize`, which builds water and waveasm (and llvm if necessary), creates a virtualenv, and does an editable install of wave in that virtualenv. It uses `uv virtualenv` and `uv pip install`, since uv installs using hard links to a cache, so a big pip package like pytorch only needs to be downloaded once.

It is opinionated about setup, but has some overrides via environment variables, eg. for the path to the LLVM to use. It could probably have more setup options, eg. right now it is hard coded to assume a ROCM install (not a cpu install), using the requirement detection script to get the right pytorch. And it always builds waveasm and water.
But I think these are all the right defaults.

### DIFF
--- a/build_tools/wave-dev-setup.sh
+++ b/build_tools/wave-dev-setup.sh
@@ -131,7 +131,7 @@ MLIR_DIR="$LLVM_BUILD/lib/cmake/mlir"
 LLD_DIR="$LLVM_BUILD/lib/cmake/lld"
 
 git_clone_depth=2
-VERBOSE=0
+VERBOSE="${VERBOSE:-0}"
 
 # ---------------------------------------------------------------------------
 # Usage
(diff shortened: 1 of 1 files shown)


build_tools/wave-dev-setup.sh
```diff
@@ -0,0 +1,569 @@
...
+# ASSUMPTIONS
+# ============================================================================
+#
+# This script assumes the following default directory layout, where WAVE_REPO
+# is the **main** Wave checkout (not a worktree):
+#
+#   WAVE_REPO/                        The main Wave git repository
+#   WAVE_REPO/../llvm-wave/           LLVM base directory (WAVE_LLVM_DIR):
+#     llvm-project/                     LLVM source (cloned from git)
+#     build/                            LLVM CMake build directory
```
ftynse: Nit: LLVM tends to build in `llvm-project/build` rather than in an adjacent directory.

build_tools/wave-dev-setup.sh
```diff
@@ -0,0 +1,569 @@
...
+# means worktrees can live anywhere and still find shared dependencies.
+#
+# The script auto-detects whether it is running inside a git worktree and
+# resolves the main repo root accordingly.
+#
+# The script uses a virtualenv at `$WAVE_DIR/.venv`.  Each worktree will have a
+# different virtualenv.  The script uses `uv virtualenv` to reduce install
+# footprint -- `uv pip install` uses hard links to `$UV_CACHE_DIR`, so a big
+# download/install like pytorch is not duplicated, and worktree installs are
+# fast.
```
ftynse: This sounds like a good idea in general! Can we update the main README, in a separate PR, to suggest using `uv` systematically?

PR #1276, opened 2026-04-07
suggest uv instead of normal pip in docs

willghatch:
As mentioned in PR comments from https://github.com/iree-org/wave/pull/1266 using `uv` is nice for development generally, so let's just standardize on it in the setup instructions.


panditsa:
Needs a wider agreement, but I like this change!

ftynse (approved):

tgymnich (approved):

PR #1282, opened 2026-04-08
Add wide_stores support for MXFP4 (4wave) GEMM (llvm backend)

xintin:
## Wide epilogue stores for bf16 MXFP4 GEMM via `v_permlane16_swap`

In MXFP4 GEMM kernels with bf16 output, the MMA accumulator layout (`F32_16x16x128_F8F6F4`) distributes results across threads in a non-contiguous pattern each thread holds 4 values spanning 4 rows but only 1 column. Without optimization, the epilogue emits 192 scalar `buffer_store_short` (2 bytes each), underutilizing memory bandwidth.

This PR adds a `wide_stores` that swaps MFMA operands (B as LHS, A as RHS) and uses `v_permlane16_swap_b32` to coalesce epilogue stores into `buffer_store_dwordx4` (16 bytes each), reducing store count from 192 to 48.

### How it works
**Swapped MFMA operands (`wide_stores=True`):** The kernel computes C^T = B × A^T (and stores C) instead of C = A × B^T, making the accumulator's 4-contiguous values align with the output memory's stride-1 dimension. The source/target write syntax maps `Register[N, M]` to `Memory[M, N]` so the output is C [M, N] directly.

**`wide_store_coalescing` pass:** An always-on graph pass that identifies eligible bf16 global Write nodes using the source/target dimension remapping pattern (indicating swapped operands). Only tags writes that satisfy all conditions: global address space, bf16 dtype, and source/target syntax. Safe to run unconditionally, non-`wide_stores` kernels are unaffected.

**`_write_permlane_pack_to_global` codegen:** At code generation time, tagged writes emit `v_permlane16_swap_b32` to exchange each thread's 4 bf16 values (packed as 2 i32 dwords) with a partner lane 16 positions apart. Both lane halves write identical 8-bf16 vectors to the same address (benign duplicate store), avoiding divergent control flow. The buffer descriptor's `valid_bytes` handles out-of-bounds suppression for dynamic shapes.

### Assembly comparison (shape 1024×3072×8192, block 256×192×256)

| Instruction | Baseline | Wide Stores | Change |
|---|---|---|---|
| `buffer_store_short` | 192 | 0 | eliminated |
| `buffer_store_dwordx4` | 0 | 48 | 16B/store |
| `v_permlane16_swap` | 0 | 96 | lane exchange |
| `v_cvt_pk_bf16_f32` | 192 | 96 | 2x reduction |
| Total asm lines | 12,115 | 9,322 | 23% fewer |

LLVM now emits `v_cvt_pk_bf16_f32` natively (bf16 softening has been fixed upstream), so the paired f32->bf16 conversion is a single instruction.

Compute-bound shapes (large K) show 6-9% gains; small-K shapes show 13-27% gains

| Shape (M, N, K) | Baseline TFLOPS | Wide Stores TFLOPS | Speedup |
|---|---|---|---|
| (315904, 384, 1792) | 1169.51 | 1557.86 | 1.332x |
| (20480, 1152, 28928) | 2714.36 | 3015.93 | 1.111x |
| (5888, 1920, 3328) | 1582.35 | 2104.81 | 1.330x |
| (46336, 2688, 8448) | 2495.16 | 2940.50 | 1.179x |
| (512, 3072, 428288) | 590.71 | 648.15 | 1.097x |
| (256, 4224, 102656) | 409.79 | 458.95 | 1.120x |
| (512, 4224, 55808) | 794.47 | 893.25 | 1.124x |
| (4864, 4608, 2560) | 1435.43 | 2123.30 | 1.479x |
| (4864, 4608, 5888) | 2083.50 | 2669.39 | 1.281x |
| (4608, 4992, 7424) | 2227.52 | 2813.12 | 1.263x |
| (2048, 5376, 63232) | 3112.73 | 3405.23 | 1.094x |
| (2048, 5760, 6400) | 2230.97 | 2690.92 | 1.206x |
| (3584, 6912, 548608) | 3272.81 | 3531.95 | 1.079x |
| (4864, 7680, 7936) | 2406.89 | 3036.57 | 1.262x |
| (13056, 10368, 15360) | 2936.09 | 3352.67 | 1.142x |
| (14848, 12672, 10496) | 2781.61 | 3254.58 | 1.170x |
| (1792, 19968, 60928) | 3115.72 | 3392.02 | 1.089x |

### DIFF
--- a/examples/python/7.1_schedule.py
+++ b/examples/python/7.1_schedule.py
@@ -32,6 +32,7 @@ from wave_lang.kernel.wave.schedules import (
 from wave_lang.kernel.wave.templates import (
     get_tagged_mxfp4_gemm,
     get_tagged_mxfp4_gemm_preshuffle_b,
+    get_tagged_mxfp4_gemm_preshuffle_b_wide_store,
     get_tagged_mxfp4_gemm_preshuffle_scales,
     get_tagged_mxfp4_gemm_preshuffle_scales_and_B,
 )
@@ -432,18 +433,16 @@ def test_dbuf_4wave_mxfp_dynamic_preshuffle_b_gemm_wide_stores(
 ):
     """Preshuffle-B MXFP4 GEMM with dynamic M, N, K and wide epilogue stores.
 
-    Uses wide_stores=True to swap MFMA operands (B as LHS, A as RHS),
+    Uses the wide_store variant to swap MFMA operands (B as LHS, A as RHS),
     aligning the accumulator's contiguous values with the output's stride-1
     dimension. The coalesce_wide_stores pass emits v_permlane16_swap_b32
     + buffer_store_dwordx4 (8 bf16 per store) instead of buffer_store_short.
     """
-    gemm, options = get_tagged_mxfp4_gemm_preshuffle_b(
+    gemm, options = get_tagged_mxfp4_gemm_preshuffle_b_wide_store(
         shape,
         block,
         wave_shape=(2, 2),
         reorder_workgroups=True,
-        output_dtype=tkl.bf16,
-        wide_stores=True,
     )
     dynamic_symbols = [tkl.sym.M, tkl.sym.N, tkl.sym.K]
     for sym in dynamic_symbols:
--- a/examples/python/7.1_schedule.py
+++ b/examples/python/7.1_schedule.py
@@ -434,7 +434,7 @@ def test_dbuf_4wave_mxfp_dynamic_preshuffle_b_gemm_wide_stores(
 
     Uses wide_stores=True to swap MFMA operands (B as LHS, A as RHS),
     aligning the accumulator's contiguous values with the output's stride-1
-    dimension. The coalesce_epilogue_stores pass emits v_permlane16_swap_b32
+    dimension. The coalesce_wide_stores pass emits v_permlane16_swap_b32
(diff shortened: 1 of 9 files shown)


examples/python/7.1_schedule.py
```diff
@@ -424,6 +424,46 @@ def test_dbuf_4wave_mxfp_dynamic_preshuffle_b_gemm(
...
+    dimension. The coalesce_wide_stores pass emits v_permlane16_swap_b32
+    + buffer_store_dwordx4 (8 bf16 per store) instead of buffer_store_short.
+    """
+    gemm, options = get_tagged_mxfp4_gemm_preshuffle_b(
+        shape,
+        block,
+        wave_shape=(2, 2),
+        reorder_workgroups=True,
+        output_dtype=tkl.bf16,
+        wide_stores=True,
```
panditsa: We don't need this option, since it's functional we should always enable it for all kernels that rely on `get_tagged_mxfp4_gemm_preshuffle_b`. Since this is functional in LLVM at the moment, we can create a variant of the kernel `get_tagged_mxfp4_gemm_preshuffle_b_wide_store`, and swap it out in existing examples that use LLVM.

panditsa: yeah using a tagged_mxfp4_gemm_preshuffle_b_wide_store is good, and we should enable it by default for the examples where it's applicable!

panditsa (approved):

PR #1283, opened 2026-04-08
Hoist loop-variant affine index ops out of scf.for during codegen

willghatch:
When dimensions are dynamic, the linearized index expressions passed to affine.apply contain the loop induction variable, preventing LICM from moving them out of the loop body.  Split each such expression into `base + iv * stride`, generate the base before the loop with gen_sympy_index_hoisted, and emit the IV-dependent part as plain arith ops that downstream LICM can reason about independently.

The key test is `lit_tests/kernel/wave/linearize_loop_affine_maps.py` which shows that no `affine.apply` ops occur inside a loop body.

Key changes:

- Add gen_sympy_index_hoisted (emitter.py): decomposes a sympy expression into a loop-invariant base (hoisted) and an IV-scaled offset (emitted in-place), with reconstruction guards that fall back to the original gen_sympy_index when the split is not provably safe.

- Refactor read_write.py: extract _get_enclosing_scf_for (walks parent regions), _hoist_before_loop, _iv_context, _gen_linear_index_offset, _hoist_dst_indices, and _linearize_memref_hoisted to replace ad-hoc hoisting scattered across _handle_read_linear_index and handle_gather_to_lds.

- Rework _build_mask to lower each bound condition individually with hoisting support, replacing the previous functools.reduce over a single sympy And expression.

- Add linearize_loop_affine_maps.py lit test verifying no affine.apply ops remain inside the pipelined scf.for body for two wave shapes.

- Update CHECK patterns in 5 existing lit tests to reflect the new code shape (fewer affine_map parameters, hoisted base ops before the loop, arith.muli/addi inside the loop).

Made-with: Cursor

### DIFF
--- a/lit_tests/kernel/wave/bounds_check.py
+++ b/lit_tests/kernel/wave/bounds_check.py
@@ -79,16 +79,15 @@ def test_bounds_check():
     # CHECK-LABEL:    test_bounds_check
     # CHECK:          func.func @read_write
     # CHECK:            %[[C227:.*]] = arith.constant 227 : index
-    #                   M row index (linearized), then N / vector mask setup
-    # CHECK:            %[[D2:.*]] = affine.apply #map{{.*}}()[%thread_id_x, %block_id_x]
-    # CHECK:            %[[D3:.*]] = arith.cmpi slt, %[[D2]], %[[C227]] : index
-    # CHECK:            vector.broadcast %[[D3]] : i1 to vector<16xi1>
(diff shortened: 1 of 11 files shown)


ftynse (commented):
Arguably, this should just be an "advanced LICM" MLIR pass rather than extra ad hoc logic in whatever emits MLIR. It is also unclear to me what is the value of having `affine.apply` to start with, expression simplification should have happened in sympy sufficiently.

harsh-nod (approved):

PR #1287, opened 2026-04-09
[water] module-level alias for WaveIndexMappingAttr

ftynse:
Use module-level alias for index mappings to decrease their verbosity in
the overall IR: only the alias name is printed at use sites and
identical mappings have the same alias. This is similar to affine maps.


tgymnich (approved):

water/lib/Dialect/Wave/IR/WaveDialect.cpp
```diff
@@ -27,13 +28,29 @@
...
 using namespace mlir;
 
+namespace {
+// Interface to support module-level aliases for WaveIndexMappingAttr.
+struct WaveOpAsmDialectInterface : public OpAsmDialectInterface {
+  using OpAsmDialectInterface::OpAsmDialectInterface;
+
+  AliasResult getAlias(Attribute attr, raw_ostream &os) const override {
+    if (isa<wave::WaveIndexMappingAttr>(attr)) {
+      os << "wim";
```
martin-luecke: On seeing this the first time I had a hard time immediately parsing "wim".
Different options for this could be: "imap", "idxm", or "idxmap" (easiest to parse for me, but long).
No strong opinion though.

martin-luecke (approved):
Really cool change to see! The attention kernel is now so much easier to read and comparing whether to indexing expressions are similar is now 1 cycle in my brain.

With this change and not using the custom parser/printer for WaveIndexDict on WaveOp anymore, is this code now dead and could be deleted?
https://github.com/iree-org/wave/blob/5e5fd76c15cb287b06c35172d0e20cc8033fd515/water/include/water/Dialect/Wave/IR/WaveInterfaces.h#L54-L57
https://github.com/iree-org/wave/blob/5e5fd76c15cb287b06c35172d0e20cc8033fd515/water/lib/Dialect/Wave/IR/WaveInterfaces.cpp#L235-L294

water/lib/Dialect/Wave/IR/WaveDialect.cpp
tgymnich: +1 for idxmap

ftynse: changed

PR #1288, opened 2026-04-09
GEMM support in waveasm LLVM path

Hardcode84:
## Summary
- Enable the water-to-WaveASM GEMM path by plumbing dynamic buffer strides through the runtime and host wrapper, preserving wrapper argument locations, and keeping structured loops available for `waveasm-translate`.
- Extend LLVM-to-WaveASM lowering for GEMM-related ops and tighten correctness around GEP byte offsets, LDS byte accounting, signed div/rem lowering, address-space handling, and `scf.for` loop control.
- Add coverage with focused WaveASM lit tests for supported and unsupported lowering cases, plus a CDNA4-gated end-to-end GEMM test.

## Notes
- `scf.for` lowering still carries a TODO for full zero-trip semantics because `waveasm.loop` is currently do-while shaped.

### DIFF
--- a/wave_lang/kernel/compiler/wave_codegen/emitter.py
+++ b/wave_lang/kernel/compiler/wave_codegen/emitter.py
@@ -71,6 +71,7 @@ from ..kernel_codegen import (
     BindingDesc,
     BoundKernelSignature,
     create_argument_locations,
+    get_dynamic_stride_arg_count,
 )
 from ...lang.wave_types import IndexSymbol
 from ...wave.compile_options import WaveCompileOptions
@@ -137,40 +138,70 @@ class WaveEmitter:
             ),
         ]
 
+    def _abi_type(self, binding: BindingDesc) -> IrType:
+        if binding.binding_type == BindingType.KERNEL_BUFFER:
+            # Buffer passed to kernel as 0D memrefs to simplify ABI.
+            element_type = IrType.parse(binding.kernel_buffer_type.dtype.ir_type_asm())
+            return MemRefType.get([], element_type=element_type)
+
+        # Scalars are passed as is.
+        return binding.as_mlir_type()
+
+    def _create_stride_argument_locations(
+        self, bindings: list[BindingDesc], base_locations: list[Location]
+    ) -> list[Location]:
+        if not self.options.dynamic_strides:
+            return []
+
+        stride_locations = []
+        for binding, base_location in zip(bindings, base_locations):
+            if binding.binding_type != BindingType.KERNEL_BUFFER:
+                continue
+            leading_count = max(0, len(binding.kernel_buffer_type.symbolic_shape) - 1)
+            argument_name = binding.name or "argument"
+            for dim_idx in range(leading_count):
+                stride_locations.append(
+                    Location.name(
+                        f"{argument_name}.stride.{dim_idx}",
+                        base_location,
+                    )
+                )
+        return stride_locations
+
+    def _ordered_kernel_buffer_bindings(
+        self, bindings: list[BindingDesc]
+    ) -> list[BindingDesc]:
+        return [
+            binding
+            for binding in bindings
+            if binding.binding_type == BindingType.KERNEL_BUFFER
+        ]
+
+    def _kernel_argument_abi(
+        self, bindings: list[BindingDesc]
+    ) -> tuple[list[IrType], list[Location], int]:
+        arg_types = [self._abi_type(binding) for binding in bindings]
+        arg_locations = create_argument_locations(bindings)
+        kernel_buffer_bindings = self._ordered_kernel_buffer_bindings(bindings)
+        stride_arg_count = get_dynamic_stride_arg_count(
+            self.options.dynamic_strides,
+            kernel_buffer_bindings,
+        )
+        stride_locations = self._create_stride_argument_locations(
+            bindings, arg_locations
+        )
+        assert len(stride_locations) == stride_arg_count
+        if stride_arg_count > 0:
+            arg_types += [IndexType.get()] * stride_arg_count
+            arg_locations += stride_locations
+        return arg_types, arg_locations, stride_arg_count
+
     def emit_func(self) -> Operation:
         bindings = self.root_sig.sig.linear_bindings
-
-        def abi_type(binding: BindingDesc):
-            if binding.binding_type == BindingType.KERNEL_BUFFER:
-                # Buffer passed to kernel as 0D memrefs to simplify ABI.
-                element_type = IrType.parse(
-                    binding.kernel_buffer_type.dtype.ir_type_asm()
-                )
-                return MemRefType.get([], element_type=element_type)
-
-            # Scalars are passed as is.
-            return binding.as_mlir_type()
-
-        arg_types = [abi_type(b) for b in bindings]
-
-        # Dynamic strides only with Wave runtime and LLVM backend (not ASM).
-        # Stride args only for leading dimensions (innermost always has unit stride).
-        stride_arg_count = 0
-        if self.options.dynamic_strides:
-            stride_arg_count = sum(
-                max(0, len(b.kernel_buffer_type.symbolic_shape) - 1)
-                for b in self.root_sig.sig.kernel_buffer_bindings
-            )
-            if stride_arg_count > 0:
-                arg_types += [IndexType.get()] * stride_arg_count
-
+        arg_types, arg_locations, stride_arg_count = self._kernel_argument_abi(bindings)
         ftype = FunctionType.get(arg_types, [])
         func_op = func_d.FuncOp(self.kernel_name, ftype, visibility="private")
-
-        locs = create_argument_locations(bindings)
-        if stride_arg_count > 0:
-            locs += [Location.unknown()] * stride_arg_count
-        entry_block = func_op.add_entry_block(locs)
+        entry_block = func_op.add_entry_block(arg_locations)
 
         # Map dynamic symbols to buffer argument indices and dimensions.
         for bind, arg in zip(bindings, entry_block.arguments):
@@ -285,8 +316,9 @@ class WaveEmitter:
         return func_op, symbol
 
     def emit_host_func(self, kernel_func: Operation) -> Operation:
-        # TODO: kernel bindings order may not be the same as the kernel function
-        # arguments order, so map kernel order to host function arguments order.
+        # Host placeholders follow the original Python signature, while kernel ABI
+        # arguments use linear_bindings order. Track both so launch args can be
+        # materialized in kernel ABI order from host-visible inputs.
         binding_map = {}
         symbol_map = {}
 
@@ -308,35 +340,10 @@ class WaveEmitter:
         bindings = self.root_sig.sig.linear_bindings
 
         ptr = llvm_d.PointerType.get()
-
-        def abi_type(binding: BindingDesc):
-            if binding.binding_type == BindingType.KERNEL_BUFFER:
-                # Buffer passed to kernel as 0D memrefs to simplify ABI.
-                element_type = IrType.parse(
-                    binding.kernel_buffer_type.dtype.ir_type_asm()
-                )
-                return MemRefType.get([], element_type=element_type)
-
-            # Scalars are passed as is.
-            return binding.as_mlir_type()
-
-        arg_types = [abi_type(b) for b in bindings]
-
-        # Match stride args added by emit_func when dynamic_strides is active.
-        stride_arg_count = 0
-        if self.options.dynamic_strides:
-            stride_arg_count = sum(
-                max(0, len(b.kernel_buffer_type.symbolic_shape) - 1)
-                for b in self.root_sig.sig.kernel_buffer_bindings
-            )
-            if stride_arg_count > 0:
-                arg_types += [IndexType.get()] * stride_arg_count
-
+        arg_types, kernel_arg_locations, stride_arg_count = self._kernel_argument_abi(
+            bindings
+        )
         ftype = FunctionType.get(arg_types, [])
-        # Preserve source locations for original args; synthesized stride args
-        # do not have an originating kernel SSA value.
-        locs = [a.location for a in kernel_func.body.blocks[0].arguments]
-        locs += [Location.unknown()] * (len(arg_types) - len(locs))
 
         gpu_module = gpu_d.module("gpu_module")
         gpu_module.parent.operation.attributes["gpu.container_module"] = UnitAttr.get()
@@ -358,7 +365,7 @@ class WaveEmitter:
 
         new_kernel_entry_block = kernel_func_wrapper.body.blocks.append(
             *arg_types,
-            arg_locs=locs,
+            arg_locs=kernel_arg_locations,
         )
 
         # Inline the kernel function into the gpu module function body and erase the original function
@@ -512,7 +519,7 @@ class WaveEmitter:
             # added by emit_func. One stride per leading dimension (rank-1)
             # for each kernel buffer.
             if self.options.dynamic_strides:
-                for binding in self.root_sig.sig.kernel_buffer_bindings:
+                for binding in self._ordered_kernel_buffer_bindings(bindings):
                     rank = len(binding.kernel_buffer_type.symbolic_shape)
                     leading_count = max(0, rank - 1)
                     arg = func_args[binding_map[id(binding)]]
--- a/wave_lang/kernel/compiler/wave_codegen/emitter.py
+++ b/wave_lang/kernel/compiler/wave_codegen/emitter.py
@@ -333,7 +333,10 @@ class WaveEmitter:
                 arg_types += [IndexType.get()] * stride_arg_count
 
         ftype = FunctionType.get(arg_types, [])
-        locs = [Location.unknown() for _ in arg_types]
+        # Preserve source locations for original args; synthesized stride args
+        # do not have an originating kernel SSA value.
+        locs = [a.location for a in kernel_func.body.blocks[0].arguments]
+        locs += [Location.unknown()] * (len(arg_types) - len(locs))
 
         gpu_module = gpu_d.module("gpu_module")
         gpu_module.parent.operation.attributes["gpu.container_module"] = UnitAttr.get()
--- a/wave_lang/kernel/compiler/wave_codegen/emitter.py
+++ b/wave_lang/kernel/compiler/wave_codegen/emitter.py
@@ -322,8 +322,18 @@ class WaveEmitter:
 
         arg_types = [abi_type(b) for b in bindings]
 
+        # Match stride args added by emit_func when dynamic_strides is active.
+        stride_arg_count = 0
+        if self.options.dynamic_strides:
+            stride_arg_count = sum(
+                max(0, len(b.kernel_buffer_type.symbolic_shape) - 1)
+                for b in self.root_sig.sig.kernel_buffer_bindings
+            )
+            if stride_arg_count > 0:
+                arg_types += [IndexType.get()] * stride_arg_count
+
         ftype = FunctionType.get(arg_types, [])
-        locs = [a.location for a in kernel_func.body.blocks[0].arguments]
+        locs = [Location.unknown() for _ in arg_types]
 
         gpu_module = gpu_d.module("gpu_module")
         gpu_module.parent.operation.attributes["gpu.container_module"] = UnitAttr.get()
@@ -394,6 +404,11 @@ class WaveEmitter:
 
         get_float64_func, get_float64_func_symbol = self._declare_runtime_func(
             "wave_get_float64", [ptr], [f64], emit_c_interface=True
+        )
+
+        # Get tensor stride function from PyObject*.
+        get_stride_func, get_stride_func_symbol = self._declare_runtime_func(
+            "wave_get_stride", [ptr, i32], [i64], emit_c_interface=True
         )
 
         # Declare host function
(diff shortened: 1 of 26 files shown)


wave_lang/kernel/wave/water.py
```diff
@@ -390,6 +390,27 @@ def add_opt(pipeline):
...
 
+    def add_transform(transform: str, entry_point: str) -> tuple[str, dict[str, Any]]:
+        nonlocal mlir_asm
+        # Erase the last '}' closing the module, append the transform, re-close.
+        last_close = mlir_asm.rfind("}")
+        if last_close != -1:
+            mlir_asm = mlir_asm[:last_close]
+        mlir_asm += transform
+        mlir_asm += "}\n"
+        return ("transform-interpreter", {"entry-point": entry_point})
```
ftynse: Doing an in-place fragile textual modification of a non-local variable is a terrible idea. We have API to modify IR properly. And I think we have enough bindings to build the script below rather than hardcode a string.

wave_lang/kernel/compiler/wave_codegen/emitter.py
```diff
@@ -322,8 +322,21 @@ def abi_type(binding: BindingDesc):
...
 
+        # Match stride args added by emit_func when dynamic_strides is active.
+        stride_arg_count = 0
+        if self.options.dynamic_strides:
+            stride_arg_count = sum(
+                max(0, len(b.kernel_buffer_type.symbolic_shape) - 1)
+                for b in self.root_sig.sig.kernel_buffer_bindings
+            )
+            if stride_arg_count > 0:
+                arg_types += [IndexType.get()] * stride_arg_count
```
ftynse: Can this logic be factored out into a function also called from `emit_func` so we don't duplicate and risk divergence?

wave_lang/kernel/compiler/wave_codegen/emitter.py
```diff
@@ -322,8 +322,21 @@ def abi_type(binding: BindingDesc):
...
+
         ftype = FunctionType.get(arg_types, [])
+        # Preserve source locations for original args; synthesized stride args
+        # do not have an originating kernel SSA value.
         locs = [a.location for a in kernel_func.body.blocks[0].arguments]
+        locs += [Location.unknown()] * (len(arg_types) - len(locs))
```
ftynse: IMO they can have something line `Location.name("stride #42 for argument", arg.loc)` so we clearly see what they relate to in the debug build.

wave_lang/kernel/compiler/wave_codegen/emitter.py
```diff
@@ -490,6 +508,24 @@ def abi_type(binding: BindingDesc):
...
+                    leading_count = max(0, rank - 1)
+                    arg = func_args[binding_map[id(binding)]]
+                    for dim_idx in range(leading_count):
+                        dim = arith_d.constant(i32, dim_idx)
+                        stride = func_d.call(
+                            get_stride_func.type.results,
+                            get_stride_func_symbol,
+                            [arg, dim],
+                        )
+                        stride = arith_d.index_cast(IndexType.get(), stride)
```
ftynse, line 533: Should we just pass it as i32 and not bother with index?

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
```diff
@@ -187,12 +188,93 @@ static Value truncToI32(Value v, Type llvmType, OpBuilder &builder,
...
+static int64_t getLLVMTypeBytes(Type ty) {
+  if (ty.isIntOrFloat())
+    return ty.getIntOrFloatBitWidth() / 8;
+  if (VectorType vecTy = dyn_cast<VectorType>(ty))
+    return vecTy.getNumElements() *
+           vecTy.getElementType().getIntOrFloatBitWidth() / 8;
+  if (LLVM::LLVMArrayType arrTy = dyn_cast<LLVM::LLVMArrayType>(ty))
+    return getLLVMTypeBytes(arrTy.getElementType()) * arrTy.getNumElements();
+  return 0;
+}
```
ftynse: Don't we have data layout support for this? It may in particular be consistent with things like sub-byte element types and non-power-of-two vector lengths.

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
```diff
@@ -187,12 +188,93 @@ static Value truncToI32(Value v, Type llvmType, OpBuilder &builder,
...
+
+/// Return true iff a GEP index is statically known to be zero.
+static bool isZeroGEPIndex(llvm::PointerUnion<IntegerAttr, Value> idx) {
+  if (isa<Value>(idx))
+    return isConstantIntValue(cast<Value>(idx), 0);
```
ftynse: isa followed by a cast is a an anti-pattern, use dyn_cast

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
```diff
@@ -187,12 +188,93 @@ static Value truncToI32(Value v, Type llvmType, OpBuilder &builder,
...
+  return isConstantIntValue(cast<IntegerAttr>(idx), 0);
+}
+
+/// Structural GEPs like [0, 0] are a no-op and can be forwarded even though we
+/// do not model nested aggregate layouts yet.
+static bool isAllZeroIndexGEP(LLVM::GEPOp op) {
+  for (llvm::PointerUnion<IntegerAttr, Value> idx : op.getIndices())
+    if (!isZeroGEPIndex(idx))
+      return false;
+  return true;
```
ftynse: llvm::all_of

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
```diff
@@ -187,12 +188,93 @@ static Value truncToI32(Value v, Type llvmType, OpBuilder &builder,
...
+/// Returns 0 for unsupported types.
+static int64_t getLLVMTypeBytes(Type ty) {
+  if (ty.isIntOrFloat())
+    return ty.getIntOrFloatBitWidth() / 8;
+  if (VectorType vecTy = dyn_cast<VectorType>(ty))
+    return vecTy.getNumElements() *
+           vecTy.getElementType().getIntOrFloatBitWidth() / 8;
+  if (LLVM::LLVMArrayType arrTy = dyn_cast<LLVM::LLVMArrayType>(ty))
+    return getLLVMTypeBytes(arrTy.getElementType()) * arrTy.getNumElements();
+  return 0;
```
ftynse: I'd rather return explicit failure

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
```diff
@@ -316,17 +414,22 @@ static LogicalResult handleWorkgroupId(OpTy op, LLVMTranslationState &st,
...
   auto &ctx = st.ctx;
   auto &builder = ctx.getBuilder();
   FailureOr<Value> src = resolve(op.getOperand(), ctx);
   if (failed(src))
     return op->emitOpError("unmapped operand in cast");
-  Type resTy = isVGPRType(src->getType()) ? (Type)ctx.createVRegType()
-                                          : (Type)ctx.createSRegType();
+  int64_t width = 1;
+  if (IntegerType intTy = dyn_cast<IntegerType>(op.getResult().getType()))
+    width = (intTy.getWidth() + 31) / 32;
```
ftynse: we should have a named helper for ceilDiv somewhere. Also explain where the magic 32 is coming from.

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
```diff
@@ -491,37 +596,64 @@ static LogicalResult handleGEP(LLVM::GEPOp op, LLVMTranslationState &st) {
...
+
+  // LDS GEP (ptr<3>): compute byte offset for ds_read/ds_write.
+  // DS instructions only accept a 32-bit vaddr, so truncate wide offsets.
+  if (addrSpace == 3) {
+    if (!isSingleIndex && !isAllZeroGEP)
+      return op->emitOpError(
+          "LDS GEP must have a single index or all-zero structural indices");
+    FailureOr<Value> baseOff = resolve(base, ctx);
+    if (failed(baseOff))
+      return op->emitOpError("unmapped LDS GEP base");
```
ftynse, line 706: Can this happen with valid input and not be caught before? If not, it should be an assertion. There are several other places like this.

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
```diff
@@ -647,6 +801,395 @@ static LogicalResult handleStore(LLVM::StoreOp op, LLVMTranslationState &st) {
...
 
+//===----------------------------------------------------------------------===//
+// LDS global / addressof handlers
+//===----------------------------------------------------------------------===//
+
+static LogicalResult handleAddressOf(LLVM::AddressOfOp op,
+                                     LLVMTranslationState &st) {
+  TranslationContext &ctx = st.ctx;
+  // Look up the global to determine LDS size.
+  LLVM::GlobalOp global = SymbolTable::lookupNearestSymbolFrom<LLVM::GlobalOp>(
```
ftynse: Store a symbol table object in the context that is passed around everywhere to avoid reconstructing it every single time, with the associated cost.

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
```diff
@@ -647,6 +801,395 @@ static LogicalResult handleStore(LLVM::StoreOp op, LLVMTranslationState &st) {
...
+      ctx.addLDSSize(sizeBytes);
+  }
+
+  // Map to a constant 0 offset in a VGPR -- LDS addressing is relative.
+  OpBuilder &builder = ctx.getBuilder();
+  ImmType immTy = ctx.createImmType(0);
+  Value zero = ConstantOp::create(builder, op.getLoc(), immTy, int64_t{0});
+  VRegType vregTy = ctx.createVRegType();
+  Value mov = V_MOV_B32::create(builder, op.getLoc(), vregTy, zero);
+  ctx.getMapper().mapValue(op.getResult(), mov);
```
ftynse: I don't understand the reasoning here. Every addressof will seemingly get turned into a address 0, how do we differentiate objects then?

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
```diff
@@ -647,6 +801,395 @@ static LogicalResult handleStore(LLVM::StoreOp op, LLVMTranslationState &st) {
...
+                                   LLVMTranslationState &st) {
+  TranslationContext &ctx = st.ctx;
+  OpBuilder &builder = ctx.getBuilder();
+  Location loc = op.getLoc();
+
+  int64_t numBytes = getBufferAccessBytes(op.getResult().getType());
+  FailureOr<Value> offset = resolve(addr, ctx);
+  if (failed(offset))
+    return op->emitOpError("unmapped LDS address");
+  // DS instructions use a 32-bit vaddr. Truncate wide offsets.
```
ftynse: Do we know this is safe? Or is there a _documented_ UB on overflow?

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
```diff
@@ -647,6 +801,395 @@ static LogicalResult handleStore(LLVM::StoreOp op, LLVMTranslationState &st) {
...
+  if (getRegSize(offset->getType()) > 1)
+    *offset = ArithTruncOp::create(builder, loc, vregTy, *offset);
+
+  Operation *loadOp = nullptr;
+  if (numBytes == 2)
+    loadOp = DS_READ_U16::create(builder, loc, TypeRange{vregTy}, *offset);
+  else if (numBytes == 4)
+    loadOp = DS_READ_B32::create(builder, loc, TypeRange{vregTy}, *offset);
+  else if (numBytes == 8) {
+    VRegType wideTy = ctx.createVRegType(2, 2);
```
ftynse, line 966: Seeing this pattern regularly, would it make sense to have a `createVRegTypeSameAlign` rather than magically repeating identical arguments every time?

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
```diff
@@ -647,6 +801,395 @@ static LogicalResult handleStore(LLVM::StoreOp op, LLVMTranslationState &st) {
...
+    loadOp = DS_READ_B64::create(builder, loc, TypeRange{wideTy}, *offset);
+  } else if (numBytes == 16) {
+    VRegType wideTy = ctx.createVRegType(4, 4);
+    loadOp = DS_READ_B128::create(builder, loc, TypeRange{wideTy}, *offset);
+  } else
```
ftynse: Nit: asymmetric parentheses are ugly

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
```diff
@@ -647,6 +801,395 @@ static LogicalResult handleStore(LLVM::StoreOp op, LLVMTranslationState &st) {
...
+  LLVM::GlobalOp global = SymbolTable::lookupNearestSymbolFrom<LLVM::GlobalOp>(
+      op, op.getGlobalNameAttr());
+  if (global) {
+    Type globalTy = global.getType();
+    int64_t sizeBytes = getLLVMTypeBytes(globalTy);
+    if (sizeBytes > 0)
+      ctx.addLDSSize(sizeBytes);
+  }
+
+  // Map to a constant 0 offset in a VGPR -- LDS addressing is relative.
```
ftynse: I don't see a limitation of this function to LDS specifically.

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
```diff
@@ -647,6 +801,395 @@ static LogicalResult handleStore(LLVM::StoreOp op, LLVMTranslationState &st) {
...
+  if (getRegSize(lhs->getType()) != 1)
+    return op->emitOpError(
+        "signed division currently supports only i32 operands");
+
+  // Signed division by a positive power-of-2 constant:
+  //   q = (x + (x < 0 ? divisor - 1 : 0)) >> log2(divisor)
+  // This preserves LLVM's trunc-toward-zero semantics for negative dividends.
+  if (std::optional<int64_t> constVal = getConstantIntValue(op.getRhs())) {
+    int64_t divisor = *constVal;
+    if (divisor > 0 && (divisor & (divisor - 1)) == 0) {
```
ftynse: There is an `isPowerOf2` in LLVM, no need to be bit-smart.

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
```diff
@@ -647,6 +801,395 @@ static LogicalResult handleStore(LLVM::StoreOp op, LLVMTranslationState &st) {
...
+      int64_t shiftAmt = llvm::Log2_64(divisor);
```
ftynse: If you really wanted to be bit smart, you would call "find first set" bit here, LLVM has that too, and it compiles down to one instruction on supported architectures.

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
```diff
@@ -647,6 +801,395 @@ static LogicalResult handleStore(LLVM::StoreOp op, LLVMTranslationState &st) {
...
+  FailureOr<Value> lhs = resolve(op.getLhs(), ctx);
+  if (failed(lhs))
+    return op->emitOpError("unmapped operand in sdiv");
+  if (getRegSize(lhs->getType()) != 1)
+    return op->emitOpError(
+        "signed division currently supports only i32 operands");
+
+  // Signed division by a positive power-of-2 constant:
+  //   q = (x + (x < 0 ? divisor - 1 : 0)) >> log2(divisor)
+  // This preserves LLVM's trunc-toward-zero semantics for negative dividends.
```
ftynse: This logic has nothing to do in lowering. It's a strength reduction, perhaps even a form of canonicalization and constant propagation, that should happen before we change the abstraction level and can easily be tested separately.

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
```diff
@@ -647,6 +801,395 @@ static LogicalResult handleStore(LLVM::StoreOp op, LLVMTranslationState &st) {
...
+  if (failed(lhs))
+    return op->emitOpError("unmapped operand in srem");
+  if (getRegSize(lhs->getType()) != 1)
+    return op->emitOpError(
+        "signed remainder currently supports only i32 operands");
+
+  // Signed remainder by a positive power-of-2 constant:
+  //   r = x & (divisor - 1)
+  //   if (x < 0 && r != 0) r -= divisor
+  // This keeps the remainder's sign consistent with LLVM srem.
```
ftynse: Ditto

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
```diff
@@ -647,6 +801,395 @@ static LogicalResult handleStore(LLVM::StoreOp op, LLVMTranslationState &st) {
...
+  Value mfma = V_MFMA_F32_16X16X16_F16::create(builder, loc, accTy, *a, *b, *c);
+  ctx.getMapper().mapValue(op.getResult(), mfma);
+  return success();
+}
+
+//===----------------------------------------------------------------------===//
+// SCF for/yield handler
+//===----------------------------------------------------------------------===//
+
+/// Forward declaration for recursive op translation.
```
ftynse: This is anti-documentation. It will show up in, e.g., autocomplete, and provide negative value compared to no documentation at all.

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
```diff
@@ -647,6 +801,395 @@ static LogicalResult handleStore(LLVM::StoreOp op, LLVMTranslationState &st) {
...
+  }
+  SRegType sregTy = ctx.createSRegType();
+  if (isSGPRType(v.getType()))
+    return v;
+  if (isImmType(v.getType()))
+    return S_MOV_B32::create(builder, loc, sregTy, v).getResult();
+  if (isVGPRType(v.getType()))
+    return V_READFIRSTLANE_B32::create(builder, loc, sregTy, v).getResult();
+  op->emitOpError(name) << " must lower to an SGPR or VGPR i32";
+  return failure();
```
ftynse, line 1107: Why not return `Value()`?

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
```diff
@@ -647,6 +801,395 @@ static LogicalResult handleStore(LLVM::StoreOp op, LLVMTranslationState &st) {
...
+  FailureOr<Value> stepScalar =
+      materializeI32SReg(*step, "scf.for step", op, ctx, builder, loc);
+  if (failed(lbScalar) || failed(ubScalar) || failed(stepScalar))
+    return failure();
+
+  // Build init args: [lower_bound, iter_args...].
+  SmallVector<Value> initArgs;
+  initArgs.push_back(*lbScalar);
+  for (Value arg : op.getInitArgs()) {
+    FailureOr<Value> resolved = resolve(arg, ctx);
```
ftynse, line 1166: Fly-by, there is rarely a need to wrap a nullable object into `FailureOr`, so I suspect `resolve` should be fixed.

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
```diff
@@ -647,6 +801,395 @@ static LogicalResult handleStore(LLVM::StoreOp op, LLVMTranslationState &st) {
...
+
+  // Create the waveasm.loop (do-while semantics).
+  LoopOp loopOp = LoopOp::create(builder, loc, initArgs);
+  Block &bodyBlock = loopOp.getBodyBlock();
+
+  // Map the induction variable (block arg 0).
+  ctx.getMapper().mapValue(op.getInductionVar(), bodyBlock.getArgument(0));
+
+  // Map iter_args (block args 1..N).
+  for (unsigned i = 0; i < op.getInitArgs().size(); ++i)
```
ftynse: Nit: don't evaluate the upper bound on every iteration.

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
Hardcode84: Can be potentially triggered with malformed IR with broken SSA.

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
Hardcode84: LDS addresses are always fit into 32bit (and much smaller in practice), values not fitting into 32bit are upstream program UB, like any other invalid pointer math.

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
Hardcode84: Overall comment on the `FailureOr<Value>` vs `Value`, I saw agent was doing it and I intentionally didn't stopped it. In upstream MLIR there are some functions which never return empty `Value` and some which do and their result needed to be checked (or worse, historically, some function were switched from never returning null to returning), which is footgun. `FailureOr` gives a clear separation on type system level.

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
Hardcode84: I can probably plumb it through our legalization pass.

wave_lang/kernel/compiler/wave_codegen/emitter.py
Hardcode84: We can, probably, but it will require more changes across the codebase (`StreamExecutable.define_entrypoint` and other places), lets keep it as index for now.

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
Hardcode84: Not sure DataLayout is currently plumbed

wave_lang/kernel/wave/water.py
Hardcode84: done

wave_lang/kernel/compiler/wave_codegen/emitter.py
Hardcode84: done

wave_lang/kernel/compiler/wave_codegen/emitter.py
Hardcode84: done

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
Hardcode84: done

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
Hardcode84: fixed

waveasm/lib/Transforms/TranslateFromLLVMDialect.cpp
Hardcode84: we can probably just overload `createVRegType` for single argument instead, but it will be too much churn for this PR.

PR #1292, opened 2026-04-13
[water] Add --mlir-edit-ir-{before,after,all} pass instrumentation

### LINKED ISSUE #1291: [water-opt] Add interactive IR editing instrumentation (--mlir-edit-ir-after)
Add interactive IR editing to water-opt, mirroring --mlir-print-ir-{before,after} but for editing:
- --mlir-edit-ir-before=<pass> / --mlir-edit-ir-before-all pauses before the named pass (or all passes), writes IR to a temp file, opens it in an editor, and resumes with the user's edits.
- --mlir-edit-ir-after=<pass> / --mlir-edit-ir-after-all does the same after.
- --mlir-edit-ir-editor=<cmd> overrides the editor (default: $EDITOR).

WIP -- first prototype implementation is functional, will clean up and add tests.

martin-luecke:
Add interactive IR editing via PassInstrumentation. After (or before) a specified pass, the current IR is written to a temporary file and compilation blocks until the user presses Enter, allowing manual edits to the IR mid-pipeline.

- `--mlir-edit-ir-after=<pass>` / `--mlir-edit-ir-before=<pass>` for a specific pass
- `--mlir-edit-ir-after-all` / `--mlir-edit-ir-before-all` for every pass
- `--mlir-edit-ir-editor=<cmd>` to specify the editor to auto-open the file (default: `$EDITOR`)
- Python `WaveCompileOptions` extended with matching fields
- Lit test covering selective firing, all-pass variants, and IR round-trip

Fixes #1291

### DIFF
--- a/water/lib/Transforms/EditIR.cpp
+++ b/water/lib/Transforms/EditIR.cpp
@@ -85,54 +85,66 @@ LogicalResult mlir::water::editIRInteractively(Operation *op, StringRef editor,
   if (!editorCmd.empty())
     tryOpenInEditor(editorCmd, tmpPath, op);
 
-  // Prompt the user and wait for Enter (or EOF).
   llvm::outs() << "=== water-edit-ir";
   if (!passLabel.empty())
     llvm::outs() << " " << passLabel;
   llvm::outs() << " ===\n"
                << "IR written to: " << tmpPath << "\n"
-               << "Edit the file, then press Enter to continue "
-                  "compilation...\n";
+               << "Press Enter to continue, or 'q' to abort compilation...\n";
   llvm::outs().flush();
 
-  // fgets instead of std::cin to avoid pulling in <iostream> (static
-  // initializers). Buffer only needs to hold '\n' + '\0'.
-  char buf[2];
-  (void)std::fgets(buf, sizeof(buf), stdin);
+  while (true) {
+    int first = std::getchar();
+    if (first == EOF || first == 'q' || first == 'Q')
+      return failure();
+    // Drain the rest of the line so nothing leaks into the next iteration.
+    for (int c = first; c != '\n' && c != EOF;)
+      c = std::getchar();
 
-  // Re-parse the (potentially edited) file.
-  auto fileOrErr = llvm::MemoryBuffer::getFile(tmpPath);
-  if (!fileOrErr)
-    return op->emitError() << "failed to read edited file: "
-                           << fileOrErr.getError().message();
+    auto fileOrErr = llvm::MemoryBuffer::getFile(tmpPath);
+    if (!fileOrErr)
+      return op->emitError() << "failed to read edited file: "
+                             << fileOrErr.getError().message();
 
-  llvm::SourceMgr sourceMgr;
-  sourceMgr.AddNewSourceBuffer(std::move(*fileOrErr), SMLoc());
-  SourceMgrDiagnosticHandler diagHandler(sourceMgr, context);
+    llvm::SourceMgr sourceMgr;
+    sourceMgr.AddNewSourceBuffer(std::move(*fileOrErr), SMLoc());
 
-  OwningOpRef<Operation *> parsedOp =
-      parseSourceFile<Operation *>(sourceMgr, context);
-  if (!parsedOp)
-    return failure(); // Parser already emitted diagnostics.
+    std::string diagStr;
+    llvm::raw_string_ostream diagOS(diagStr);
+    SourceMgrDiagnosticHandler diagHandler(sourceMgr, context, diagOS);
 
-  if (op->getName() != (*parsedOp)->getName())
-    return op->emitError()
-           << "edited IR has a different top-level operation ('"
-           << (*parsedOp)->getName() << "' vs '" << op->getName() << "')";
+    OwningOpRef<Operation *> parsedOp =
+        parseSourceFile<Operation *>(sourceMgr, context);
 
-  // Replace the operation's regions, attributes, and properties.
-  unsigned numRegions = op->getNumRegions();
-  for (unsigned i = 0; i < numRegions; ++i) {
-    Region &existing = op->getRegion(i);
-    Region &parsed = (*parsedOp)->getRegion(i);
+    if (parsedOp) {
+      if (op->getName() != (*parsedOp)->getName())
+        return op->emitError()
+               << "edited IR has a different top-level operation ('"
+               << (*parsedOp)->getName() << "' vs '" << op->getName() << "')";
 
-    existing.getBlocks().clear();
-    existing.getBlocks().splice(existing.getBlocks().end(), parsed.getBlocks());
+      unsigned numRegions = op->getNumRegions();
+      for (unsigned i = 0; i < numRegions; ++i) {
+        Region &existing = op->getRegion(i);
+        Region &parsed = (*parsedOp)->getRegion(i);
+        existing.getBlocks().clear();
+        existing.getBlocks().splice(existing.getBlocks().end(),
+                                    parsed.getBlocks());
+      }
+      op->setAttrs((*parsedOp)->getAttrDictionary());
+      op->copyProperties((*parsedOp)->getPropertiesStorage());
+      return success();
+    }
+
+    // Parse failed -- print diagnostics and re-prompt.
+    // The file is left untouched so that line numbers remain correct.
+    llvm::outs() << "Failed to parse edited IR:\n"
+                 << diagStr
+                 << "Press Enter to retry, or 'q' to abort compilation...\n";
+    llvm::outs().flush();
+
+    if (!editorCmd.empty())
+      tryOpenInEditor(editorCmd, tmpPath, op);
   }
-
-  op->setAttrs((*parsedOp)->getAttrDictionary());
-  op->copyProperties((*parsedOp)->getPropertiesStorage());
-  return success();
 }
 
 //===----------------------------------------------------------------------===//
--- a/water/lib/Transforms/EditIR.cpp
+++ b/water/lib/Transforms/EditIR.cpp
@@ -77,7 +77,7 @@ LogicalResult mlir::water::editIRInteractively(Operation *op, StringRef editor,
              << "failed to open temporary file for writing: " << ec.message();
     if (!passLabel.empty())
       tmpFile << "// -----// IR Edit " << passLabel << " //----- //\n";
-    op->print(tmpFile);
+    op->print(tmpFile, OpPrintingFlags().enableDebugInfo());
     tmpFile << '\n';
   }
 
--- a/water/lib/Transforms/EditIR.cpp
+++ b/water/lib/Transforms/EditIR.cpp
@@ -10,7 +10,6 @@
 #include "mlir/Parser/Parser.h"
 #include "mlir/Pass/Pass.h"
 
-#include "llvm/Support/Debug.h"
 #include "llvm/Support/FileSystem.h"
 #include "llvm/Support/FileUtilities.h"
 #include "llvm/Support/MemoryBuffer.h"
@@ -21,8 +20,6 @@
 
 #include <cstdio>
 #include <string>
-
-#define DEBUG_TYPE "edit-ir"
 
 using namespace mlir;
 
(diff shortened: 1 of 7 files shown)


water/lib/Transforms/EditIR.cpp
```diff
@@ -0,0 +1,174 @@
...
+               << "edit-ir: editor '" << editorCmd << "' not found in PATH\n");
+    return;
+  }
+
+  SmallVector<StringRef> args = {editorCmd, path};
+  std::string errMsg;
+  llvm::sys::ExecuteNoWait(*program, args, /*Env=*/std::nullopt,
+                           /*Redirects=*/{}, /*MemoryLimit=*/0, &errMsg);
+  if (!errMsg.empty())
+    LLVM_DEBUG(llvm::dbgs()
```
ftynse: Nit: `LDBG() <<`

water/lib/Transforms/EditIR.cpp
```diff
@@ -0,0 +1,174 @@
...
+  if (auto env = llvm::sys::Process::GetEnv("EDITOR"))
+    return *env;
+  return "";
+}
+
+/// Try to open `path` in `editorCmd`. Failures are non-fatal (the user can
+/// always open the file manually via the printed path).
+static void tryOpenInEditor(StringRef editorCmd, StringRef path) {
+  auto program = llvm::sys::findProgramByName(editorCmd);
+  if (!program) {
```
ftynse: Prefer an explicit emptiness check

water/lib/Transforms/EditIR.cpp
```diff
@@ -0,0 +1,174 @@
...
+    LLVM_DEBUG(llvm::dbgs()
+               << "edit-ir: editor '" << editorCmd << "' not found in PATH\n");
```
ftynse: I'd say we want to emit a diagnostic here rather than silently return. Same below.

water/lib/Transforms/EditIR.cpp
```diff
@@ -0,0 +1,174 @@
...
+  {
+    llvm::raw_fd_ostream tmpFile(tmpPath, ec);
+    if (ec)
+      return op->emitError()
+             << "failed to open temporary file for writing: " << ec.message();
+    if (!passLabel.empty())
+      tmpFile << "// -----// IR Edit " << passLabel << " //----- //\n";
+    op->print(tmpFile);
+    tmpFile << '\n';
+  }
```
ftynse: Why do we need a scope here?

water/lib/Transforms/EditIR.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+  std::string editorCmd = resolveEditor(editor);
+  if (!editorCmd.empty())
+    tryOpenInEditor(editorCmd, tmpPath);
+
+  // Prompt the user and wait for Enter (or EOF).
+  llvm::errs() << "=== mlir-edit-ir";
```
ftynse: Normal output should go to `llvm::outs`, `errs` is for errors

water/lib/Transforms/EditIR.cpp
```diff
@@ -0,0 +1,174 @@
...
+    llvm::errs() << " " << passLabel;
+  llvm::errs() << " ===\n"
+               << "IR written to: " << tmpPath << "\n"
+               << "Edit the file, then press Enter to continue "
+                  "compilation...\n";
+
+  // fgets instead of std::cin to avoid pulling in <iostream> (static
+  // initializers).
+  char buf[64];
+  (void)std::fgets(buf, sizeof(buf), stdin);
```
ftynse: Why 64 chars if we only wait for one?

water/lib/Transforms/EditIR.cpp
```diff
@@ -0,0 +1,174 @@
...
+                           << fileOrErr.getError().message();
+
+  llvm::SourceMgr sourceMgr;
+  sourceMgr.AddNewSourceBuffer(std::move(*fileOrErr), SMLoc());
+  SourceMgrDiagnosticHandler diagHandler(sourceMgr, context);
+
+  OwningOpRef<Operation *> parsedOp =
+      parseSourceFile<Operation *>(sourceMgr, context);
+  if (!parsedOp)
+    return op->emitError("failed to parse edited IR");
```
ftynse: Normally I expect the parser to have emitted a diagnostic already

water/lib/Transforms/EditIR.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+  // Replace the operation's regions and attributes with the parsed content.
+  for (unsigned i = 0; i < op->getNumRegions(); ++i) {
```
ftynse: Nit: don't evaluate the end condition on every iteration

water/lib/Transforms/EditIR.cpp
```diff
@@ -0,0 +1,174 @@
...
+  llvm::SourceMgr sourceMgr;
+  sourceMgr.AddNewSourceBuffer(std::move(*fileOrErr), SMLoc());
+  SourceMgrDiagnosticHandler diagHandler(sourceMgr, context);
+
+  OwningOpRef<Operation *> parsedOp =
+      parseSourceFile<Operation *>(sourceMgr, context);
+  if (!parsedOp)
+    return op->emitError("failed to parse edited IR");
+
+  // Replace the operation's regions and attributes with the parsed content.
```
ftynse: If the user modified the root operation, this may result in invalid IR. We should check that the kind of operation remains the same. (It is also sufficient to continue the pass manager execution).

water/lib/Transforms/EditIR.cpp
```diff
@@ -0,0 +1,174 @@
...
+  for (unsigned i = 0; i < op->getNumRegions(); ++i) {
+    Region &existing = op->getRegion(i);
+    Region &parsed = (*parsedOp)->getRegion(i);
+
+    existing.getBlocks().clear();
+    existing.getBlocks().splice(existing.getBlocks().end(), parsed.getBlocks());
+  }
+
+  op->setAttrs((*parsedOp)->getAttrDictionary());
```
ftynse: I wonder if this also carries over properties.

water/lib/Transforms/EditIR.cpp
```diff
@@ -0,0 +1,174 @@
...
+  (void)std::fgets(buf, sizeof(buf), stdin);
+
+  // Re-parse the (potentially edited) file.
+  auto fileOrErr = llvm::MemoryBuffer::getFile(tmpPath);
+  if (!fileOrErr)
+    return op->emitError() << "failed to read edited file: "
+                           << fileOrErr.getError().message();
+
+  llvm::SourceMgr sourceMgr;
+  sourceMgr.AddNewSourceBuffer(std::move(*fileOrErr), SMLoc());
```
ftynse: Perhaps a good idea to create an `SMLoc` that somehow describes the IR was hand-edited.

water/lib/Transforms/EditIR.cpp
ftynse: OTOH, I'd probably just loop back to editing the file here. It will be very frustrating for the user to edit the file manually and have it discarded entirely because of a syntax error. Perhaps emit diagnostics as comments into the same file somehow.

This also needs a mechanism for the user to abort the compilation entirely.

water/tools/water-opt/WaterOptMain.cpp
```diff
@@ -50,6 +52,64 @@
 using namespace mlir;
 using namespace llvm;
 
+//===----------------------------------------------------------------------===//
+// --mlir-edit-ir-{before,after} CLI options
+//===----------------------------------------------------------------------===//
+
+static cl::opt<bool> editIRBeforeAll(
+    "mlir-edit-ir-before-all",
```
ftynse: Let's prefix this with `water` rather than `mlir`. Makes it easier to understand where the option is defined.

water/lib/Transforms/EditIR.cpp
```diff
@@ -0,0 +1,174 @@
...
+//===----------------------------------------------------------------------===//
+
+/// Resolve which editor command to use: the explicit override if non-empty,
+/// otherwise $EDITOR. Returns an empty string when neither is available.
+static std::string resolveEditor(StringRef editorOverride) {
+  if (!editorOverride.empty())
+    return editorOverride.str();
+  if (auto env = llvm::sys::Process::GetEnv("EDITOR"))
+    return *env;
+  return "";
```
tgymnich: An optional would be better here

water/tools/water-opt/WaterOptMain.cpp
```diff
@@ -50,6 +52,64 @@
...
+    cl::init(""));
+
+static cl::opt<std::string> editIREditor(
+    "mlir-edit-ir-editor",
+    cl::desc("Editor command for --mlir-edit-ir-{before,after}[-all] "
+             "(default: $EDITOR)"),
+    cl::init(""));
+
+/// Build a pass filter from an -all bool flag and a single-pass string flag.
+static water::PassFilter buildEditFilter(bool all, const std::string &name) {
```
tgymnich:
```suggestion
static water::PassFilter buildEditFilter(bool all, StringRef name) {
```

water/lib/Transforms/EditIR.cpp
```diff
@@ -0,0 +1,174 @@
...
+  auto program = llvm::sys::findProgramByName(editorCmd);
+  if (!program) {
+    LLVM_DEBUG(llvm::dbgs()
+               << "edit-ir: editor '" << editorCmd << "' not found in PATH\n");
+    return;
+  }
+
+  SmallVector<StringRef> args = {editorCmd, path};
+  std::string errMsg;
+  llvm::sys::ExecuteNoWait(*program, args, /*Env=*/std::nullopt,
```
tgymnich, line 53: If you were to wait here, you wouldn't need the user to press enter once they are done.

water/lib/Transforms/EditIR.cpp
```diff
@@ -0,0 +1,174 @@
...
+    tryOpenInEditor(editorCmd, tmpPath);
+
+  // Prompt the user and wait for Enter (or EOF).
+  llvm::errs() << "=== mlir-edit-ir";
+  if (!passLabel.empty())
+    llvm::errs() << " " << passLabel;
+  llvm::errs() << " ===\n"
+               << "IR written to: " << tmpPath << "\n"
+               << "Edit the file, then press Enter to continue "
+                  "compilation...\n";
```
tgymnich: check if either `editorCmd` is empty or `tryOpenInEditor` failed, only prompt the user in that case. (if the editor opens, you can just wait for it to close instead of prompting the user)

water/lib/Transforms/EditIR.cpp
```diff
@@ -0,0 +1,174 @@
...
+  if (!editorOverride.empty())
+    return editorOverride.str();
+  if (auto env = llvm::sys::Process::GetEnv("EDITOR"))
+    return *env;
+  return "";
+}
+
+/// Try to open `path` in `editorCmd`. Failures are non-fatal (the user can
+/// always open the file manually via the printed path).
+static void tryOpenInEditor(StringRef editorCmd, StringRef path) {
```
tgymnich: this should probably return a `LogicalResult`

water/lib/Transforms/EditIR.cpp
martin-luecke: Scope ensures raw_fd_ostream flushes its internal buffer and closes the fd before the editor reads the file. 
We could also have an explicit `tmpFile.flush()` if you prefer
Added a comment.

water/lib/Transforms/EditIR.cpp
martin-luecke: added an explicit `copyProperties`

water/lib/Transforms/EditIR.cpp
martin-luecke: The location already shows the name of the file used for editing, i.e. water-edit-XXXXXX.mlir
An error message will e.g. read as
```
/var/folders/rs/8903yyq978b8my25n1r1pljr0000gn/T/water-edit-fd0c85.mlir:1:1: error: custom op 'INVALID' is unknown (tried 'builtin.INVALID' as well)
```
Do you prefer something more explicit?

water/lib/Transforms/EditIR.cpp
martin-luecke: `ExecuteAndWait` only works when the editor blocks (`terminal editors`, `code --wait`, scripts), but GUI editors like `code` or `cursor` without `--wait` fork and return immediately. The prompt also supports the no-editor case, where the user opens the file manually, and the automation case, where an external process modifies the file.

water/lib/Transforms/EditIR.cpp
martin-luecke: see above. I am trying to be defensive when opening an editor does not work and support building automation around this feature

water/lib/Transforms/EditIR.cpp
ftynse: What happens when there is already a location attached prior to the edit, i.e., the explicit `loc(#loc1)` next to the malformed op?

martin-luecke: Added `OpPrintingFlags().enableDebugInfo()` to preserve original locations through printing/parsing roundtrip.
We could wrap the existing locs with `FusedLoc`, but that would require walking all ops for this.
Or do you see another option?

water/lib/Transforms/EditIR.cpp
martin-luecke: Done 👍

martin-luecke:
Implemented this in a loop so users may re-edit if there is a syntax error. PTAL

tgymnich (approved):
LGTM

PR #1296, opened 2026-04-23
Add stream-K MXFP4 GEMM kernel and tests

willghatch:
This is a functional streamk for mxfp4.

Persistent-kernel approach: a fixed number of CTAs iterate over work units, each covering a K-range of an output tile.  Uses atomic_add for accumulation (caller must zero-init C).

Additionally, fix atomic_add codegen to use write-side index mapping

handle_generic_atomic called transform_index_on_mapping with the default is_read=True, so for atomic_add ops with an IndexMapping the *input* side of the mapping was used to compute the destination address.  When the mapping carries write-side dynamic offsets (e.g. CTA_M_OFFSET, CTA_N_OFFSET in stream-K kernels) those offsets were silently ignored, causing all CTAs to write to the same tile.

Pass is_read=False so the output side of the mapping is used instead.

Made-with: Cursor

### DIFF
--- a/tests/kernel/wave_gemm_test.py
+++ b/tests/kernel/wave_gemm_test.py
@@ -51,6 +51,7 @@ from wave_lang.kernel.wave.constraints import (
 from wave_lang.kernel.wave.utils.mxfp_utils import (
     generate_gemm_afp4wfp4_inputs,
     torchScaledGemmMXFP4,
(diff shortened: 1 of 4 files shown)


willghatch:
@harsh-nod As mentioned, this is one of the splitk/streamk branches that was waiting for the splitk PR.

